PR #166, opened 2014-11-26
Github tag should be normalized to taskwarrior tags

djmitche:
Non-alphanumerics should be collapsed to `_`, if I understand tag syntax correctly.


ralphbean:
Guh, yes it does.  thanks for noticing @djmitche.

djmitche:
sorry about that

ralphbean:
Merged into develop with some work in 634601f

Issue #132, opened 2014-06-25
The jira(and maybe other adaptors) should state their dependencies


nikolavp:
I just installed bugwarrior on a new machine and was surprised with the message 
```
'jira' in [my_name] is not a valid service.
```

while it was trying to initialize the configuration file.

I had to install(looked at the code)
```
sudo pip install jira
```
to be able to use the jira integration. I am not familiar with the discovery mechanism of the services but can you at least provide a message which tells the user what to install instead of saying that this is an invalid option(I thought that my file was for another version)

PR #175, opened 2014-12-29
Adding minimal documentation regarding what external packages are requir...

coddingtonbear:
...ed for each service.

At least partially addresses #132.


ralphbean:
Good stuff.  Thanks @coddingtonbear

PR #178, opened 2015-01-07
Fix initial setup error on OSX.

mavant:
This is a two-character change. I think the original was just a typo.


fradeve:
For some reason this error showed up when I've installed 1.0.2 from pip today, I had to fix this manually.

ralphbean:
Yeah, we haven't had a release in a while.  We're currently waiting on the release of taskwarrior-2.4.1 and we'll cut a new tarball of bugwarrior after that.

PR #183, opened 2015-01-20
Show a message to the user in the event that we were unable to perform t...

coddingtonbear:
...he operation.

As it turns out, the `trace` logger method does not print the trace to the console.


ralphbean:
Yup, cool.

PR #184, opened 2015-01-20
Add uda export command

coddingtonbear:
Currently these are printed to the console during the sync operation, but they're prefixed with logging metadata and are likely to be missed or misused.

This adds a single command that can be ran at any time to generate the current list of active UDAs


ralphbean:
Yeah!  Looks great.

PR #185, opened 2015-01-26
Hack to let task-2.4.x search for url UDAs.

ralphbean:
As discussed in ralphbean/taskw#87.


coddingtonbear:
:+1:

coddingtonbear:
Oh -- as long as we know that'll work for integer PKs, too, of course.

ralphbean:
> Oh -- as long as we know that'll work for integer PKs, too, of course.

Just checked and it works.  :)

PR #190, opened 2015-02-06
XDG support

mathstuf:
Adds support for `XDG_CONFIG_HOME` storage of the configuration file with fallback to the original file.

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -28,7 +28,7 @@ lst = list
 def pull(dry_run):
     """ Pull down tasks from forges and add them to your taskwarrior tasks.
 
-    Relies on configuration in ~/.bugwarriorrc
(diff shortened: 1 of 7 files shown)


ralphbean:
I don't think the fallback works correctly.  This returns ``None`` on my system:

```python
BaseDirectory.load_first_config('bugwarrior')
```

So that means that this fails:

```python
>>> os.path.join(BaseDirectory.load_first_config('bugwarrior'), 'bugwarriorrc')
  File "/usr/lib64/python2.7/posixpath.py", line 77, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

mathstuf:
Oh, oops. Fix incoming.

ralphbean:
WFM.

PR #192, opened 2015-02-07
config: add --flavor option

### LINKED ISSUE #189: Allow setting a different configuration file
I use two task lists, one for personal and another for work. It would be nice to be able to have bugwarrior pull into each one separately. I can think of two ways to do this:

  - add a flag to bugwarrior to specify the configuration file (which I can alias away); or
  - allow multiple `[general]` configuration sections with separate `taskrc` values and `targets` listings.

mathstuf:
Flavors allow bugwarrior to be used to populate multiple taskrc lists.
For example, one might have a task list for work and personal use
separately.

---
Fixes #189.

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -23,19 +23,28 @@ from bugwarrior.db import (
 lst = list
 
 
+def _get_section_name(flavor):
+    if flavor:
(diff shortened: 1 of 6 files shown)


ralphbean:
Awesome.  :+1:!

PR #193, opened 2015-02-07
config: give a meaningful error message for empty targets

mathstuf:


ralphbean:
Nice and simple.  Thanks!

PR #194, opened 2015-02-07
gitlab: remove 'username' configuration

mathstuf:
It isn't used for anything. Instead, gitlab's queries return all
repositories the user has access to.

---
Turns out it isn't used for anything in the code other than as a suffix to the `gitlab://` keyring URL. Also, the API doesn't support querying based on it anyways.


ralphbean:
:+1:

PR #195, opened 2015-02-08
gitlab: verify SSL certs

mathstuf:
Leftover from debugging.


ralphbean:
Ooof.  Good catch.  Thanks!

PR #200, opened 2015-02-09
Bitbucket filter repo

mathstuf:
---
Started off as just adding filtering to bitbucket, but I also added pull request logic. It's really sad that all these sites have slightly different APIisms. Bitbucket is, so far, the worst documented out of itself, gitlab, and github :( . Pinging #129 since now there's a mix of v1 and v2 usage here.

### DIFF
--- a/bugwarrior/docs/services/bitbucket.rst
+++ b/bugwarrior/docs/services/bitbucket.rst
@@ -18,6 +18,34 @@ The above example is the minimum required to import issues from
 Bitbucket.  You can also feel free to use any of the
 configuration options described in :ref:`common_configuration_options`.
 
+Service Features
+----------------
+
+Include and Exclude Certain Repositories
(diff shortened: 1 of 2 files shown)


ralphbean:
> Pinging #129 since now there's a mix of v1 and v2 usage here.

Oof!  That functions as a prompt for me to get v2 usage working on the whole.

> It's really sad that all these sites have slightly different APIisms.

Yeah.  It's gotten much better over time -- it used to be complete nonsense when it existed as a tool intended only for me.  It's much better now than it once was.

Thanks for this!  Merging away.

mathstuf:
> Oof! That functions as a prompt for me to get v2 usage working on the whole.

That'd be great if one were a superset of the other… But migrating what is available in v2 would be nice since v1 is the one that would (presumably) be going away first. Maybe v3 will appear too :/ .

PR #203, opened 2015-02-10
Defer importing services until they are needed.

### LINKED ISSUE #132: The jira(and maybe other adaptors) should state their dependencies

ralphbean:
This should let us expose odd ImportErrors with dependencies in a more
natural way to users (and make the dependencies of the services somewhat
more obvious -- those deps are documented now, but this should help
too).

More informative tracebacks here should fix #132.

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -11,6 +11,7 @@ from twiggy import log
 
 from taskw.task import Task
 
+from bugwarrior.utils import DeferredImportingDict
 from bugwarrior.config import asbool
 from bugwarrior.db import MARKUP, URLShortener, ABORT_PROCESSING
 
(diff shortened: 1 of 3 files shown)


mathstuf:
Looks OK to me. Might be worth adding a test for trying to load a module which doesn't exist to test that the right exception is raised when dependencies are missing.

PR #213, opened 2015-03-03
Alter default JIRA query to handle situations in which instances do not use the column names we are expecting.

coddingtonbear:
This will almost certainly be equivalent to the original query for most users; I just happen to have switched recently to a company that used non-standard ticket status names, and knew enough about the JIRA API to know that the proper column to check is `resolution`.

Let me know if you have any questions, sir!


ralphbean:
Sounds good.  Thanks Adam!

PR #214, opened 2015-03-05
Gitlab pagination

mathstuf:
Gitlab *does* paginate, but our instance hadn't had enough traffic for me to notice it until recently. Also adds some more data and fixes a doc typo. Important commit at the base of the branch in case they are wanted separately.


ralphbean:
Looks good to me.  Thanks @mathstuf.

PR #216, opened 2015-04-16
Phabricator service: Adding option to filter on users and projects

ivan-cukic:
WIP.

The first patch fixes the issue of the python module for phabricator service being called phab, and not phabricator.

The second introduces options to filter out the phabricator tasks and diffs according to the project and the user. (the phabricator service is not really useful now when you have a phabricator setup with many users and projects)

If there is the interest in merging this to mainline, I'll write the documentation for the new options.

PR #217, opened 2015-04-17
Fixes import of PhabricatorService.

dczmer:
Looks like the module was named phab instead of phabricator.


dczmer:
Sorry, I just noticed pull request #216 addresses the same thing

PR #218, opened 2015-04-19
Replace HTTPS link as the certificate is not valid

jonan:


coddingtonbear:
It might be better to instead set the link for the primary domain: https://bugwarrior.readthedocs.org.

jonan:
You're completely right, I didn't notice the full linked had a valid certificate

ralphbean:
Thanks guys!

PR #221, opened 2015-06-26
Add tests to MANIFEST.in

koobs:
Add tests to MANIFEST.in so that they are included in the sdist on PyPI.

This allows users & packagers downstream to QA using release distributions.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,3 +3,4 @@ include README.rst
 include LICENSE.txt
 recursive-include docs *
 recursive-include bugwarrior/docs *.rst
+recursive-include tests *


ralphbean:
> This allows users & packagers downstream to QA using release distributions.

:100:

Thanks!

ralphbean:
@koobs, it should be available [on pypi now](https://pypi.python.org/pypi/bugwarrior).

koobs:
Thanks Ralph :) Hope you didn't re-roll the distfile, it will break downstream packages that rely on it and its checksum ;) Quick bump version!

koobs:
I have another change coming, hang on a tick :)

PR #222, opened 2015-06-26
Enable setuptools test command

koobs:
Enable the canonical setuptools test command to invoke the tests via nose as per documentation:

http://nose.readthedocs.org/en/latest/setuptools_integration.html


ralphbean:
Sure, :+1:!

PR #225, opened 2015-07-14
Escaping fixes + only_on_new_tasks notifications + jira estimate

devenv:
As it cuts the description after some characters  there is sometimes an unmatched bracket
e.g. from jira: "Some bug (only in dev..."
TaskWarrior kinda freaks out at this...


devenv:
Added an option not to fire a notification when pulled via cron and nothing new received...

devenv:
+ in the description triggers task update...
May be these escaping issues should be tested in TW...

devenv:
Added jira estimate UDA

ralphbean:
Thanks for the patches @devenv, but I'm going to close this PR without merging.  Some of the changes need to be submitted to other libraries, other changes need explanation, etc..

PR #226, opened 2015-07-23
Add BZ status features

ryansb:
Add a UDA for bugzilla statuses (for priority weighting based on status) and make "open" statuses configurable.


ralphbean:
Yup!  Solid.  Thanks!

PR #227, opened 2015-07-28
jira estimate UDA

devenv:


bugwarrior/services/jira.py
```diff
@@ -13,6 +13,7 @@ class JiraIssue(Issue):
     URL = 'jiraurl'
     FOREIGN_ID = 'jiraid'
     DESCRIPTION = 'jiradescription'
+    ESTIMATE = 'estimate'
```
coddingtonbear: Unfortunately UDAs are global, so we should really do what we can to not pollute the user's UDAs with service-specific ones.  Could I convince you into changing this to `jiraestimate`?

bugwarrior/services/jira.py
```diff
@@ -91,6 +97,11 @@ def get_summary(self):
             return self.record['fields']['summary']['value']
         return self.record['fields']['summary']
 
+    def get_estimate(self):
+        if self.extra.get('jira_version') == 4:
+            return self.record['fields']['timeestimate']['value']
+        return self.record['fields']['timeestimate'] / 60 / 60
```
coddingtonbear: Are we guaranteed to have a numeric value here? Maybe we should catch `TypeError` and `KeyError' so we can return `None` if either the value is null or that key doesn't exist.

devenv:
Added exceptions + changed to jiraestimate :)

bugwarrior/docs/services/jira.rst
```diff
@@ -109,3 +109,4 @@ Provided UDA Fields
 +---------------------+---------------------+---------------------+
 | ``jiraurl``         | URL                 | Text (string)       |
 +---------------------+---------------------+---------------------+
+| ``estimate``        | Estimate            | Decimal (numeric)   |
```
ralphbean: Ah, can you change the name here in the docs too?  One last thing! :D

devenv:
Hehe, forgot that one :)
Done

ralphbean:
Looks great.  Thanks!  :)

PR #231, opened 2015-08-23
Support needinfo bugs where you are not CC/assignee/reporter

ryansb:
In newer BZ installations, the NEEDINFO status has been replaced with the `needinfo` flag. This commit adds a config option to also query for bugs where you are a requestee on the needinfo flag.

This also adds a "bugzillaneedinfo" UDA so queries like `task bugzillaneedinfo.any: list` return all your needinfo bugs.


ralphbean:
Dope.  Thank you!

PR #233, opened 2015-08-24
Gitlab work in progress flag

mathstuf:


ralphbean:
:+1:.

(bummer that's there's no ``boolean`` UDA type)

mathstuf:
Indeed. Though I guess those are tags?

PR #235, opened 2015-08-25
Only import active Gitlab issues and merge requests

LordGaav:
The Gitlab service seems to always import all issues, including closed ones. This PR changes that behaviour.


ralphbean:
Sure, :+1:.  Thanks!

mathstuf:
I'd like to import closed issues (sometimes there are tasks to continue based on it or it was closed improperly). Could an option be exposed for this? Also, the new involved issues logic should probably add "is:open" to its query with the option set.

PR #237, opened 2015-09-04
Make comment annotation configurable

ryansb:
This patch adds a general config option `annotation_comments` that defaults to True (current behavior) and lets users turn off annotations for issue comments. This is super handy if you're on large github issue threads.

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -75,6 +75,12 @@ class IssueService(object):
                 config.get(self.main_section, 'annotation_links')
             )
 
+        self.annotation_comments = True
+        if config.has_option(self.main_section, 'annotation_comments'):
+            self.annotation_comments = asbool(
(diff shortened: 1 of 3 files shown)


bugwarrior/services/__init__.py
```diff
@@ -165,6 +171,8 @@ def build_annotations(self, annotations, url):
         final = []
         if self.annotation_links:
             final.append(url)
+        if not self.annotation_comments:
+            return final
```
coddingtonbear: Kind of nitpicking, but out of an abundance of caution driven by past bugs I've created in methods having multiple (sometimes surprising) `return`s, I'd prefer it if we used this setting to gate whether the loop below is executed rather than returning early.

coddingtonbear:
Other than that, I think this is great.

ryansb:
@coddingtonbear Makes sense - changed.

ralphbean:
:+1:, thanks guys!

PR #241, opened 2015-09-15
Turn legacy_matching off by default.

ralphbean:
We get weird bug reports for it all the time.

We've had the new code in place long enough that many users have
hopefully migrated to the new UDA scheme.  This should turn it off for
new users by default, hopefully providing a better experience.


coddingtonbear:
+1!

ralphbean:
woot!

PR #249, opened 2015-10-27
Rhbz handle open needinfo

mathstuf:
---
The `requestee` key isn't necessarily required and isn't provided by the API if so. Hit using rhbz.


ralphbean:
Nice find!

Issue #256, opened 2015-11-11
Gitlab example service is missing "gitlab.host" on readthedocs.org


muxync:
To avoid this error:
```
CRITICAL:command|oh noes
TRACE Traceback (most recent call last):
TRACE   File "/usr/local/lib/python2.7/dist-packages/bugwarrior/command.py", line 45, in pull
TRACE     config = load_config(main_section)
TRACE   File "/usr/local/lib/python2.7/dist-packages/bugwarrior/config.py", line 156, in load_config
TRACE     validate_config(config, main_section)
TRACE   File "/usr/local/lib/python2.7/dist-packages/bugwarrior/config.py", line 133, in validate_config
TRACE     get_service(service).validate_config(config, target)
TRACE   File "/usr/local/lib/python2.7/dist-packages/bugwarrior/services/gitlab.py", line 412, in validate_config
TRACE     die("[%s] has no 'gitlab.host'" % target)
TRACE   File "/usr/local/lib/python2.7/dist-packages/bugwarrior/config.py", line 98, in die
TRACE     sys.exit(1)
TRACE SystemExit: 1
```

Something like this should be added to https://bugwarrior.readthedocs.org/en/latest/services/gitlab.html
```
gitlab.host = gitlab.com
```
Where is the source for the documentation?  The link on readthedocs.org is broken as per #254.

coddingtonbear:
Good find, @muxync -- could you file a pull request adding that?

All of the documentation is generated from files in the `bugwarrior/docs` folder -- the specific example you've cited above is here: https://github.com/ralphbean/bugwarrior/blob/develop/bugwarrior/docs/services/gitlab.rst.

muxync:
Thanks, I can't believe I missed that!  See PR #258.

erm3nda:
Docs still not showing the gitlab options

ryneeverett:
@erm3nda Not sure what you mean. I can see them.

erm3nda:
Sorry, it's on the docs, but not in the bugwarriorrc sample

PR #258, opened 2015-11-11
Add gitlab.host to example Gitlab target

muxync:
See #256.


ralphbean:
Looks good!  Thanks @muxync!

Issue #259, opened 2015-11-11
Gitlab issues always updated even when no changes have been made


muxync:
Every time bugwarrior-pull is executed the corresponding task is updated regardless of if the issue on Gitlab was updated.  For example:
```
INFO:db|Updating task 4a55b0ac-0bb5-48bb-8f79-b9aebd270254, (bw)Is#1 - demo bugwarrior gitlab issue; gitlabcreatedon: datetime.datetime(2015, 11, 11, 17, 25, 47, tzinfo=tzutc()) -> datetime.datetime(2015, 11, 11, 17, 25, 47, 957000, tzinfo=tzutc()); gitlabupdatedat: datetime.datetime(2015, 11, 11, 17, 25, 47, tzinfo=tzutc()) -> datetime.datetime(2015, 11, 11, 17, 25, 47, 957000, tzinfo=tzutc())
```
The problem can be corrected by discarding microseconds for the issue 'created' and 'updated' times.  Are microseconds necessary here?

ralphbean:
Looks like this is solved in #260.  You're the best @muxync!

PR #260, opened 2015-11-11
Ignore microseconds for gitlab

muxync:
To prevent issue updates on every bugwarrior-pull.  See #259.


ralphbean:
Oh, nice one!  Thanks!

PR #264, opened 2015-11-21
Fix the broken tests

gdetrez:
- some string was bytes instead of unicode
- microseconds needed to be ignored in the test as well


ralphbean:
Great, thanks :)

PR #265, opened 2015-11-21
Fix some subtle option parsing problems:

gdetrez:
- If the section doesn't exist, `_bool_option` should return the given
  default value, not `False`
- The fourth argument of `conf.get()` is not a fallback value but controls
  how the value is parsed ([1])
- `"False"` is true-ish, which was causing `legacy_matching` to be true by
  default

[1] https://docs.python.org/2/library/configparser.html#ConfigParser.ConfigParser.get

In addition. the `legacy_matching` option behave erratically if the issue description contains pipes (`|`). But since it seems deprecated, I don't know if it's worth fixing.

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -1,4 +1,4 @@
-from ConfigParser import NoOptionError
+from ConfigParser import NoOptionError, NoSectionError
 import os
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


ralphbean:
Oh, good good good.  Thank you!

FWIW, I'd love to kill off ``legacy_matching`` two minor version bumps from now.

PR #266, opened 2015-11-21
Add missing test dependencies

gdetrez:
I had to install pypandoc and pyac for all the tests to pass.  Note that
pypandoc probably requires pandoc to be installed as well.


ralphbean:
Oh, nice.  Thanks!

PR #273, opened 2016-01-25
Minimal CI and documentation fixes.

ryneeverett:
Simply running `nosetests` wont work because dependencies aren't installed
in the virtualenv. You'd have to manually `pip install` each dependency
like travis currently does.

Nose recommends this as "the easiest way to run tests with nose".
http://nose.readthedocs.org/en/latest/api/commands.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
- - "2.6"
  - "2.7"
 before_install:
   - sudo apt-get update -qq
--- a/bugwarrior/docs/common_configuration.rst
+++ b/bugwarrior/docs/common_configuration.rst
@@ -17,7 +17,7 @@ Optional options include:
   Defaults to ``True``.
 * ``annotation_links``: When ``True`` will include a link to the ticket as an
   annotation. Defaults to ``False``.
-* ``annotation_comments``: When ``True`` skips putting issue comments into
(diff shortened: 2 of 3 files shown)


coddingtonbear:
I *think* the best way forward might be to just instruct people to run `python setup.py test` for running tests; that should install the necessary dependencies automatically.

ryneeverett:
@coddingtonbear `python setup.py nosetests` seems to be the same thing but with nosetests as the testrunner. If you want to stop suggesting people use nose as a testrunner that's fine by me, but this way you can have your cake and eat it too.

coddingtonbear:
@ryneeverett -- my understanding is that since we've [set `nose.collector` as our test suite](https://github.com/ralphbean/bugwarrior/blob/develop/setup.py#L57), nose will *always* be the test runner, is that not your understanding?

coddingtonbear:
Actually -- I have to admit that I hadn't read your PR very carefully before posting my first comment above.  I'm +1 on making either the change you've suggested or the one I have.

ryneeverett:
> my understanding is that since we've set nose.collector as our test suite, nose will always be the test runner, is that not your understanding?

@coddingtonbear That makes sense, but when I run `python setup.py test` I don't get the nose output, I get standard unittest output.

ralphbean:
Good stuff.  Thanks both!

PR #274, opened 2016-01-25
Use TASKRC environmental variable when assigned.

ryneeverett:


ryneeverett:
Based on #273 since that's necessary for travis to pass.

ryneeverett:
@ralphbean No apology necessary. FYI, I rebased my other PR's so the diffs are more readable.

PR #275, opened 2016-01-25
bitbucket.login is a required setting

ryneeverett:
The implementation seems to be the same as github.


ralphbean:
Cool, :+1:.  Thanks!

PR #276, opened 2016-01-26
bitbucket: More v2 API. Progress on #129.

ryneeverett:
- Moving most things to the v2 API makes some things cleaner.
- One DRY get_data function that takes an optional api kwarg.
- Still no way to get detailed comment data without making an extra api
  call, so let's put that off for now.

### DIFF
--- a/bugwarrior/services/bitbucket.py
+++ b/bugwarrior/services/bitbucket.py
@@ -41,7 +41,7 @@ class BitbucketIssue(Issue):
             'annotations': self.extra['annotations'],
 
             self.URL: self.extra['url'],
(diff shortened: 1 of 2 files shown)


ralphbean:
@ryneeverett, is this all working for you locally?  It looks fine to me, but I haven't combed through the v2 API docs yet myself.

ryneeverett:
@ralphbean Yep, I've been using all this stuff for a few weeks now without a hitch. I just fixed merge conflicts and rebased.

ralphbean:
Nice.  :+1: then, and thanks!

PR #277, opened 2016-01-27
Add --interactive flag to bugwarrior-pull.

ryneeverett:


ralphbean:
Ha, that should do it.  :)  :+1:  Thanks!

PR #281, opened 2016-01-30
Bitbucket closed status

ryneeverett:
This status was recently added and still isn't in the API docs, but the
API does return it.
https://bitbucket.org/site/master/issues/5179/add-a-status-of-closed-to-issues-bb-10249


ralphbean:
Cool.  :+1: here.

PR #285, opened 2016-02-15
Password oracle improvements

gdetrez:
* Fix a bug that prevented using an `@oracle:` token in gitlab
* improve error reporting in case a `@oracle:eval:` commands fails
* fix a bug that made bugwarrior hang when the oracle failed

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -71,6 +71,8 @@ def pull(dry_run, flavor, interactive):
                 lockfile_path
             )
         )
+    except RuntimeError as e:
+        log.name('command').critical("Aborted (%s)" % e)
     except:
         log.name('command').trace('error').critical('oh noes')
 
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -56,18 +56,25 @@ def get_service_password(service, username, oracle=None, interactive=False):
         password = getpass.getpass(prompt)
     elif oracle.startswith('@oracle:eval:'):
         command = oracle[13:]
-        p = subprocess.Popen(
-            command,
-            shell=True,
-            stdout=subprocess.PIPE,
-            #stderr=subprocess.STDOUT
-        )
-        password = p.stdout.read()[:-1]
+        return oracle_eval(command)
 
     if password is None:
         die("MISSING PASSWORD: oracle='%s', interactive=%s for service=%s" %
             (oracle, interactive, service))
     return password
+
+
+def oracle_eval(command):
+    """ Retrieve password from the given command """
+    p = subprocess.Popen(
+        command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    p.wait()
+    if p.returncode == 0:
+        return p.stdout.readline().strip()
+    else:
+        die(
+            "Error retrieving password: `{command}` returned '{error}'".format(
+                command=command, error=p.stderr.read().strip()))
 
 
 def load_example_rc():
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -26,10 +26,6 @@ CACHE_REGION = dogpile.cache.make_region().configure(
 )
 
 
-# Sentinel value used for aborting processing of tasks
-ABORT_PROCESSING = 2
-
-
 class URLShortener(object):
     _instance = None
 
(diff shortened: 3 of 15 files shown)


gdetrez:
Sorry about all the opening and closing of PRs :confused:

ralphbean:
Hey!  No worries.  Thanks for all the contributions here!

I've been busy and travelling.  I hope to get back to these to review them soon!

ralphbean:
I think the first two fixes are awesome and should be merged.. but the third one that changes the way errors are propagated back from the services to the main thread, I'm not as fond of.

First, it may just not be necessary.  A call to ``die()`` mostly just calls ``sys.exit(1)`` which itself raises an exception:

```python
try:
    sys.exit(1)
except SystemExit:
    print "caught it"
```

We (I think) just keep to try and catch that exception when the services are initializing themselves in order to cleanly shut everything down.

That aside -- we currently place a "sentinel value" on the outgoing queue indicating that everything should start shutting down, instead of raising exceptions.  If we did decide to start using exceptions as the mechanism to signal failure, then I guess I'd like to avoid having many custom-defined exceptions.  There are hundreds of built-in exceptions we could re-use (at very least, we could extend a nice-looking one).

Would you be open to rebasing this change to remove the Exception propagation stuff, and re-doing that with the existing sentinel machinery?  It's not necessarily a bad change (exceptions make a lot of sense), I'd just like to avoid making waves there unless we have a good reason to.

gdetrez:
I didn't change the way error are propagated back to the main thread: the sentinel value passed in the multiprocessing queue is still there, but I changed the exception propagation inside of both threads (main and worker). As you pointed out, my comment was not really correct: I didn't use more exceptions, just different ones :wink:. The new commits now only use built-in exceptions.

An issue with `SystemExit` was that it is not a regular exception, in the sense that it isn't caught by `except Exception`.  The change in 95f452b should make bugwarrior more robust to those system-exiting exceptions.

The problem that I'm trying to solve is that the important error message (the fact that the oracle command failed) is buried over five pages of mostly irrelevant output. First `die` outputs 4 pages of example configs (and in the case of an error in the oracle, none of it is relevant) and then there are the stack traces of SystemExit and RuntimeError. In  a4e6744, I suggest to catch those exceptions separately mostly to suppress the stack trace. But that still leaves me with four pages to scroll up to see what went wrong.

gdetrez:
I rebased the PR on the latest develop commits and properly fixed the bitbucket test (thanks to @ryneeverett).

ryneeverett:
Looking good, though I still don't care for the `config_has` thing. IMO `service.password=` as a shorthand for `service.password=@oracle:keyring` is not worth the extra complexity in the code base and would also be more confusing for the end user than explicitly specifying `@oracle:keyring`.

gdetrez:
@ryneeverett I do understand your concerns and I think I share your opinion on the interpretation of an empty password or token. But what I see as a bug here is only the inconsistency (i.e. the fact that an empty `github.token` is interpreted differently than an empty `github.password` for example). I don't feel it's my place to decide what is the correct interpretation so I'm trying to fix it in the less disruptive way by aligning `github.token` with what seems to be the majority interpretation among different services.

Now if there's a consensus that the overall interpretation of empty values should change in bugwarrior (i.e. that every empty `xxx.password=` should be treated as a missing value), I'd be happy to help with the refactoring while this code is still (relatively) fresh in my mind, but I'd rather do that in an other pull request as this one is already becoming a bit of an hodgepodge of different modifications.

ryneeverett:
Preserving the undocumented "majority interpretation" (that is, already inconsistent behavior) of a corner case (which you're likely the first person to discover) is adding to the hodgepodge. The behavior of an undocumented configuration is ambiguous and should remain ambiguous unless we decide to intentionally support it or break it -- which I agree should be a separate PR.

ralphbean:
Yeah, that behavior is unintentional.  Thanks much for cycling on this.  :+1: to taking up the question of how to resolve the ambiguity in a separate issue/PR.  I'm going to merge it as-is.

@ryneeverett, @gdetrez, would you two be interesting in becoming bugwarrior maintainers?  I could give you commit rights on the repo and empower you to call shots and merge pull requests in the future.  I'd only ask that you not commit directly to the ``develop`` branch and instead submit PRs that at least one other maintainer can inspect and review.

ralphbean:
> I'm in.

Done!

gdetrez:
@ralphbean I would be happy to help wherever I can.

@ryneeverett Thanks! And thanks for all your help :heart:!

PR #286, opened 2016-02-20
Avoid time out with Phabricator installations with huge userbase

johl:
If self.shown_user_phids or self.shown_project_phids is set, restrict API calls to user_phids or project_phids to avoid time out with Phabricator installations with huge userbase.


ralphbean:
:+1: to the concept here.  I ask, @johl, can you add these to the documentation for the phab plugin before we merge?

johl:
@ralphbean Sure, will do.

johl:
@ralphbean Added something to the documentation.

ralphbean:
Wonderful.  Thank you!  :custard:

PR #287, opened 2016-02-28
Move some test dependencies to install dependencies

gdetrez:
This concerns jira, megaplan, pypandoc and pyac which are used in the
services jira, megaplan and activecollab.

I also removed them from the travic setup since they should now be
installed automatically.


ralphbean:
I intentionally omitted these from the install-time requirements for a reason back-when.  I believe the issue was that the ``jira`` dep required a version of ``requests`` that conflicted with another plugin.  So I made them optional so that users could sort out that mess on their own (not the best out-of-the-box experience).

Perhaps those problems in the ecosystem have since been resolved.  Do you have any insight on this?  Have you seen any issues?  I'd be willing to try it and see if we get any feedback.

ralphbean:
> @ralphbean I have a different proposal: using the extras_require to declare them as "optional dependencies".

:+1: agreed.  This is the way to do it.

ralphbean:
Looks very good.  Thanks @gdetrez!

PR #288, opened 2016-03-04
Fix bitbucket undocumented API change.

ryneeverett:
I'm getting `state` rather than `status` for issues, though `status` is
the documented name <https://confluence.atlassian.com/bitbucket/issues-resource-296095191.html>.


ralphbean:
Nice.  Seems odd to see so many inconsistencies from a site like bitbucket.

PR #289, opened 2016-03-05
Fix nosetests.

ryneeverett:
When these requirements were in tests_require nose installed them
automatically. <s>Now that they've been moved to extras_require they'll
have to be installed manually.</s> We can get this behavior back by adding the extras packages to `tests_require`.


ralphbean:
:+1:

PR #290, opened 2016-03-05
Add Coveralls coverage testing to CI.

ryneeverett:
I don't think the number matters that much and I wouldn't want to make
coverage a hard rule for PR's, but I think keeping an eye on the affect
our changes make would help us move in the right direction.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,12 @@ before_install:
   - pip install mock
   - pip install unittest2
   - cd $TRAVIS_BUILD_DIR
-install: pip install .[jira,megaplan,activecollab]
-script: nosetests -w tests
+install:
+  - pip install .[jira,megaplan,activecollab]
+  - pip install coveralls
+  - pip install coverage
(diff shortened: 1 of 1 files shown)


ryneeverett:
*(Ironically?)* I haven't tested this out. Based on [the docs](https://github.com/coagulant/coveralls-python#usage-travis-ci) I think all we'd have to make a free coveralls account for this to work.

ralphbean:
I've got an account there.  Let's try it!

ryneeverett:
I went into coveralls and turned it on so we'll see. I'm hoping there are no PR comments because I find them to be unjustifiably loud negative feedback. According to [the docs](https://coveralls.zendesk.com/hc/en-us/articles/201774895-Notifications) it looks like we'd have to go to extra effort to turn them on. I just can't imagine why so many people have done that though.

ryneeverett:
Realized I could go into travis and force builds. Works fine on latest commits, but when I tried on the latest PR it failed with a weird git error. I'm thinking it has something to do with the fact that it's already merged.

PR #292, opened 2016-03-11
Replace development config with --debug flag.

ryneeverett:
The 'development' configuration option is not currently documented. Rather than document
it, I suggest using a flag because it's more convenient (doesn't require
constantly editing a file) and self-documenting.


ralphbean:
Nice.  Much better.  :+1: to merge.

PR #293, opened 2016-03-25
Purge unittest2 now that we only test in 2.7+.

ryneeverett:
We dropped python 2.6 from CI in #273 so the backport is no longer
needed.


ralphbean:
Good and clean.  :+1: to merge, @ryneeverett.

PR #299, opened 2016-04-06
Replace remaining urrlib2 usage with requests.

ryneeverett:
Note: I do not have any way to test this out on these services.

PR #300, opened 2016-04-06
Service Client Base Class

ryneeverett:
This seems to be the direction the abstraction is going in and facilitates DRYing up the json response handling.

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -488,6 +488,23 @@ class Issue(object):
         return '<%s>' % self.__unicode__()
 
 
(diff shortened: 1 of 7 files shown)


ralphbean:
:+1: to the ``ServiceClient`` change.

Let's wait to see if we get a response on #299.

ryneeverett:
Rebased and should be ready to go.

@ralphbean I used the `json_response` method in the taiga service and it doesn't have exactly the same behavior as the validation code you wrote. Is that okay?

ralphbean:
Yeah, that's fine.  Thanks @ryneeverett!  Good to merge here.

PR #301, opened 2016-04-12
A service for pulling issues from a taiga instance.

ralphbean:
It works!  Any feedback?

### DIFF
--- a/bugwarrior/services/taiga.py
+++ b/bugwarrior/services/taiga.py
@@ -23,7 +23,7 @@ class TaigaIssue(Issue):
             'label': 'Taiga URL',
         },
         FOREIGN_ID: {
-            'type': 'string',
+            'type': 'numeric',
             'label': 'Taiga Issue ID'
         },
     }
--- a/bugwarrior/services/taiga.py
+++ b/bugwarrior/services/taiga.py
@@ -57,7 +57,7 @@ class TaigaService(IssueService):
     def __init__(self, *args, **kw):
         super(TaigaService, self).__init__(*args, **kw)
         self.url = self.config_get('base_uri')
-        self.auth_token = self.config_get_password('auth_token', 'nousername')
+        self.auth_token = self.config_get_password('auth_token')
         self.label_template = self.config_get_default(
             'label_template', default='{{label}}', to_type=six.text_type
         )
--- a/bugwarrior/services/taiga.py
+++ b/bugwarrior/services/taiga.py
@@ -1,0 +1,132 @@
+from __future__ import absolute_import
+
+import requests
+import six
+
+from bugwarrior.db import CACHE_REGION as cache
+from bugwarrior.config import die
+from bugwarrior.services import IssueService, Issue
+
+
+class TaigaIssue(Issue):
+    SUMMARY = 'taigasummary'
+    URL = 'taigaurl'
+    FOREIGN_ID = 'taigaid'
+
+    UDAS = {
+        SUMMARY: {
+            'type': 'string',
+            'label': 'Taiga Summary'
+        },
+        URL: {
+            'type': 'string',
+            'label': 'Taiga URL',
+        },
+        FOREIGN_ID: {
+            'type': 'string',
+            'label': 'Taiga Issue ID'
+        },
+    }
+    UNIQUE_KEY = (URL, )
+
+    def to_taskwarrior(self):
+        return {
+            'project': self.extra['project'],
+            'annotations': self.extra['annotations'],
+            self.URL: self.extra['url'],
+
+            'priority': self.origin['default_priority'],
+            'tags': self.record['tags'],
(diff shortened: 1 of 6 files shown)


bugwarrior/services/taiga.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class TaigaService(IssueService):
+    ISSUE_CLASS = TaigaIssue
+    CONFIG_PREFIX = 'taiga'
+
+    def __init__(self, *args, **kw):
+        super(TaigaService, self).__init__(*args, **kw)
+        self.url = self.config_get('base_uri')
+        self.auth_token = self.config_get_password('auth_token', 'nousername')
```
ryneeverett: Would it be better make `login` an optional argument in the `IssueService` class? E.g.:

```py
    def config_get_password(self, key, login='nousername')
```

bugwarrior/services/taiga.py
```diff
@@ -0,0 +1,132 @@
...
+            'label_template': self.label_template,
+        }
+
+    @classmethod
+    def validate_config(cls, config, target):
+        for option in ('taiga.auth_token', 'taiga.base_uri'):
+            if not config.has_option(target, option):
+                die("[%s] has no '%s'" % (target, option))
+
+        IssueService.validate_config(config, target)
```
ryneeverett: Isn't it more idiomatic to use `super` here?

bugwarrior/services/taiga.py
```diff
@@ -0,0 +1,132 @@
...
+    CONFIG_PREFIX = 'taiga'
+
+    def __init__(self, *args, **kw):
+        super(TaigaService, self).__init__(*args, **kw)
+        self.url = self.config_get('base_uri')
+        self.auth_token = self.config_get_password('auth_token', 'nousername')
+        self.label_template = self.config_get_default(
+            'label_template', default='{{label}}', to_type=six.text_type
+        )
+        self.session = requests.session()
```
ryneeverett: Any reason not to pass the headers here? See [sessions API](http://docs.python-requests.org/en/master/api/#sessionapi). E.g.:

```py
self.session = requests.session(headers={
    'Accept': 'application/json',
    'Authorization': 'Bearer %s' % auth_token})
```

ryneeverett:
Could you add a test module with at least the `test_to_taskwarrior` test case?

bugwarrior/services/taiga.py
ralphbean: Likely, yes.  Let's do that.  :)

bugwarrior/services/taiga.py
ralphbean: Yes, will change.

bugwarrior/services/taiga.py
ralphbean: No reason not to.  Good critique!

bugwarrior/services/taiga.py
ralphbean: Heh, I don't think I've ever actually used ``super(...)`` from a `@classmethod` before.

Would it look like this?

```python
super(cls, cls).validate_config(config, target)
```

Will mess with it later and report back.

ralphbean: Nope, it's got to be `super(TaigaService, cls).validate_config(config, target)`.  The value of ``cls`` is the grandchild class, which would be inadvertently correct in this case, but would fail if ``TaigaService`` were extended.

ralphbean: I'll save fixing this for another PR, since it is done non-idiomatically across all the services.

bugwarrior/services/taiga.py
ralphbean: Fun fact, the sessions API changed and ``requests.session()`` no longer accepts any arguments.

However, you can still update the headers on the session after creation.

ryneeverett: Are you sure about that? I linked to the "master" docs.

ralphbean: Yeah, try it:

```python
>>> import requests
>>> requests.Session(headers={})
    requests.Session(headers={})
TypeError: __init__() got an unexpected keyword argument 'headers'
```

Also, check the [api changes section of the master docs](http://docs.python-requests.org/en/master/api/#api-changes) for an explicit mention.

ryneeverett: Oh I see, I mistook the `Session` properties for parameters.

ryneeverett:
Looks good. My only reservation is the lack of a `test_to_taskwarrior` test case.

ralphbean:
> My only reservation is the lack of a test_to_taskwarrior test case.

Agreed.  I finally got time to do it.  Should be all finished up now.

ryneeverett:
LGTM

ralphbean:
Thanks @ryneeverett!

PR #303, opened 2016-04-18
Fetch github comments only when they will be used

ryansb:
Prior to this change, comments on Github issues were fetched
(generating their own API request) whether or not they were saved in
taskwarrior per the `annotation_comments` config option. This change
only fetches them if `annotation_comments` is True, saving requests &
time.


ralphbean:
:+1:

PR #305, opened 2016-04-22
A gerrit service.

ralphbean:
It works!

I noticed while doing this that the IssueService.validate_config call is still
non-idiomatic, fwiw.

### DIFF
--- a/bugwarrior/services/gerrit.py
+++ b/bugwarrior/services/gerrit.py
@@ -96,7 +96,9 @@ class GerritService(IssueService, ServiceClient):
             '?q=is:open+is:reviewer' + \
             '&o=MESSAGES&o=DETAILED_ACCOUNTS'
         response = self.session.get(url)
-        changes = json.loads(response.text[4:])
+        # The response has some ")]}'" garbage prefixed.
+        body = response.text[4:]
+        changes = json.loads(body)
 
         for change in changes:
             extra = {
--- a/bugwarrior/services/gerrit.py
+++ b/bugwarrior/services/gerrit.py
@@ -92,6 +92,8 @@ class GerritService(IssueService, ServiceClient):
         IssueService.validate_config(config, target)
 
     def issues(self):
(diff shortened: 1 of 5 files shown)


ryneeverett:
Did you intend to commit gerrit.py?

bugwarrior/services/gerrit.py
```diff
@@ -0,0 +1,123 @@
...
+            if not config.has_option(target, option):
+                die("[%s] has no '%s'" % (target, option))
+
+        IssueService.validate_config(config, target)
+
+    def issues(self):
+        url = self.url + '/a/changes/' + \
+            '?q=is:open+is:reviewer' + \
+            '&o=MESSAGES&o=DETAILED_ACCOUNTS'
+        response = self.session.get(url)
```
ryneeverett: Why not:

```py
response = self.session.get(
    self.url,
    {'q': 'is:open+is:reviewer', 'o': ['MESSAGES', 'DETAILED_ACCOUNTS']})
```

bugwarrior/services/gerrit.py
```diff
@@ -0,0 +1,123 @@
...
+        changes = json.loads(response.text[4:])
```
ryneeverett: What are you slicing out of the response here?

ralphbean:
> Did you intend to commit gerrit.py?

Aww... no.  :(  Time to change that pw.

bugwarrior/services/gerrit.py
ralphbean: It's some ``)]}'`` garbage.  Will add a comment.

bugwarrior/services/gerrit.py
ralphbean: Ah, because requests will percent-encode the ``:`` characters, and gerrit doesn't like that.  :(

ralphbean:
@ryneeverett, updated and rebased.

ryneeverett:
LGTM

PR #307, opened 2016-04-25
Complain when users specify old config values.

ralphbean:
In 81e8d74d0, we accidentally changed the config value from
``only_if_assigned`` to ``$SERVICE.only_if_assigned``.  That's okay by me,
just because it makes things consistent.  But, it breaks people's existing
configs (it broke mine).

This change adds some warnings which will halt bugwarrior and prompt you to
update your config.

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -196,7 +196,7 @@ class IssueService(object):
     def include(self, issue):
         """ Return true if the issue in question should be included """
         only_if_assigned = self.config_get_default(
-            'only_if_assigned', None, asbool)
+            'only_if_assigned', None)
 
         if only_if_assigned:
             owner = self.get_owner(issue)
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -191,7 +191,12 @@ class IssueService(object):
     @classmethod
     def validate_config(cls, config, target):
         """ Validate generic options for a particular target """
-        pass
+        if config.has_option(target, 'only_if_assigned'):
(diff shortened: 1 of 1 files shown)


ryneeverett:
> That's okay by me, just because it makes things consistent.

Isn't the current consistency that [Common Service Configuration Options](http://bugwarrior.readthedocs.org/en/latest/common_configuration.html#common-service-configuration-options) do not have the `$SERVICE.<option>` scheme?

If we wanted to eliminate this distinction, I would think we'd want to go in the opposite direction -- removing the `$SERVICE.` part -- because it's a lot of text in the config file not adding any information since it's already specified by the `service` setting. But I don't feel strongly about that.

coddingtonbear:
The `$SERVICE.<option>` format is my fault, and you can feel free to remove it if you so desire, but I do personally find it useful given that INI files have special behavior w/r/t how the section named `[DEFAULT]` works.

ryneeverett:
From a quick search, my understanding is that if a key isn't found in a given section, it will fall back to the one found in `[DEFAULT]`.

If that is correct, then I agree that putting service-specific values in their own namespace makes sense. We don't want to make falling back to a nonsensical default value a possibility. *Because nobody reuses their usernames and passwords, right? Err, bad example.*

And since `only_if_assigned` takes a username, it is also service-specific and belongs in the `$SERVICE` namespace. But the other "Common Service Configuration Options" are not service-specific and should remain as they are.

Does that make sense?

coddingtonbear:
Ahh, sure, that does make sense!

ryneeverett:
Cool. LGTM.

ralphbean:
OK.  Will merge this soon.

PR #308, opened 2016-05-07
Simplify SerivceClient.json_response interface.

ryneeverett:
The only difference is that errors will include the full url (i.e,
inluding params), which is probably a good thing anyway.

This also fixes the regression I introduced into taiga in which I forgot
to include the url.


ralphbean:
Nice.  :+1: here.

PR #309, opened 2016-05-13
Refactor githubutils into a ServiceClient.

ryneeverett:
Part of me hates replacing all these pure functions with a class, but I
find it much more readable, less repetitive, and more congruent with the
other services.


ralphbean:
Nice!  :+1:

PR #313, opened 2016-05-20
Bitbucket fixes

gdetrez:
This fixes a couple of issues I noticed with the bitbucket service in the latest version:

* bugwarrior was only pulling a small number of issues from bitbucket
* when using `only_if_assigned`, no issues were imported

### DIFF
--- a/bugwarrior/services/bitbucket.py
+++ b/bugwarrior/services/bitbucket.py
@@ -174,8 +174,7 @@ class BitbucketService(IssueService, ServiceClient):
 
     def fetch_issues(self, tag):
         response = self.get_collection('/repositories/%s/issues/' % (tag))
-        for issue in response:
-            yield (tag, issue)
+        return [(tag, issue) for issue in response]
 
     def fetch_pull_requests(self, tag):
         response = self.get_collection('/repositories/%s/pullrequests/' % tag)
@@ -219,7 +218,7 @@ class BitbucketService(IssueService, ServiceClient):
             if repo.get('has_issues')
         ])
 
-        issues = sum([list(self.fetch_issues(repo)) for repo in repo_tags], [])
+        issues = sum([self.fetch_issues(repo) for repo in repo_tags], [])
         log.name(self.target).debug(" Found {0} total.", len(issues))
 
         closed = ['resolved', 'duplicate', 'wontfix', 'invalid', 'closed']
--- a/bugwarrior/services/bitbucket.py
+++ b/bugwarrior/services/bitbucket.py
@@ -204,8 +204,10 @@ class BitbucketService(IssueService, ServiceClient):
         )
 
     def get_owner(self, issue):
-        tag, issue = issue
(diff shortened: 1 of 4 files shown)


bugwarrior/services/bitbucket.py
```diff
@@ -157,12 +173,13 @@ def validate_config(cls, config, target):
         IssueService.validate_config(config, target)
 
     def fetch_issues(self, tag):
-        response = self.get_data('/repositories/%s/issues/' % (tag))
-        return [(tag, issue) for issue in response['values']]
+        response = self.get_collection('/repositories/%s/issues/' % (tag))
+        for issue in response:
+            yield (tag, issue)
```
ralphbean: Why employ a generator here when we just collapse it with a ``list()`` below?

Also, why a generator here and not in ``fetch_pull_requests``.  I don't mind using one over the other, just so long as its consistent.

bugwarrior/services/bitbucket.py
```diff
@@ -204,8 +204,10 @@ def get_annotations2(self, tag, issue, issue_obj, url):
         )
 
     def get_owner(self, issue):
-        tag, issue = issue
-        return issue.get('responsible', {}).get('username', None)
+        _, issue = issue
+        assignee = issue.get('assignee', None)
+        if assignee is not None:
+            return assignee.get('username', None)
```
ryneeverett: Is this a documented change?

gdetrez: Yes, it's part of the new API

bugwarrior/services/bitbucket.py
gdetrez: Hum... good point ;-)

bugwarrior/services/bitbucket.py
ryneeverett: IMO the oneliner is more readable but I'll leave that to your discretion.

ryneeverett: Link to the new API docs for the curious: https://confluence.atlassian.com/bitbucket/version-2-423626329.html#Version2-Issues

gdetrez: @ryneeverett the behavior is slightly different: the one-liner crashes if `'assignee'` is a valid key in `issue` but the value is `None` (which does happen in the data returned by bitbucket).

bugwarrior/services/bitbucket.py
gdetrez: I think I planned to use only iterators but it turned out to require too much changes to the method `issues`. I reverted to a list in the last commit.

bugwarrior/services/bitbucket.py
ryneeverett: I get it now, thanks!

ryneeverett:
:+1: to merge

ralphbean:
Go Team!  :rocket:

PR #319, opened 2016-05-26
Trello service

gdetrez:
Here's a tentative at a new service to pull tasks from trello.

The idea is that each open card in a board is a task and archiving the card, or moving it out of the included lists closes the task.

It doesn't yet pull comments.


bugwarrior/services/trello.py
```diff
@@ -0,0 +1,159 @@
...
+    def api_request(self, url, **params):
+        """
+        Make a trello API request. This takes an absolute url (without protocol
+        and host) and a list of argumnets and return a GET request with the
+        key and token from the configuration
+        """
+        params['key'] = self.config_get('api_key'),
+        params['token'] = self.config_get('token'),
+        url = "https://api.trello.com" + url
+        return requests.get(url, params=params).json()
```
ryneeverett: Could you add the `ServiceClient` mixin to this class and use it's `json_response` method here? It just keeps the IO error handling consistent across services.

bugwarrior/services/trello.py
```diff
@@ -0,0 +1,159 @@
...
+        Returns a list of dicts representing issues from a remote service.
+        """
+        # First, we get the board name
+        board_id = self.config_get('board')
+        board = self.api_request(
+            "/1/boards/{board_id}".format(board_id=board_id), fields='name')
+        for lst in self.get_lists(board_id):
+            listextra = dict(boardname=board['name'], listname=lst['name'])
+            for card in self.get_cards(lst['id']):
+                yield self.get_issue_for_record(card, extra=listextra)
```
ryneeverett: Would you mind adding a test for this method? These lines appear to be the only ones not covered so far.

mathstuf:
So, this is a Nice To Have™, but it'd be convenient to be able to pull all Trello boards an account can access rather than having to list each board separately. Alternatively, supporting a list of board IDs could make it easier.

gdetrez:
@mathstuf what about making `trello.board` accept a list (renaming it to `trello.boards` or maybe `trello.include_boards` for consistency with the other services) and making it optional so that, if omitted, bugwarrior pulls cards from all bord the user is a member of?

bugwarrior/services/trello.py
gdetrez: Done. I also took the liberty to refactor `json_response` to make it easier to use.

mathstuf:
@gdetrez Yeah, either sounds good. I'd just like to avoid a block per board.

bugwarrior/services/__init__.py
```diff
@@ -481,22 +481,21 @@ def __repr__(self):
...
+        raise IOError(
+            "Non-200 status code %r; %r; %r" % (
+                response.status_code, response.url, response.text,
+            ))
+    if callable(response.json):
+        # Newer python-requests
+        return response.json()
+    else:
+        # Older python-requests
+        return response.json
```
ryneeverett: I agree that it is "too much work to inherit from a mixin for a single static
method" but I disagree with this change.

I think that `ServiceClient` is a pattern starting to emerge in the services and that more opportunities to move logic out of the services and into `ServiceClient` will present themselves over time. If `ServiceClient` had been around for a while and no such opportunities had presented themselves, I would be more amenable to this change. But it hasn't been around very long and I think that the existence of this class will make those opportunities easier to identify and take advantage of.

Even without this DRY aspect, I think it facilitates the standardization of the service structure and makes it easier to make sense of the services that have a standalone `ServiceClient` -- ActiveCollab2, Github, Redmine, and TeamLab. It is not immediately obvious what the purpose of these client classes is, but you can look at their base class, read the docstring, and understand how their purpose is distinct from the other classes.

gdetrez: This is all rather hypothetical. And I think that a common pattern doesn't necessarily imply shared code and shared code doesn't necessarily imply a common parent class.

But I can live with it either way, so if you feel strongly about it tell me and I'll revert the commit.

> you can look at their base class, read the docstring, and understand how their purpose is distinct from the other classes.

In an ideal world, I could just look at the subclass' docstring to understand its purpose, but I understand we don't live in such a world :wink:.

ryneeverett: > a common pattern doesn't necessarily imply shared code and shared code doesn't necessarily imply a common parent class.

I agree with this, which is why I'm now emphasizing the communicative value `ServiceClient` rather than code sharing.

I would like to see further standardization of the service structure and interfaces. The more standardized they are, the less effort it is to jump into a new one.

Python's minimalist design means that classes serve multiple valid purposes. The most straightforward way to express that a class serves a common purpose or interface is a base class, even if that base class doesn't actually do anything.

Several of the services have `ServiceClient`'s whether you want to give them a base class or not. If `ServiceClient` is a good pattern, I think it should have a base class. If it's a bad pattern, we should purge them and roll them into `IssueService`.

> But I can live with it either way, so if you feel strongly about it tell me and I'll revert the commit.

I'd prefer reverting and giving `ServiceClient` a bit more time to prove itself a useful abstraction. Anyone else should feel free to weigh in.

gdetrez: > I'd prefer reverting and giving ServiceClient a bit more time to prove itself a useful abstraction. 
 
Done

gdetrez:
Hello. I'm back :sunglasses:

I refactored my trello branch against the latest develop and added support for pulling comments as annotations. I'm more or less done with the features I planned for this so as soon as you're happy with it, this can now be merged.

bugwarrior/services/trello.py
```diff
@@ -0,0 +1,198 @@
...
+                yield card
+
+    def get_comments(self, card_id):
+        """ Returns an iterator for the comments on a certain card. """
+        params = {'filter': 'commentCard', 'memberCreator_fields': 'username'}
+        comments = self.api_request(
+            "/1/cards/{card_id}/actions".format(card_id=card_id),
+            **params)
+        for comment in comments:
+            assert comment['type'] == 'commentCard'
```
ryneeverett: Did you mean to leave this assertion in? If so, could you explain why it's needed?

gdetrez: I first put it there for my own sake during development but then I figured it didn't hurt to leave it there and it could help diagnose some unexpected API behavior from trello.

ryneeverett: Fair enough.

bugwarrior/docs/services/trello.rst
```diff
@@ -0,0 +1,155 @@
...
+Options
+-------
+
+.. describe:: trello.api_key
+
+   Your Trello API key, available from https://trello.com/app-key
+
+.. describe:: trello.token
+
+   Trello token, see bellow for how to get it.
```
ryneeverett: *below

bugwarrior/docs/services/trello.rst
```diff
@@ -0,0 +1,155 @@
...
+
+.. describe:: trello.include_boards
+
+   The list of board to include. If omitted, bugwarrior will use all board
```
ryneeverett: *boards?

bugwarrior/docs/services/trello.rst
```diff
@@ -0,0 +1,155 @@
...
+   the authenticated user is a member of.
+   This can be either the board ids of the board "short links".  The later is
```
ryneeverett: *latter

bugwarrior/docs/services/trello.rst
```diff
@@ -0,0 +1,155 @@
...
+   ``trello.include_lists``.
+
+.. describe:: trello.exclude_lists
+
+   If set, skip cards from lists whose name is present in
+   ``trello.exclude_lists``.
+
+.. describe:: trello.import_labels_as_tags
+
+   A boolean that indicate whether the Trello labels should be imported as tags
```
ryneeverett: *indicates

ryneeverett:
You might have a look at the (potential) typos first but :+1: to merge regardless.

ralphbean:
Oh, last thing.  Add trello to the ``README`` and then we're good to go.  :+1: for you to merge this @gdetrez.  :)

PR #320, opened 2016-05-28
Github: add an option to skip user issues

### LINKED ISSUE #314: An option for "assigned issues"
In the github service, assigned issues are added to TW unconditionally. Shouldn't they also respect constraint such as `include_repos` and `exclude_repos`? Or maybe there should be an option to have all assigned issues *in addition* to the other pulled issues.

gdetrez:
This is a new option to fix #314. I set the default to be backward compatible so it shouldn't change bugwarrior's behavior for existing users.

### DIFF
--- a/bugwarrior/docs/services/github.rst
+++ b/bugwarrior/docs/services/github.rst
@@ -97,6 +97,12 @@ by adding the following configuration option::
 Get involved issues
 +++++++++++++++++++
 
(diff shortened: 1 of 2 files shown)


ryneeverett:
:+1: to merge

ralphbean:
:+1: here too.

PR #321, opened 2016-05-30
Update docs link to readthedocs.io.

ryneeverett:
See <https://blog.readthedocs.com/securing-subdomains/>:

> Projects will automatically be redirected, and this redirect will remain in place for the foreseeable future. Still, you should plan on updating links to your documentation after the new domain goes live.


ralphbean:
:+1:

PR #323, opened 2016-06-06
Include issues from github private repos. Fix #282

### LINKED ISSUE #282: Private github tasks are not pulled
I've create a token with `repo`, Full control of private repositories, scope. When I access the github API using `curl` I get what I expect, the issues for the private repo I retrieved.

I used that: `curl -u fifigyuri:<token> https://api.github.com/repos/netronixgroup/environet-ui/issues`

And I've got the list of private repo's issues. But bugwarrior does not load these issues for me. This is my config file:

```
[general]
targets = netronix_github
taskrc = ~/.taskrc
shorten = True
annotation_links = True

[netronix_github]
service = github
project_template = {{project|lower}}
github.username = netronixgroup
github.login = fifigyuri
github.token = <token>
github.import_labels_as_tags = True
github.label_template = gh-{{label}}
github.filter_pull_requests = True
```

The public issues I have are pulled, but I'd like to see the private ones only. What went wrong? Where did I make a mistake? Thank you for your help!

ryneeverett:
It seems that the distinction was not properly documented for some time,
but it now is. `/user/repos` includes all repos the authorized user has
access to, but `/users/:username/repos` only includes public repos.
http://stackoverflow.com/questions/21907278/github-api-using-repo-scope-but-still-cant-see-private-repos

### DIFF
--- a/bugwarrior/services/github.py
+++ b/bugwarrior/services/github.py
@@ -13,9 +13,12 @@ class GithubClient(ServiceClient):
     def __init__(self, auth):
         self.auth = auth
 
-    def get_repos(self, username):
-        tmpl = "https://api.github.com/users/{username}/repos?per_page=100"
-        url = tmpl.format(username=username)
+    def get_repos(self, username, public_only=True):
+        url = (
+            "https://api.github.com/users/{username}/repos?per_page=100".format(
+                username=username)
+            if public_only else
+            "https://api.github.com/user/repos?per_page=100")
         return self._getter(url)
 
     def get_involved_issues(self, username):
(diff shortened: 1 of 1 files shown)


gdetrez:
I'm not sure I understand it correctly but it seems that the way `public_only` is set will prevent you from getting issues from a private repositories belonging to an other user or an organization.

Now that I think about it, it seems that there is actually no way to make sure we include all the repositories without systematically requesting both `/user/repos` and `/user/:username/repos`.

ryneeverett:
@gdetrez You're right. I actually realized this back when I was triaging #282 but I couldn't understand my own comment just a minute ago:

>  I think the solution would be to make a call to https://api.github.com/user/repos and then filter out the ones that aren't owned by a github.username.

ryneeverett:
Closing as there's basically nothing right about this.

ryneeverett:
> Now that I think about it, it seems that there is actually no way to make sure we include all the repositories without systematically requesting both /user/repos and /user/:username/repos.

Why would we still need to call /user/:username/repos? User-owned issues should be included in /user/repos.

gdetrez:
> Why would we still need to call /user/:username/repos? User-owned issues should be included in /user/repos.

If one wants to track issues from a (public) repository they are not a member of.

ryneeverett:
Oh, got it.

PR #324, opened 2016-06-06
Include github private repos (take 2). Fix #282.

### LINKED ISSUE #282: Private github tasks are not pulled

ryneeverett:
It seems that the distinction was not properly documented for some time,
but it now is. `/user/repos` includes all repos the authorized user has
access to, but `/users/:username/repos` only includes public repos.
http://stackoverflow.com/questions/21907278/github-api-using-repo-scope-but-still-cant-see-private-repos

**Note**: This doesn't break anything for me but I'm not super familiar with the configuration options and am not sure if this might accidentally change their behavior.

### DIFF
--- a/bugwarrior/services/github.py
+++ b/bugwarrior/services/github.py
@@ -14,9 +14,12 @@ class GithubClient(ServiceClient):
         self.auth = auth
 
     def get_repos(self, username):
-        tmpl = "https://api.github.com/users/{username}/repos?per_page=100"
-        url = tmpl.format(username=username)
-        return self._getter(url)
+        user_repos = self._getter(
+            "https://api.github.com/user/repos?per_page=100")
+        public_repos = self._getter(
+            "https://api.github.com/users/{username}/repos?per_page=100".format(
+                username=username))
+        return user_repos + public_repos
 
     def get_involved_issues(self, username):
         tmpl = "https://api.github.com/search/issues?q=involves%3A{username}&per_page=100"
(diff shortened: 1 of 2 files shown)


ryneeverett:
As @gdetrez pointed out this is still not quite right as it doesn't allow pulling public repos you don't own.

gdetrez:
It looks fine to me now :+1:

ryneeverett:
Cool. It would be nice is somebody with private github repos confirmed that this works before merging.

ralphbean:
I would, but I don't have any private repos myself.  :/

ralphbean:
I tried soliciting testers from some private-repo-users I know, but no one bit.

Let's go ahead and merge this and then see if anything pops up.

ryneeverett:
Fair enough. I've been using this with public repos without issue.

PR #325, opened 2016-06-13
Improve JIRA integration

joshainglis:
Include Fix Version
Prevent Jira from logging user out of web client
Get ticket creation time.

### DIFF
--- a/bugwarrior/services/jira.py
+++ b/bugwarrior/services/jira.py
@@ -1,11 +1,40 @@
 from __future__ import absolute_import
 
-from jinja2 import Template
-from jira.client import JIRA
+from datetime import datetime
+
 import six
+from jinja2 import Template
+from jira.client import JIRA as BaseJIRA
+from requests.cookies import RequestsCookieJar
 
 from bugwarrior.config import asbool, die
 from bugwarrior.services import IssueService, Issue
+
+# The below `ObliviousCookieJar` and `JIRA` classes are MIT Licenced.
+# They were taken from this wonderful commit by @GaretJax
+# https://github.com/GaretJax/lancet/commit/f175cb2ec9a2135fb78188cf0b9f621b51d88977
+# Prevents Jira web client being logged out when API call is made.
+class ObliviousCookieJar(RequestsCookieJar):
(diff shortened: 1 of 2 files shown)


ryneeverett:
I think fixing the tests would be a matter of adding a reasonable `created` value to the `arbitrary_record['fields']` we're testing with [here](https://github.com/ralphbean/bugwarrior/blob/develop/tests/test_jira.py#L34).

joshainglis:
Hi @ryneeverett,
Added new fields into the test suite. All passes now :)

ryneeverett:
Thanks! Code looks fine to me.

I don't know how others might feel about copying code from another (MIT Licensed) project. We should probably technically copy in his license somewhere but I don't know if anyone really cares. I'm going to leave this open for others to weigh in.

bugwarrior/services/jira.py
```diff
@@ -1,19 +1,49 @@
...
+        pass
+
+    def copy(self):
+        """Make sure to return an instance of the correct class on copying."""
+        return ObliviousCookieJar()
+
+
+# 'Borrowed' from this wonderful commit by @GaretJax
+# https://github.com/GaretJax/lancet/commit/f175cb2ec9a2135fb78188cf0b9f621b51d88977
+# Prevents Jira web client being logged out when API call is made.
```
ralphbean: Can you add one more inline comment here that mentions the snippet is licensed MIT.

I know that's unorthodox, but the code is small enough that I feel comfortable including it in our larger GPL file.

You *can* include MIT code in a GPL project, but you typically put an explicit license header on the MIT *file* indicating the license and the copyright holder.

Here you link to the original code (good) and mention the author (good).  Mentioning the original license should be enough.

ralphbean:
I left one comment about the license on that snippet.

Thanks! This looks great (and I can't wait to use it...).

joshainglis:
Thanks for the feedback @ryneeverett & @ralphbean.
I've added an inline comment noting the MIT Licence on the `ObliviousCookieJar` and `Jira` classes.

ralphbean:
Awesome!

PR #333, opened 2016-06-25
Kill the --debug option.

ralphbean:
It is confusing when there is already a ``log.level = DEBUG`` option in
the config which has nothing to do with **this** debug option.

I think we added this flag back when we first added multiprocessing
support as a mechanism for devs to debug that stuff.. but it's been in
place and working well now for a long time.  Let's kill this.

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -31,8 +31,7 @@ def _get_section_name(flavor):
 @click.option('--dry-run', is_flag=True)
 @click.option('--flavor', default=None, help='The flavor to use')
 @click.option('--interactive', is_flag=True)
-@click.option('--debug', is_flag=True)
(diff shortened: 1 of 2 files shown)


ryneeverett:
I can't confirm at the moment but my recollection is that without using the `--debug` flag to make the pull single-processed `pdb` beakpoints don't work.

ryneeverett:
Yeah, breakpoints are skipped over when multiprocessing.

ralphbean:
Nice catch.  Thanks @ryneeverett.  :)

coddingtonbear:
Ooo, sorry for not noticing you guys having this conversation.  I'm fairly sure that I added that to make it easier to debug in exactly the way that @ryneeverett guessed.

PR #346, opened 2016-07-05
Modify headers instead of overwriting.

### LINKED ISSUE #345: Github missing User-Agent header
This regression was introduced in #332. Since I'm the first to run into it it may only occur using oauth authentication.

```console
ERROR:bugwarrior.services:Worker for [my_github] failed: Non-200 status code 403; u'https://api.github.com/user/repos?per_page=100'; u'Request forbidden by administrative rules. Please make sure your request has a User-Agent header (http://developer.github.com/v3/#user-agent-required). Check https://developer.github.com for other possible causes.\n'
  File "bugwarrior/bugwarrior/services/__init__.py", line 515, in _aggregate_issues
  File "bugwarrior/bugwarrior/services/github.py", line 360, in issues
    all_repos = self.client.get_repos(self.username)
  File "bugwarrior/bugwarrior/services/github.py", line 24, in get_repos
    "https://api.github.com/user/repos?per_page=100")
  File "bugwarrior/bugwarrior/services/github.py", line 73, in _getter
    json_res = self.json_response(response)
  File "bugwarrior/bugwarrior/services/__init__.py", line 494, in json_response
    response.status_code, response.url, response.text,
IOError: Non-200 status code 403; u'https://api.github.com/user/repos?per_page=100'; u'Request forbidden by administrative rules. Please make sure your request has a User-Agent header (http://developer.github.com/v3/#user-agent-required). Check https://developer.github.com for other possible causes.\n'
```

ralphbean:
There are some default headers attached to requests sessions:

```python
>>> import requests
>>> s = requests.session()
>>> print(s.headers)
{
    'Connection': 'keep-alive',
    'Accept-Encoding': 'gzip, deflate',
    'Accept': '*/*',
    'User-Agent': 'python-requests/2.10.0',
}
```

In #332 we accidentally started overwriting all of those defaults.
This change brings them back into play and should fix #345.

### DIFF
--- a/bugwarrior/services/github.py
+++ b/bugwarrior/services/github.py
@@ -17,7 +17,7 @@ class GithubClient(ServiceClient):
         self.session = requests.Session()
         if 'token' in self.auth:
             authorization = 'token ' + self.auth['token']
-            self.session.headers = {'Authorization': authorization}
+            self.session.headers['Authorization'] = authorization
 
     def get_repos(self, username):
         user_repos = self._getter(


ryneeverett:
Nice catch, just tested and this is indeed the fix.

PR #348, opened 2016-07-07
Debian BTS support

### LINKED ISSUE #327: Support for Debian BTS
The Debian bug tracker (https://www.debian.org/Bugs/) is the primary place where Debian bugs are tracked.

There is a SOAP interface "documented" at:

https://wiki.debian.org/DebbugsSoapInterface

I am planning to package this software in Debian (Debian bug is https://bugs.debian.org/810629) and it would be awesome to also use this for Debian bugs.

irl:
This pull request adds a service ```bts``` for the Debian Bug Tracking System (BTS) that is used to track bugs on packages distributed as part of the Debian Operating System.

This pull request closes #327.

### DIFF
--- a/bugwarrior/docs/services/bts.rst
+++ b/bugwarrior/docs/services/bts.rst
@@ -58,6 +58,20 @@ feature, you can add this line to your service configuration::
 
     bts.udd = True
 
+Excluding bugs marked pending
++++++++++++++++++++++++++++++
+
+Debian bugs are not considered closed until the fixed package is present in the
+Debian archive. Bugs do cease to be outstanding tasks however as soon as you have
+completed the work, and so it can be useful to exclude bugs that you have marked
+with the pending tag in the BTS.
+
+This is the default behaviour, but if you feel you would like to include bugs that
+are marked as pending in the BTS, you can disable this by adding this line to your
+service configuration::
+
+    bts.ignore_pending = False
+
 Excluding sponsored and NMU'd packages
 ++++++++++++++++++++++++++++++++++++++
 
@@ -106,5 +120,7 @@ Provided UDA Fields
 | ``btspackage``      | Binary Package      | Text (string)       |
 +---------------------+---------------------+---------------------+
 | ``btsforwarded``    | Forwarded URL       | Text (string)       |
++---------------------+---------------------+---------------------+
+| ``btsstatus``       | Status              | Text (string)       |
 +---------------------+---------------------+---------------------+
 
--- a/bugwarrior/docs/services/bts.rst
+++ b/bugwarrior/docs/services/bts.rst
@@ -1,11 +1,20 @@
 Debian Bug Tracking System (BTS)
 ================================
 
-You can import tasks from your {{ service_name_humane }} instance using
-the ``{{ service_name }}`` service name.
+You can import tasks from the Debian Bug Tracking System (BTS) using
+the ``bts`` service name. Debian's bugs are public and no authentication
+information is required by bugwarrior for this service.
 
-Instructions
-------------
+Additional Requirements
+-----------------------
+
+You will need to install the following additional packages via ``pip``:
+
+ * ``PySimpleSOAP``
+ * ``python-debianbts``
+
+.. note:: If you have installed the Debian package for bugwarrior, this
+          dependency will already be satisfied.
 
 Example Service
 ---------------
@@ -28,14 +37,12 @@ Include all bugs for packages
 +++++++++++++++++++++++++++++
 
 If you would like more bugs than just those you are the owner of, you can specify
-the ```bts.packages``` option.
+the ``bts.packages`` option.
 
-For example if you wanted to include bugs on the ```hello``` package, you can add
-this line to your service configuration:
+For example if you wanted to include bugs on the ``hello`` package, you can add
+this line to your service configuration::
 
-```
-bts.packages = hello
-```
+    bts.packages = hello
 
 More packages can be specified seperated by commas.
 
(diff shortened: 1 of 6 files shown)


ryneeverett:
Nice. Also, don't forget to add debianbts as an extra dependency in `extras_require` in setup.py.

irl:
Excluding a Unicode bug, I'm quite happy with the actual functionality of this. Review would be useful.

Will work on tests/docs next.

irl:
The Unicode issue turned out to be #350 and not a bug in this service.

irl:
In efforts to try to break this, I loaded in all the bugs from the Debian Python Modules Team (639 bugs) and all the bugs from the Debian QA Team (2067 bugs). Other than the Unicode logging issue #350 and a few occurrences of [TW-1805](https://bug.tasktools.org/browse/TW-1805) which has been resolved for the next task release already, everything works!

Review and/or merge please.

bugwarrior/services/bts.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+
+
+class BTSService(IssueService, ServiceClient):
+    ISSUE_CLASS = BTSIssue
+    CONFIG_PREFIX = 'bts'
+
+    def __init__(self, *args, **kw):
+        super(BTSService, self).__init__(*args, **kw)
+        self.bts = debianbts
```
ryneeverett: This looks a lot like the pattern found in other services in which a client is instantiated and assigned as a class attribute. However, I find assigning a library as a class attribute like this odd as it's always available in the module scope.

irl: Needs to be replaceable by the tests on a per-instance basis.

bugwarrior/services/bts.py
```diff
@@ -0,0 +1,175 @@
...
+        self.ignore_src = self.config_get_default('ignore_src', default=None)
+
+    @classmethod
+    def validate_config(cls, config, target):
+#        if not config.has_option(target, 'trac.base_uri'):
+#            die("[%s] has no 'base_uri'" % target)
+#        elif '://' in config.get(target, 'trac.base_uri'):
+#            die("[%s] do not include scheme in 'base_uri'" % target)
+
+        IssueService.validate_config(config, target)
```
ryneeverett: What's going on here? Can this method be deleted?

irl: Wow. Can't believe I missed that. I left that commented out to remind me to fill it in.

It essentially will just verify that bts.email is set to a valid email address, and that the true/false options can be parsed as such.

Setting back to [WIP].

bugwarrior/services/bts.py
ryneeverett: I see. I just looked through the services and it looks like every single service which uses a client library has to instantiate it, so this service presents a new case. I think mocking the library in the tests (with `mock.patch`) would be a more straightforward way of handling this.

irl: Ok, I've not done that before, should be fun to learn. (:

ryneeverett: There are examples in several of the test modules. I think it would be `with mock.patch('debianbts', FakeBtsLib):`.

irl: Oh cool, that's really easy actually. I learnt a thing. Thanks!

bugwarrior/docs/services/bts.rst
```diff
@@ -0,0 +1,110 @@
...
+          dependency will already be satisfied.
+
+Example Service
+---------------
+
+Here's an example of a Debian BTS target::
+
+    [debian_bts]
+    service = bts
+    email = username@debian.org
```
boutil: shouldn't this be `bts.email = username@debian.org` instead?

irl: Yes, it should be.

bugwarrior/services/bts.py
irl: I can't get this to work at all. ):

I think the problem is that I'm trying to patch a module, not a class.

ryneeverett: My bad, I think the target is relative to the module in which the patching is occurring. So it would be `with mock.patch('bts.debianbts', FakeBtsLib):`

irl:
```
def get_mock_service(self, *args, **kwargs):
    with mock.patch('bugwarrior.services.bts.debianbts', FakeBTSLib):
            service = super(TestBTSService, self).get_mock_service(
                *args, **kwargs)
    return service
```

It runs, but the patch is entirely ignored.

irl:
```
def get_mock_service(self, *args, **kwargs):
    bts.debianbts = FakeBTSLib()
    service = super(TestBTSService, self).get_mock_service(
        *args, **kwargs)
    return service
```

How about this? Replacing it in the scope of the service module but not having the service as an attribute of the service class.

irl:
Rebased on the develop branch, fixed up code style. I think there's everything here.

irl:
Review and/or merge please. (:

ryneeverett:
I made a PR against your branch.

irl:
Thanks @ryneeverett, I see that was easier than I was making it.

ryneeverett:
No problem @irl, I always forget how mock targeting works.

ryneeverett:
Code looks good to me. The one thing I haven't looked into is the consistency of the configuration options compared with the other services. I'm not all that familiar with the extra options so I'd have to look into it but maybe somebody else can evaluate better.

irl:
I think the way that the BTS works is not very consistent with other bug trackers. Bugs are assigned to packages, only on psuedo packages are bugs regularly assigned to people. I was thinking about this, and I'll think about it a little more, but I'm very happy with the way the options currently work.

ralphbean:
This looks good to me too.  Thanks so much for cycling on it @irl, @ryneeverett.

Let's merge it.  :)

PR #349, opened 2016-07-07
Make Jira installation example in docs ZSH compatible

ubuntudroid:
In ZSH running `pip install bugwarrior[jira] doesn't work as the shell will try to interpret bugwarrior[jira] as a command. Therefore I've added a hint on how to install the Jira plugin on ZSH which requires adding quotation marks around the installation target.

### DIFF
--- a/bugwarrior/docs/getting.rst
+++ b/bugwarrior/docs/getting.rst
@@ -20,10 +20,6 @@ and Active Collab but those require extra dependencies that are installed by
 specifying ``bugwarrior[service]`` in the commands above. For example, if you
 want to use bugwarrior with Jira::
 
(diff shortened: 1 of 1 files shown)


ryneeverett:
Wouldn't the quotes syntax work in any shell? Could we just add quotes to the existing suggested command rather than add a zsh-specific syntax?

ubuntudroid:
Thought so as well, but wasn't sure. That's why I went for the explicit distinction.

ryneeverett:
Just tested in bash and the quotes aren't a problem.

ubuntudroid:
Unifed the command for all shells.

ryneeverett:
Thanks!

PR #353, opened 2016-07-09
Make bitbucketid numeric.

ryneeverett:
This silences all the warnings from taskw that a number is being coerced
into a string.

I'm not sure what the ramifications of this might be. Looking at the json files, this doesn't appear to have caused an automatic migration or duplication among existing tasks. Any thoughts on how to proceed would be appreciated.

### DIFF
--- a/bugwarrior/services/bitbucket.py
+++ b/bugwarrior/services/bitbucket.py
@@ -24,7 +24,7 @@ class BitbucketIssue(Issue):
             'label': 'Bitbucket URL',
         },
         FOREIGN_ID: {
-            'type': 'string',
+            'type': 'numeric',
             'label': 'Bitbucket Issue ID',
         }
     }


coddingtonbear:
No danger here, as far as I'm aware!  I think it might possibly cause affected tasks to appear to be changed were we to re-save those tasks, but I wouldn't call that a big deal.

ryneeverett:
Yes, this indeed does seem to work fine. In fact, aside from silencing the warnings, this doesn't seem to actually change anything. The odd thing is, the bitbucket_id is still a string in the json, even in new tasks. Does taskwarrior really store numeric-typed data as a string?

coddingtonbear:
It appears so; just glancing at the data in my `pending.data`, it looks like all fields are quoted as strings.  I guess taskwarrior just uses the data type as a guideline for how to interpret the data.

ryneeverett:
Thanks for confirming.

PR #357, opened 2016-07-25
Catch IOError when loading config, useful message

### LINKED ISSUE #354: IOError when starting bugwarrior-uda without a config file
From Debian bug 830833 (http://bugs.debian.org/830833):

```
Dear Maintainer,

the documentation says:

> For using this data in reports, it is recommended that you add these UDA
> definitions to your ~/.taskrc file. You can generate your list of UDA
> definitions by running the following command:
>
>     bugwarrior-uda

I was curious, so I ran `bugwarrior-uda`. But I did not create a config file
first, which seems to bother it:

	$ bugwarrior-uda
		File "/usr/bin/bugwarrior-uda", line 9, in <module>
			load_entry_point('bugwarrior==1.4.0', 'console_scripts', 'bugwarrior-uda')()
		File "/usr/lib/python2.7/dist-packages/click/core.py", line 716, in __call__
			return self.main(*args, **kwargs)
		File "/usr/lib/python2.7/dist-packages/click/core.py", line 696, in main
			rv = self.invoke(ctx)
		File "/usr/lib/python2.7/dist-packages/click/core.py", line 889, in invoke
			return ctx.invoke(self.callback, **ctx.params)
		File "/usr/lib/python2.7/dist-packages/click/core.py", line 534, in invoke
			return callback(*args, **kwargs)
		File "/usr/lib/python2.7/dist-packages/bugwarrior/command.py", line 138, in uda
			conf = load_config(main_section)
		File "/usr/lib/python2.7/dist-packages/bugwarrior/config.py", line 169, in load_config
			"utf-8",
		File "/usr/lib/python2.7/codecs.py", line 896, in open
			file = __builtin__.open(filename, mode, buffering)
	IOError: [Errno 2] No such file or directory: '/home/rohieb/.config/bugwarrior/bugwarriorrc'

A clearer message about a missing config file would be more appropriate here.

This feature is probably suited to get familiar with the code base, which is the
reason why I tagged this bug report with "newcomer".

Cheers,

 - Roland
```

irl:
Fixes #354

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -74,8 +74,8 @@ def pull(dry_run, flavor, interactive, debug):
         )
     except RuntimeError as e:
         log.critical("Aborted (%s)" % e)
-    except SystemExit:
-        pass
+    except SystemExit as e:
+        sys.exit(e.code)
     except:
         log.exception('oh noes')
 
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -1,4 +1,5 @@
 import os
+import sys
 
 from lockfile import LockTimeout
 from lockfile.pidlockfile import PIDLockFile
@@ -15,6 +16,7 @@ from bugwarrior.db import (
 )
 
 import logging
+logging.basicConfig()
 log = logging.getLogger(__name__)
 
 
@@ -44,7 +46,12 @@ def pull(dry_run, flavor, interactive, debug):
         main_section = _get_section_name(flavor)
 
         # Load our config file
-        config = load_config(main_section, interactive)
+        try:
+            config = load_config(main_section, interactive)
+        except IOError:
+            log.critical("Could not load configuration. "
+                         "Maybe you have not created a configuration file.")
+            sys.exit(1)
 
         lockfile_path = os.path.join(get_data_path(), 'bugwarrior.lockfile')
         lockfile = PIDLockFile(lockfile_path)
@@ -67,6 +74,8 @@ def pull(dry_run, flavor, interactive, debug):
         )
     except RuntimeError as e:
         log.critical("Aborted (%s)" % e)
+    except SystemExit:
+        pass
     except:
         log.exception('oh noes')
 
@@ -136,7 +145,12 @@ def set(target, username):
 @click.option('--flavor', default=None, help='The flavor to use')
 def uda(flavor):
     main_section = _get_section_name(flavor)
-    conf = load_config(main_section)
+    try:
+        conf = load_config(main_section)
+    except IOError:
+        log.critical("Could not load configuration. "
+                     "Maybe you have not created a configuration file.")
+        sys.exit(1)
     print "# Bugwarrior UDAs"
     for uda in get_defined_udas_as_strings(conf, main_section):
         print uda


bugwarrior/command.py
```diff
@@ -67,6 +74,8 @@ def pull(dry_run, flavor, interactive, debug):
         )
     except RuntimeError as e:
         log.critical("Aborted (%s)" % e)
+    except SystemExit:
+        pass
```
ryneeverett: I think this defeats the purpose of `sys.exit(1)` because when you `pass` on the exception the exit code ends up being 0. Why is this necessary?

irl: Ooops. This was as otherwise the generic catch all exceptions triggers and you get a traceback which clouds the helpful error message.

ryneeverett: I figured that was the rationale. Is there a way we can suppress the traceback while still throwing a bad exit code?

bugwarrior/command.py
```diff
@@ -74,8 +74,8 @@ def pull(dry_run, flavor, interactive, debug):
         )
     except RuntimeError as e:
         log.critical("Aborted (%s)" % e)
-    except SystemExit:
-        pass
+    except SystemExit as e:
+        sys.exit(e.code)
```
irl: @ryneeverett I think this is the cleaner solution.

bugwarrior/command.py
```diff
@@ -15,6 +16,7 @@
 )
 
 import logging
+logging.basicConfig()
```
ryneeverett: Why is this now needed?

ryneeverett:
Why not catch the exception in bugwarrior/config.py's `load_config` and avoid the duplication?

bugwarrior/command.py
irl: Without this I'm seeing "no handler available for logger". This is not needed as a result of this change, just none of the log calls were showing anything for some reason. This may be a regression from #331 or may be a result of my set up.

ryneeverett:
Nevermind, I guess `load_config` is used internally and we wouldn't want to suppress those exceptions. So why not define a function `_load_config` in this module that does the exception suppression?

bugwarrior/command.py
ryneeverett: I'm seeing the same behavior and I'm at a loss for why that is. All the other logging seems to work fine.

ryneeverett: Oh I see. It's run by `bugwarrior.config.validate_config` which isn't called until the end of `bugwarrior.config.load_config` -- *after* the exception you're catching. Rather than set up logging twice I think we need to move the logging setup in `validate_config` to earlier to the entry point.

ryneeverett: Hmm, well we can't quite do that because it depends on the logging settings in the configuration!

irl:
This isn't something I'm going to be able to look at much more this week, @ryneeverett or @ralphbean feel free to close this pull request and take a more elegant stab at it.

ryneeverett:
I made a PR against @irl's branch.

bugwarrior/command.py
```diff
@@ -28,6 +29,19 @@ def _get_section_name(flavor):
...
 
+def _try_load_config(main_section, interactive):
+    try:
+        return load_config(main_section, interactive)
+    except IOError:
+        # Our standard logging configuration depends on the bugwarrior
+        # configuration file which just failed to load.
+        logging.basicConfig()
+
+        log.critical("Could not load configuration. "
```
ralphbean: Should this be ``log.exception`` instead?  As it stands, the reason for the ``IOError`` is hidden.  Did it fail due to permissions?  Or due to a non-existent file?

ryneeverett: I attempted to address this in [irl#3](https://github.com/irl/bugwarrior/pull/3).

bugwarrior/command.py
```diff
@@ -34,6 +33,10 @@ def _try_load_config(main_section, interactive):
     try:
         return load_config(main_section, interactive)
     except IOError:
+        # Our standard logging configuration depends on the bugwarrior
+        # configuration file which just failed to load.
+        logging.basicConfig()
```
irl: I like this.

ryneeverett:
No comments for a week so I'd say go ahead and merge this.

PR #362, opened 2016-08-03
Gitlab todos

mathstuf:
---
Pushed for initial review if you're inerested; plan on testing it against our gitlab instance tomorrow at work.


mathstuf:
OK, tested now :) .

mathstuf:
@ralphbean What's up with the test failures there? Did I miss something?

bugwarrior/services/gitlab.py
```diff
@@ -336,13 +386,38 @@ def get_repo_merge_requests(self, rid):
...
                 continue
             issues[issue['id']] = (rid, issue)
         return issues
 
+    def get_todos(self):
+        tmpl = '{scheme}://{host}/api/v3/todos'
+        todos = {}
+        try:
+            fetched_todos = self._fetch_paged(tmpl)
+        except:
```
ryneeverett: Can we catch a specific exception here?

tests/test_gitlab.py
```diff
@@ -54,7 +58,7 @@ class TestGitlabIssue(AbstractServiceTest, ServiceTest):
             "id": 1,
             "title": "v1.0",
             "description": "",
-            "due_date": "2012-07-20",
+            "due_date": arbitrary_duedate.date().isoformat(),
```
ryneeverett: I think the `date()` part is the main problem with the tests. It converts type from `datetime` to `date` which causes it to lose it's timezone awareness.

bugwarrior/services/gitlab.py
mathstuf: `HTTPError`? Not sure that that is comprehensive though.

tests/test_gitlab.py
mathstuf: Well, the actual format doesn't have the time associated with it… I guess the internal "it's a date, just put the time on it too" logic is doing this. Taskwarrior only has a single "datetime" type, so I'm not sure how best to fix this.

mathstuf:
OK, I did indeed miss something in the tests: there was no expected `gitlabduedate` field in the last dictionary. Adding it there makes things better :) .

bugwarrior/services/gitlab.py
```diff
@@ -332,6 +332,7 @@ def _fetch(self, tmpl, **kwargs):
         if not self.verify_ssl:
             requests.packages.urllib3.disable_warnings()
         response = requests.get(url, headers=headers, verify=self.verify_ssl, **kwargs)
+        response.raise_for_status()
```
ryneeverett: Is there any advantage to this approach over catching the `IOError` that would be raised by `json_response` (https://github.com/ralphbean/bugwarrior/blob/develop/bugwarrior/services/__init__.py#L489)? If you prefer this approach, could you add a comment to this line like `# raise HTTPError`? Otherwise it wont be obvious what's going on here without browsing commit history.

bugwarrior/services/gitlab.py
```diff
@@ -320,9 +371,9 @@ def get_repo_issues(self, rid):
         issues = {}
         try:
             repo_issues = self._fetch_paged(tmpl)
-        except:
+        except HTTPError:
```
ryneeverett: Pretty sure this isn't a built-in and needs to be imported.

bugwarrior/services/gitlab.py
mathstuf: Ah, yeah, that makes sense.

ralphbean:
:+1: to merge from me.

PR #363, opened 2016-08-03
Taskdata dir

mathstuf:


ryneeverett:
What is the use case where the environment variable is inadequate?

bugwarrior/config.py
```diff
@@ -185,7 +185,12 @@ def get_taskrc_path(conf, main_section):
 
 
 def get_data_path():
-    return os.path.expanduser(os.getenv('TASKDATA', '~/.task'))
+    path = os.getenv('TASKDATA', '~/.task')
+    if conf.has_option(main_section, 'taskdata'):
```
ryneeverett: `conf` -> `config`

bugwarrior/config.py
```diff
@@ -185,7 +185,12 @@ def get_taskrc_path(conf, main_section):
...
+        path = conf.get(main_section, 'taskdata')
```
ryneeverett: s/conf/config/g

mathstuf: It is `conf` up in `get_taskrc_path`. Is that wrong?

mathstuf:
I can already set the `taskrc` in a per-flavor stanza; it'd be nice to handle the directory as well. Though I suppose asking `task` what it thinks the directory is would be the best route since that would just use the `taskrc`'s setting (and do the `TASKDATA` fallback that way).

mathstuf:
Yeah, `task _show | grep '^data.location='` will give the location that task uses for its lists including `TASKDATA` in the computation.

bugwarrior/config.py
ryneeverett: That's a function parameter which doesn't exist in this function. Travis tests failed due to `NameError: global name 'conf' is not defined`.

mathstuf: Oh, missed that it's passed in as an argument above. Probably mis-copied :/ . Anyways, not necessary for `task _show`.

ryneeverett: I guess the config needs to be passed into this function for this to work.

mathstuf:
Well, this is as far as I got tonight. Currently, the bitbucket test hangs for whatever reason, but I don't have pudb handy here and with all the fighting Python I did already today, I don't think being in one would be productive this late at night anyways.

There's probably a better design, but I've been doing Rust for the past week and Python feels…squishy now and I can't seem to get things quite right. I'll take another look tomorrow when I'll hopefully be able to see a better way of doing this.

ryneeverett:
I took my own crack at this in #364.

mathstuf:
Thanks; closing in favor of it. Much cleaner :) .

PR #364, opened 2016-08-06
Get data location from taskwarrior.

ryneeverett:
Since we support multiple config flavors, we need to ask taskwarrior
(given the current flavor's taskrc) what the data location is.

Thanks to @mathstuf for taking the lead on this. I copied some of the
code from #363 and tried for a cleaner implementation of his idea.

# TODO

- [x] Move `task _show` calling/parsing to taskw.
- [x] Change data file mode to 0600.
- [x] Test using `TASKDATA`.

### DIFF
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -220,10 +220,7 @@ def get_data_path(config, main_section):
     data_path = data_location[len(line_prefix):].rstrip()
 
     if not data_path:
-        data_path = os.getenv('TASKDATA', '~/.task')
-        log.warning(
-            'Unable to determine the data location. Falling back to %s.' %
-            data_path)
+        raise RuntimeError('Unable to determine the data location.')
 
     return os.path.normpath(os.path.expanduser(data_path))
 
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -213,7 +213,9 @@ def get_data_path(config, main_section):
     # the `_` subcommands properly (`rc:` can't be used for them).
     line_prefix = 'data.location='
     tw_show = subprocess.Popen(
-        ('task', '_show'), stdout=subprocess.PIPE, env={'TASKRC': taskrc})
+        ('task', '_show'),
+        stdout=subprocess.PIPE,
+        env={'PATH': os.getenv('PATH'), 'TASKRC': taskrc})
     data_location = subprocess.check_output(
         ('grep', '-e', '^' + line_prefix), stdin=tw_show.stdout)
     tw_show.wait()
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -215,7 +215,7 @@ def get_data_path(config, main_section):
     tw_show = subprocess.Popen(
         ('task', '_show'), stdout=subprocess.PIPE, env={'TASKRC': taskrc})
     data_location = subprocess.check_output(
-        ('grep', line_prefix), stdin=tw_show.stdout)
+        ('grep', '-e', '^' + line_prefix), stdin=tw_show.stdout)
     tw_show.wait()
     data_path = data_location[len(line_prefix):].rstrip()
 
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import sys
 
+from functools32 import lru_cache
 import six
 
 import logging
@@ -204,9 +205,27 @@ def get_taskrc_path(conf, main_section):
     )
 
 
-def get_data_path():
-    return os.path.expanduser(os.getenv('TASKDATA', '~/.task'))
+@lru_cache()
+def get_data_path(config, main_section):
+    taskrc = get_taskrc_path(config, main_section)
 
+    # We cannot use the taskw module here because it doesn't really support
+    # the `_` subcommands properly (`rc:` can't be used for them).
+    line_prefix = 'data.location='
+    tw_show = subprocess.Popen(
+        ('task', '_show'), stdout=subprocess.PIPE, env={'TASKRC': taskrc})
+    data_location = subprocess.check_output(
+        ('grep', line_prefix), stdin=tw_show.stdout)
+    tw_show.wait()
+    data_path = data_location[len(line_prefix):].rstrip()
+
+    if not data_path:
+        data_path = os.getenv('TASKDATA', '~/.task')
+        log.warning(
+            'Unable to determine the data location. Falling back to %s.' %
+            data_path)
+
+    return os.path.normpath(os.path.expanduser(data_path))
 
 # This needs to be imported here and not above to avoid a circular-import.
 from bugwarrior.services import get_service
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -212,10 +212,18 @@ def get_data_path(config, main_section):
     # We cannot use the taskw module here because it doesn't really support
     # the `_` subcommands properly (`rc:` can't be used for them).
     line_prefix = 'data.location='
+
+    env={
+        'PATH': os.getenv('PATH'),
+        'TASKRC': taskrc,
+    }
+    # If TASKDATA is set but empty, taskwarrior's data.location is empty.
+    taskdata = os.getenv('TASKDATA')
+    if taskdata:
+        env['TASKDATA'] = taskdata
+
     tw_show = subprocess.Popen(
-        ('task', '_show'),
-        stdout=subprocess.PIPE,
-        env={'PATH': os.getenv('PATH'), 'TASKRC': taskrc})
+        ('task', '_show'), stdout=subprocess.PIPE, env=env)
     data_location = subprocess.check_output(
         ('grep', '-e', '^' + line_prefix), stdin=tw_show.stdout)
     tw_show.wait()
--- a/tests/base.py
+++ b/tests/base.py
@@ -1,4 +1,8 @@
 import mock
+import shutil
+import os
+import os.path
+import tempfile
 import unittest
 
 import responses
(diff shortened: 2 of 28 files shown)


bugwarrior/config.py
```diff
@@ -184,9 +185,27 @@ def get_taskrc_path(conf, main_section):
...
+def get_data_path(config, main_section):
+    taskrc = get_taskrc_path(config, main_section)
 
+    # We cannot use the taskw module here because it doesn't really support
+    # the `_` subcommands properly (`rc:` can't be used for them).
+    line_prefix = 'data.location='
+    tw_show = subprocess.Popen(
+        ('task', '_show'), stdout=subprocess.PIPE, env={'TASKRC': taskrc})
+    data_location = subprocess.check_output(
+        ('grep', line_prefix), stdin=tw_show.stdout)
```
mathstuf: Probably want to anchor the search pattern with `^` here. I also like passing `-e`, but that's me.

bugwarrior/config.py
```diff
@@ -184,9 +185,27 @@ def get_taskrc_path(conf, main_section):
...
+    tw_show.wait()
+    data_path = data_location[len(line_prefix):].rstrip()
+
+    if not data_path:
+        data_path = os.getenv('TASKDATA', '~/.task')
+        log.warning(
+            'Unable to determine the data location. Falling back to %s.' %
+            data_path)
```
mathstuf: I'm not so sure it should fall back. Using the wrong data list is not really something to play around with lightly. If `data.location` isn't set, either `task` has changed quite a bit out from underneath us or things are already in quite a deep hole.

bugwarrior/config.py
ryneeverett: Good call.

bugwarrior/config.py
ryneeverett: I was thinking the more likely scenario than `data.location` not being set would be that either the `task` binary isn't in the `PATH` or `taskwarrior _show`'s output changes and breaks our regex. Perhaps it would be better to just break in those cases than to screw up the data.

mathstuf: If `task` isn't in the path, how are any other `taskw.TaskWarriorShellout` commands going to work? Bugwarrior doesn't offer any other way to find the binary.

The `_show` output is part of task's shell completion support; I doubt it will change that much since it's likely to stay simple.

ryneeverett: Fair enough.

ryneeverett:
I'm mystified by the travis failures. The tests run fine locally.

ryneeverett:
Removing `env={'TASKRC': taskrc}` seems to fix the build.

ryneeverett:
This is crazy. So for some reason `os.pipe()` is returning integers rather than file descriptors which is raising the "No such file or directory" exceptions when it tries to open them. It gets worse. **This only happens on Travis' Ubuntu distribution!** The tests run fine in a vanilla ubuntu docker container, but exhibit this failure on travis' ubuntu image. Argh!

ralphbean:
:+1: in general to the change.  As for the tests, it's not worth agonizing over for travis.  Would it be easier to just mock out ``get_data_path`` elsewhere?

tests/base.py
```diff
@@ -64,6 +70,23 @@ def get_int(section, name):
 
         return service
 
+    @classmethod
+    def setUpClass(cls):
+        cls.tempdir = tempfile.mkdtemp(prefix='bugwarrior')
+        cls.taskrc = os.path.join(cls.tempdir, 'taskrc')
+        lists_path = os.path.join(cls.tempdir, 'lists')
+        os.mkdir(lists_path)
```
mathstuf: D'oh.

ryneeverett:
@gdetrez might take a look given the overlap with #370.

bugwarrior/config.py
```diff
@@ -204,9 +205,26 @@ def get_taskrc_path(conf, main_section):
...
 
 
-def get_data_path():
-    return os.path.expanduser(os.getenv('TASKDATA', '~/.task'))
+@lru_cache()
+def get_data_path(config, main_section):
+    taskrc = get_taskrc_path(config, main_section)
 
+    # We cannot use the taskw module here because it doesn't really support
+    # the `_` subcommands properly (`rc:` can't be used for them).
```
gdetrez: Wouldn't it be better to send a patch to taskw? There seems to be code there to parse the taskwarrior config file, so it should be easy to have it parse the output of `task _show` instead of a file.

gdetrez:
After reading through the changes and doing some quick tests, I don't think this would interfere with PR #370.

But since this `bitbucket_refresh_token` is actually some sort of password, even if it has limited access, have you considered storing it in the keychain instead of a plain-text file? In addition to better security, that would also relieve bugwarrior from having to locate the `TASKDATA` directory since this token seems to be the only thing that is stored there.

(If you decide to store it in plain text, I think we should at least set its mode to `600` when the data file is created.)

bugwarrior/config.py
ryneeverett: Makes sense.

ryneeverett:
> But since this bitbucket_refresh_token is actually some sort of password, even if it has limited access, have you considered storing it in the keychain instead of a plain-text file? In addition to better security, that would also relieve bugwarrior from having to locate the TASKDATA directory since this token seems to be the only thing that is stored there.

I think storing the token in the keychain would be a good feature, but I'm not sure it should be the only option. Given that we support storing passwords in plaintext in the configuration file, I think the primary advantage of token authentication is that it's a security improvement over password authentication if you're not using a keychain at all.

> (If you decide to store it in plain text, I think we should at least set its mode to 600 when the data file is created.)

Wouldn't this mean bugwarrior would need root permissions? Seems like the security disadvantages of that would outweigh the benefits.

ryneeverett:
> >  (If you decide to store it in plain text, I think we should at least set its mode to 600 when the data file is created.)

> Wouldn't this mean bugwarrior would need root permissions? Seems like the security disadvantages of that would outweigh the benefits.

Wow, I have no idea what I was thinking there. You're quite right.

mathstuf:
Wait, why would 600 permissions require root perms? That is rw by the owner which should be the user running bugwarrior, no?

ryneeverett:
> Wait, why would 600 permissions require root perms? That is rw by the owner which should be the user running bugwarrior, no?

Correct, I wasn't making sense.

ryneeverett:
@mathstuf Any reason not to use `taskw`'s config processing rather than `task _show`? I didn't look into the implementation details but it seems to work.

gdetrez:
According to what I saw, the current implementation in `taskw` only looks at the rc file. So it doesn't take `TASKDATA` into account and you'll get an exception if `data.locatian` is not set in the rc file.

ryneeverett:
@gdetrez Ah, thanks. I'll add a test case for that scenario to make sure.

ryneeverett:
@gdetrez Glad I tested that. You might checkout the tests I added to `test_config.py` to make sure I got the behaviour right.

tests/test_config.py
```diff
@@ -69,3 +70,43 @@ def test_BUGWARRIORRC_empty(self):
...
+        """
+        # Empty taskrc.
+        with open(self.taskrc, 'w'):
+            pass
+
+        os.environ['TASKDATA'] = ''
+
+        with self.assertRaisesRegexp(RuntimeError,
+                                     'Unable to determine the data location.'):
+            config.get_data_path(self.config, 'general')
```
gdetrez: I'm not sure about this last case: it seems that taskwarrior itself doesn't require you to set `data.location` and will use the default `~/.task` if not set:
```sh
~ ❯❯❯ TASKRC=$(mktemp) task _show | grep data.location
data.location=~/.task
```

ryneeverett: Nice catch.

ryneeverett:
> I think that your initial idea to parse the output of task _show was better,

It was @mathstuf's idea and you're probably right.

I was annoyed by the fact that due to taskwarrior's interpretation of `TASKDATA=''` as an empty `data.location`, we can't simply pass it into the environment. It seems like, in configuring the environment, we're implementing almost all the logic in order to solicit a known response out of taskwarrior.

But making taskwarrior give us the value is probably still best. This way if taskwarrior's behavior changes it will break the tests and not bugwarrior, whereas otherwise it would break bugwarrior and not the tests.

ryneeverett:
I don't have an opinion on taskw at this point. Hopefully one of the taskw maintainers -- @ralphbean and @coddingtonbear -- will weigh in.

mathstuf:
Is there anything holding this up?

ryneeverett:
See @gdetrez's (somewhat buried now) comment:

> Wouldn't it be better to send a patch to taskw? There seems to be code there to parse the taskwarrior config file, so it should be easy to have it parse the output of task _show instead of a file.

We were hoping to hear from a taskw maintainer (@ralphbean and @coddingtonbear) on this issue.

ryneeverett:
@gdetrez Any opposition to going ahead and merging this and dealing with potential taskw refactoring later?

tests/test_config.py
```diff
@@ -81,3 +70,41 @@ def test_BUGWARRIORRC_empty(self):
...
+    def test_taskrc_datalocation(self):
+        """
+        When TASKDATA is not set, data.location in taskrc should be respected.
+        """
+        os.environ['TASKDATA'] = ''
+        self.assertDataPath(self.lists_path)
+
+    def test_unassigned(self):
+        """
+        When data path is not assigned, raise an exception.
```
gdetrez: I think this comment is incorrect

gdetrez:
I found an comment that need fixing but otherwise I'm good :-)

PR #370, opened 2016-08-27
Add support for environment variable $BUGWARRIORRC

gdetrez:
* If `$BUGWARRIORRC` is set, then bugwarrior will try to read its
  configuration file from its content
* Extract the code that guess the path to the configuration file in its
  own function.
* Add some unit tests

I got a bit tired of editing my own `.bugwarriorrc` to test bugwarrior. This is inspired by the similar variable `TASKRC` used by taskwarrior.

### DIFF
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -5,7 +5,6 @@ import subprocess
 import sys
 
 import six
-from xdg import BaseDirectory
 
 import logging
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


bugwarrior/config.py
```diff
@@ -11,6 +11,11 @@
 log = logging.getLogger(__name__)
 
 
+# The name of the environment variable that can be used to ovewrite the path
+# to the bugwarriorrc file
+BUGWARRIORRC = "BUGWARRIORRC"
```
ryneeverett: This variable doesn't seem necessary to me (not likely to be mistyped or changed) but your call.

bugwarrior/config.py
```diff
@@ -150,28 +155,36 @@ def validate_config(config, main_section):
...
+def get_config_path():
+    """
+    Determine the path to the config file. This will return, in this order of
+    precedence:
+    - the value of $BUGWARRIORRC if set
+    - $XDG_CONFIG_HOME/bugwarrior/bugwarriorc if exists
+    - <dir>/bugwarrior/bugwarriorc if exists, for dir in $XDG_CONFIG_DIRS
+    - ~/.bugwarriorrc if exists
+    - $XDG_CONFIG_HOME/bugwarrior/bugwarriorc otherwise
+    """
```
ryneeverett: Good stuff. Could you correct the docs to reflect this? I'm thinking in particular of the first line of the "How to Configure" section which is now incorrect.

tests/test_config.py
```diff
@@ -0,0 +1,85 @@
...
+class TestGetConfigPath(unittest.TestCase):
+
+    def setUp(self):
+        self.tmpdir = mkdtemp()
+        self.old_HOME = os.environ.get('HOME', None)
+        self.old_xdg_config_home = xdg.BaseDirectory.xdg_config_home
+        self.old_xdg_config_dirs = xdg.BaseDirectory.xdg_config_dirs
+        os.environ['HOME'] = self.tmpdir
+        if config.BUGWARRIORRC in os.environ:
+            del os.environ[config.BUGWARRIORRC]
```
ryneeverett: Should we restore the BUGWARRIORRC environment variable in tearDown just like HOME?

bugwarrior/config.py
ryneeverett: Well maybe not incorrect but legacy.

tests/test_config.py
gdetrez: Yeah, that's probably safer

bugwarrior/config.py
gdetrez: I can add `BUGWARRIORC` but I also wanted to document the previous undocumented change (the xdg stuff), and now that I look closer at it, there are some things I'm not sure of. In particular:

* `~/.config/bugwarrior/bugwarriorrc` is now preferred over `~/.bugwarriorrc`. I don't think that's wrong but I want to make sure it's intentional.
* `/etc/xdg/bugwarrior/bugwarriorrc` will allo take precedence  over `~/.bugwarriorrc`. While it might make sense to have a system-wide bugwarrior configuration, I'd prefer if it was the other way around, i.e. that it would only be used if `~/.bugwarriorrc` is **not** present.

So I would like to at least change the order from

1. `$XDG_CONFIG_HOME`
2. `$XDG_CONIFG_DIRS`
3. `~/.bugwarriorrc`

to

1. `$XDG_CONFIG_HOME`
2. `~/.bugwarriorrc`
3. `$XDG_CONIFG_DIRS`

Any objection?

ralphbean: None here.  Go for it!

ryneeverett: Sounds right to me too.

ralphbean:
Looks good here.  Thanks for hacking on this @gdetrez!  :+1: to merge from me.

PR #371, opened 2016-08-29
Fix gitlab include/exclude repos

gdetrez:
The documentation said that to include/exclude a repository you should
simply write the project name, like `bar` but actually, the service
expects a namespace in the repository name, e.g. `foo/bar`.

This commit does two things:

* If the user writes a repository name without the namespace,
  automatically adds the value of `gitlab.login` as the namespace.
* Update the gitlab documentation accordingly


ralphbean:
LGTM!  Thanks!

Issue #365, opened 2016-08-08
githubmilestone 


vlekakis:
The {{githubmilestone}} seems to be a number rather the actual milestone.
Not sure if this what you meant, but does not seem the case from this: http://bugwarrior.readthedocs.io/en/latest/services/github.html

PR #372, opened 2016-08-29
Use milestone title instead of id for githubmilestone

gdetrez:
Note that the milestone id that was used before is the intertal github
id for the row, not the sequencial milestone number included in the url.
As such I didn't keep it at all as I assumed that it was fairly useless
for everyone.

The milestone number may eventually be useful but I guess we should
wait until someone asks for it explicitely before adding it to avoid
some sort of "UDA creap".

Fix issue #365


ralphbean:
Nice catch.  Thank you!

PR #373, opened 2016-08-29
Fix some problems with unicode

gdetrez:
Because of the import

    from __future__ import unicode_literals

We can pass all the arguments as unicode and let the logging library
take care of the encoding.


ralphbean:
:+1:, thank you.  :)

PR #378, opened 2016-09-05
bts: Reject configuration if UDD options specified but UDD not used

irl:
Just some extra config validation.


ryneeverett:
I can see the case where this might be a gotcha -- You configure UDD and it doesn't work and that's frustrating. I can also see the case where the validation might be frustrating -- You temporarily disable UDD and it fails and yells at you for using options that don't make sense.

The problem you're trying to help users avoid could be a lot more frustrating, whereas the one I'm suggesting is annoying but easy to fix. It's a tradeoff and as the maintainer of this service I think that's your shot to call.

:+1: to merge.

irl:
I think if you temporarily disable UDD, you would comment out all the lines that have "udd" in them. I think these are named well enough that the temporarily disabling case shouldn't be a large issue.

PR #381, opened 2016-09-09
Import JIRA sprint names as labels.

ralphbean:
I'd like to import more jira attributes (like use the end date of the sprint as
the due date for active issues in the sprint), but that's a bit more
complicated and can wait until later.

### DIFF
--- a/bugwarrior/docs/services/jira.rst
+++ b/bugwarrior/docs/services/jira.rst
@@ -68,8 +68,8 @@ If your Jira instance is only available over HTTPS, and you're running into
 
     jira.verify_ssl = False
 
-Import Labels as Tags
-+++++++++++++++++++++
+Import Labels and Sprints as Tags
++++++++++++++++++++++++++++++++++
 
 The Jira issue tracker allows you to attach labels to issues; to
 use those labels as tags, you can use the ``jira.import_labels_as_tags``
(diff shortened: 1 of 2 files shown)


bugwarrior/docs/services/jira.rst
```diff
@@ -68,18 +68,22 @@ If your Jira instance is only available over HTTPS, and you're running into
...
 
     jira.import_labels_as_tags = True
 
-Also, if you would like to control how these labels are created, you can
-specify a template used for converting the Jira label into a Taskwarrior
-tag.
+You can also import the names of any sprints associated with an issue as tags,
+by setting the ``jira.import_sprints`` option::
+
+    jira.import_sprints = True
```
coddingtonbear: I know you're going to hate me for suggesting this, but even though I might regret naming `import_labels_as_tags` what I did name it, it might be less confusing for future readers if both <thing-to-tag> import configuration options were named similarly.

ralphbean: No hate here.  It's reasonable!

bugwarrior/services/jira.py
```diff
@@ -206,10 +237,26 @@ def __init__(self, *args, **kw):
...
             'label_template', default='{{label}}', to_type=six.text_type
         )
 
+        self.sprint_field = None
+        if self.import_sprints_as_tags:
+            sprint_fields = [field for field in self.jira.fields()
+                             if field['name'] == 'Sprint']
+            if len(sprint_fields) > 1:
+                log.warn("More than one sprint custom field found.  Ignoring sprints.")
+                self.import_sprints_as_tags = False
```
ryneeverett: I don't know anything about jira but why not add multiple sprint tags?

ryneeverett: I don't know anything about jira but why not add multiple sprint tags?

ralphbean: OK, this is dealt with in b2ca93d

PR #382, opened 2016-09-14
Improve travis coverage

ryneeverett:

PR #384, opened 2016-09-17
Test db module

ryneeverett:
<s>Based on #364 and #382, so not really a reasonable thing to review right now but I thought I'd throw it up here to avoid duplication of effort and showcase codecov. I may add more tests in the meantime.</s> Rebased and should be good to go.

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -56,10 +56,6 @@ class MultipleMatches(Exception):
     pass
 
 
-def normalize_description(issue_description):
-    return issue_description[:issue_description.index(' .. http')]
-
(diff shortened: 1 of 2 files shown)


ryneeverett:
Well [here's the codecov](https://codecov.io/gh/ralphbean/bugwarrior/tree/4d82f13ca2f5638818d7bb36477821f330874d5b). I'm guessing it doesn't show up in "checks" since codecov didn't run against master and thus doesn't know the "coverage diff".

ralphbean:
Looks good to me, @ryneeverett.  Let's merge it in.

PR #386, opened 2016-09-19
Fix unicode issues in notifications

gdetrez:
Problem came from mixing unicode and non-unicode strings in %-formatting.  Fixed by using only unicode, which should make the code more python 3 compatible as well.

This also supress a warning when using gobject notifications

### DIFF
--- a/bugwarrior/notifications.py
+++ b/bugwarrior/notifications.py
@@ -1,4 +1,5 @@
-import datetime
+from __future__ import unicode_literals
+
 import os
 import urllib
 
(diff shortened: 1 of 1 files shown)


ryneeverett:
A regression test would be nice. This is the kind of bug likely occur when we port to python3.

gdetrez:
I actually made a tentative before open the PR but the current state of the code make it difficult to write isolated tests (multiple responsibilites, modules imported in function body preventing patching...)

My dilemma is that it seems that the changes required to be able to write a test will be much more important than the current fix, and without existing integration tests for the different notification systems supported, I'm not sure I'm not going to break something else.

But I can give it an other try and I can at least do manual integration testing with the notification backends that are available on fedora.

ralphbean:
:+1: to merge from me.

ryneeverett:
My attitude is, if the test is hard to write then "testability" is a feature that can be put off for a future PR. Go ahead and merge if you don't have time for that right now.

PR #387, opened 2016-09-22
Feature/python3

wfranzini:
Hi,

I've done some work to make bugwarrior pass the testsuite also using python3.4.

In order to succeed I've update also pyac and python-megaplan adding dependency to yet
unreleased version only in tox.ini

Hope this help

ciao
walter franzini

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -95,7 +95,7 @@ def hamdist(str1, str2):
 
 def get_managed_task_uuids(tw, key_list, legacy_matching):
     expected_task_ids = set([])
-    for keys in list(key_list.values()):
+    for keys in key_list.values():
         tasks = tw.filter_tasks({
             'and': [('%s.any' % key, None) for key in keys],
             'or': [
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -1,6 +1,10 @@
 from __future__ import unicode_literals
+from future import standard_library
+standard_library.install_aliases()
+from builtins import zip
+from builtins import object
 
-from ConfigParser import NoOptionError, NoSectionError
+from configparser import NoOptionError, NoSectionError
 import os
 import re
 import subprocess
@@ -91,7 +95,7 @@ def hamdist(str1, str2):
 
 def get_managed_task_uuids(tw, key_list, legacy_matching):
     expected_task_ids = set([])
-    for keys in key_list.values():
+    for keys in list(key_list.values()):
         tasks = tw.filter_tasks({
             'and': [('%s.any' % key, None) for key in keys],
             'or': [
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -21,7 +21,9 @@ log = logging.getLogger(__name__)
 MARKUP = "(bw)"
 
 
-DOGPILE_CACHE_PATH = os.path.expanduser('~/.cache/dagd.dbm')
+DOGPILE_CACHE_PATH = os.path.expanduser(
+    os.getenv('XDG_CACHE_HOME', '~/.cache') + '/dagd.dbm')
+
 if not os.path.isdir(os.path.dirname(DOGPILE_CACHE_PATH)):
     os.makedirs(os.path.dirname(DOGPILE_CACHE_PATH))
 CACHE_REGION = dogpile.cache.make_region().configure(
--- a/bugwarrior/docs/generate_service_template.py
+++ b/bugwarrior/docs/generate_service_template.py
@@ -76,7 +76,7 @@ if __name__ == '__main__':
     rows = []
     for name, obj in inspect.getmembers(module):
         if inspect.isclass(obj) and issubclass(obj, Issue):
-            for field_name, details in list(obj.UDAS.items()):
+            for field_name, details in obj.UDAS.items():
                 rows.append(
                     [
                         '``%s``' % field_name,
--- a/bugwarrior/docs/generate_service_template.py
+++ b/bugwarrior/docs/generate_service_template.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import inspect
 import os
 import sys
@@ -5,6 +6,7 @@ import sys
 from jinja2 import Template
 
 from bugwarrior.services import Issue
+from functools import reduce
 
 
 def make_table(grid):
@@ -74,7 +76,7 @@ if __name__ == '__main__':
     rows = []
     for name, obj in inspect.getmembers(module):
         if inspect.isclass(obj) and issubclass(obj, Issue):
-            for field_name, details in obj.UDAS.items():
+            for field_name, details in list(obj.UDAS.items()):
                 rows.append(
                     [
                         '``%s``' % field_name,
@@ -97,4 +99,4 @@ if __name__ == '__main__':
             'uda_table': make_table(rows)
         })
 
-    print rendered
+    print(rendered)
--- a/bugwarrior/services/bitbucket.py
+++ b/bugwarrior/services/bitbucket.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from builtins import filter
 
 import requests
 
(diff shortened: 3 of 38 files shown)


irl:
Awesome.

For Debian, I'm looking to get Python 3.5 support. Supporting 3.4 is probably a good idea also.

Can we set up tests to look at all of these?

I see we have a feature/py3 branch, but "This branch is 873 commits behind develop.".

wfranzini:
the support for python3 is in PR #387 which is ahead of develop branch.  Adding 3.5 is really simple: it requires adding py35 in the newly created tox.ini

wfranzini:
The changes in this PR has been generated by futurize, so I've not looked at the details since all the tests passed.

I also have a branch created by hand and the need to convert python3 iterators to list is due to `log.debug(..., len(..))` statement, since in python `filter` and `map` does not seems to support `len()`.

BTW my other branch has one test still failing that I'm unable to 'fix'.

wfranzini:
Hi,

I've fixed one more compatibility problem in Issue.__str__.

tox.ini
```diff
@@ -0,0 +1,16 @@
...
+setenv =
+       XDG_CACHE_HOME={envtmpdir}/
+# NOTE:
+# * as of 2016-09-20 pyac 1.4 does not support python3, so we must use
+#   the github repository
+# * as of 2016-09-20 python-megaplan does not support python3, so a
+#   github repository with the needed fixes has been created
+deps = pysimplesoap
+     py34: git+https://github.com/kostajh/pyac.git
+     py34: git+https://github.com/wfranzini/python-megaplan.git
```
ryneeverett: pyac 1.5 was just released. Any reason you haven't made a PR to upstream python-megaplan?

wfranzini: If I'm not wrong python-megaplan is hosted on bitbucket under mercurial, I've yet to understand how to make a PR.

wfranzini:
@ryneeverett I've no problem removing the `list` conversion, however I need some suggestion how to keep `debug` statement that print the number of issues in the "list" or permission to drop such debug messages.

ryneeverett:
I would say in those cases coercing to `list` is fine, but I think many of the current conversions are not necessary. Correct me if I'm wrong about that.

wfranzini:
@ryneeverett I've investigated a bit the problem and found that the result of the following code change when executed under python2 or python3:

``` 
a = [1, 2, 3, 4, 5]
b = filter(lambda x: True, a)
list(b)
4 in b
```
so the list coercion on log.debug statements is not a viable solution. I should get rid of `filter` and possibly `map` and use list comprehension.

I'm open to suggestions?

ryneeverett:
I'd say that's another valid use case for coercing to list, but I'm still not sure that every list conversion in this PR is necessary. I'll see if I can find a specific example.

bugwarrior/db.py
```diff
@@ -89,7 +95,7 @@ def hamdist(str1, str2):
 
 def get_managed_task_uuids(tw, key_list, legacy_matching):
     expected_task_ids = set([])
-    for keys in key_list.values():
+    for keys in list(key_list.values()):
```
ryneeverett: We're using `values()` as an iterator in this statement. Why `list`?

wfranzini: Because `values()` in py3 is a view and not an iterator.   The following code work as expected in py2 while it raise a RuntimeError in py3.4:
```
a = {1: 1, 2: 2, 3: 3}
for i in a.values():
     a[-1] = 1
```
Maybe today the bw code does not expose this effect, but without `list()` tomorrow an innocent change may break something.

this is a change introduced by futurize to *ensure*, IMHO, the same behavior in both version.

If you think `list()` should still be removed I'll work on it.

bugwarrior/services/bitbucket.py
```diff
@@ -199,20 +200,20 @@ def get_owner(self, issue):
     def issues(self):
         user = self.config.get(self.target, 'bitbucket.username')
         response = self.get_collection('/repositories/' + user + '/')
-        repo_tags = filter(self.filter_repos, [
+        repo_tags = list(filter(self.filter_repos, [
```
wfranzini: It's used twice in the same method, in py3 the second time it's "empty".  The following snippet rais an AssertionError in py3, while it succeed in py2:
```
a = [1, 2, 3, 4, 5]
b = filter(lambda x: True, a)
c = [el for el in b]
d = [el for el in b]
assert len(c) == len(d)
```
I think futurize is right here :-)

bugwarrior/docs/generate_service_template.py
```diff
@@ -74,7 +76,7 @@ def row_comparator(left_row, right_row):
     rows = []
     for name, obj in inspect.getmembers(module):
         if inspect.isclass(obj) and issubclass(obj, Issue):
-            for field_name, details in obj.UDAS.items():
+            for field_name, details in list(obj.UDAS.items()):
```
wfranzini: A view in py3

bugwarrior/services/bitbucket.py
ryneeverett: Ok, I concede on this one.

bugwarrior/db.py
ryneeverett: Views have an `iter` method. Are you saying it is never appropriate to use it due to this risk?

wfranzini: I'm not knowledgeable enough on Python to criticize view, I'm just trying to make sense of changes introduced by futurize in order to answer your objections :-)
The example above is only to show that the same source code can expose very different behavior depending on the version of Python used, not to criticize the behaviors.

ryneeverett: Fair enough. I'm opposed to converting to `list` in cases where duck typing meets our need. If you disagree we can request comments from other contributors.

bugwarrior/services/gitlab.py
```diff
@@ -405,7 +410,7 @@ def issues(self):
                 self.get_repo_issues(rid)
             )
         log.debug(" Found %i issues.", len(issues))
-        issues = filter(self.include, issues.values())
+        issues = list(filter(self.include, list(issues.values())))
```
ryneeverett: Again, don't think inner `list` is needed.

bugwarrior/services/gitlab.py
```diff
@@ -453,7 +458,7 @@ def issues(self):
             todos = self.get_todos()
             log.debug(" Found %i todo items.", len(todos))
             if not self.include_all_todos:
-                todos = filter(self.include_todo(repos), todos.values())
+                todos = list(filter(self.include_todo(repos), list(todos.values())))
```
ryneeverett: inner `list`

bugwarrior/services/phab.py
```diff
@@ -83,16 +84,16 @@ def issues(self):
                 issues_owner = self.api.maniphest.query(status='status-open', ownerPHIDs=self.shown_user_phids)
                 issues_cc = self.api.maniphest.query(status='status-open', ccPHIDs=self.shown_user_phids)
                 issues_author = self.api.maniphest.query(status='status-open', authorPHIDs=self.shown_user_phids)
-                issues = list(list(issues_owner.iteritems()) + list(issues_cc.iteritems()) + list(issues_author.iteritems()))
+                issues = list(list(issues_owner.items()) + list(issues_cc.items()) + list(issues_author.items()))
```
ryneeverett: The outer list should be unnecessary.

bugwarrior/services/github.py
```diff
@@ -348,7 +349,7 @@ def include(self, issue):
     def issues(self):
         all_repos = self.client.get_repos(self.username)
         assert(type(all_repos) == list)
-        repos = filter(self.filter_repos, all_repos)
+        repos = list(filter(self.filter_repos, all_repos))
```
ryneeverett: Looks like a once-time use iterator.

bugwarrior/services/phab.py
```diff
@@ -83,16 +84,16 @@ def issues(self):
...
                 # Delete duplicates
                 seen = set()
                 issues = [item for item in issues if str(item[1]) not in seen and not seen.add(str(item[1]))]
             if self.shown_project_phids is not None:
                 issues = self.api.maniphest.query(status='status-open', projectsPHIDs = self.shown_project_phids)
-                issues = list(issues.iteritems())
+                issues = list(issues.items())
```
ryneeverett: I think this point is still valid.

wfranzini: Yup. Also the same stuff two lines below?

ryneeverett:
Continued in #406...

Issue #377, opened 2016-09-02
bugwarrior pulls in merged pull requests but leaves them pending


bowlofeggs:
I turned on ```github.involved_issues = True``` today and noticed that it left merged pull requests as "pending" tasks. For example, this pull request was closed long ago:

https://github.com/psss/did/pull/91

However, it was pulled in as pending:

```
$ task 461
No command specified - assuming 'information'.

Name             Value                                                                
ID               461
Description      91: Support did yesterday.                                           
                   2016-09-02 12:37:30 @psss - Thanks for the patch, Randy!           
Status           Pending
Project          did                                                                  
Entered          2016-09-02 12:37:30 (15min)
Last modified    2016-09-02 12:37:30 (15min)                                          
Virtual tags     ANNOTATED ORPHAN PENDING READY UDA UNBLOCKED LATEST PROJECT PRIORITY
UUID             725afd7f-e34e-45c9-ae42-ddf766053f28                                 
Urgency           7.8
Priority         H                                                                    
[githubbody      fixes #79]
[githubcreatedon 1456877280]                                                          
[githubnumber    91]
[githubrepo      psss/did]                                                            
[githubtitle     Support did yesterday.]
[githubtype      pull_request]                                                        
[githubupdatedat 1459773748]
[githuburl       https://github.com/psss/did/pull/91]                                 

    project             1 *    1 =      1
    annotations       0.8 *    1 =    0.8
    UDA priority.H      1 *    6 =      6
                                   ------
                                      7.8

Date                Modification                                               
2016-09-02 12:37:30 Annotation of '@psss - Thanks for the patch, Randy!' added.
```

I assume this might have happened with issues as well, but I did not find any examples in my tasks (there are a ton of them now, and also it failed to finish pulling due to #376).

gdetrez:
If any service fails, bugwarrior will abort and will not touch the tasks(see issue #279), which might explain why the PR is not marked as done.

beav:
I hit this issue as well with 1.4.0, I pulled in a number of PRs but the merged PRs were still marked as 'Pending'.

beav:
@gdetrez The merged PRs still appear when I tested with that.

FWIW here is my config entry for github:

```
[github]
service = github
github.username = beav
github.login = beav
github.token = <token>
add_tags = github
github.import_labels_as_tags = True
github.label_template = GH_{{label}}
github.involved_issues = True
github.filter_pull_requests = True
github.include_user_issues = False
```

[...]

PR #389, opened 2016-10-03
Only query for open github PRs (#377)

beav:
Previously, setting `involved_issues = True` would pull in PRs that were in any
state, including merged PRs. This patch amends the involved issues query to
only search for open PRs.


ralphbean:
Nice find!

PR #392, opened 2016-10-04
[WIP] bts: Allow to ignore bugs by tag

irl:
This commit adds the ability to ignore bugs that have been marked as "wontfix". It also adds a new UDA containing the tags on bugs as a side effect.

### DIFF
--- a/bugwarrior/docs/services/bts.rst
+++ b/bugwarrior/docs/services/bts.rst
@@ -62,15 +62,24 @@ Excluding bugs marked pending
 +++++++++++++++++++++++++++++
 
 Debian bugs are not considered closed until the fixed package is present in the
-Debian archive. Bugs do cease to be outstanding tasks however as soon as you have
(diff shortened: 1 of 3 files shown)


ryneeverett:
They're tagged as wontfix but not closed?

irl:
wontfix means that the current maintainer is not interested in fixing the problem. They may still accept a patch if it's well done, or a future maintainer may feel differently about the bug.

ryneeverett:
Ah, that makes sense. However, I wonder if it wouldn't be better to have an option to import bts tags as taskwarrior tags and then another option to ignore arbitrary tags?

irl:
That sounds like a decent idea, I'll take a go at that.

ralphbean:
Hey @irl - we're at the PyCon Sprints and using the time to review the open pull requests here (which I've neglected for too long).  I'm going to close this one since it's been inactive for a while.  Feel free to resubmit if you want to bring it back.

PR #394, opened 2016-10-12
redmine: import first 100 issues instead 25

lzap:
This is quick and dirty patch to increase current limitation from 25 records
returned to 100 (hard coded maximum). Fore more, we'd need to implement
pagination. Since I have less then 100 bugs assigned, I am not interested in
that for the moment. But who knows :-)

### DIFF
--- a/bugwarrior/docs/services/redmine.rst
+++ b/bugwarrior/docs/services/redmine.rst
@@ -4,6 +4,8 @@ Redmine
 You can import tasks from your Redmine instance using
 the ``redmine`` service name.
 
+Only first 100 issues are imported at the moment.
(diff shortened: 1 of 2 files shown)


lzap:
For the record, the relevant page you are looking for is at http://www.redmine.org/projects/redmine/wiki/Rest_api

beav:
I have less than 25 issues, but this patch did not break anything when I applied it and ran a pull.

kostajh:
@lzap @beav please also have a look at my PR for #395 which includes this. Thanks!

lzap:
Thanks for the test fix!

Issue #390, opened 2016-10-03
Improved support for Redmine


kostajh:
Hi @ralphbean, as a heads up I'm working on a larger PR to improve the support available for Redmine (and a bunch of fields, support for annotations, etc). Let me know if you have any questions or concerns in the meantime!

lzap:
Could you elaborate LIMIT/OFFSET? Currently only max of 25 records are found, this could be increased to 100 (maximum allowed) or ideally with pagination: http://www.redmine.org/projects/redmine/wiki/Rest_api

lzap:
I created a simple patch to increase the limit to 100: https://github.com/ralphbean/bugwarrior/pull/394

Pagination would be super cool :-)

ralphbean:
Seems like this was resolved by #395.

PR #395, opened 2016-10-12
Redmine improvements

kostajh:
This PR is for #390. I'm posting this now in case someone wants to carry on, I'm going to be mostly offline for the next few weeks. (If you do work on this please post here so we're not working to cross purposes!)

Still TODO:
- Load comments as annotations
- Support for custom fields
- Pagination support for more than 100 records?
- Fix tests

Everything here is working for me, btw.

### DIFF
--- a/bugwarrior/services/redmine.py
+++ b/bugwarrior/services/redmine.py
@@ -143,6 +143,7 @@ class RedMineIssue(Issue):
             created_on = self.parse_date(created_on).replace(microsecond=0)
         if spent_hours:
             spent_hours = str(spent_hours) + ' hours'
+            spent_hours = self.get_converted_hours(spent_hours)
         if estimated_hours:
             estimated_hours = str(estimated_hours) + ' hours'
             estimated_hours = self.get_converted_hours(estimated_hours)
--- a/bugwarrior/services/redmine.py
+++ b/bugwarrior/services/redmine.py
@@ -40,6 +40,15 @@ class RedMineIssue(Issue):
     SUBJECT = 'redminesubject'
     ID = 'redmineid'
     DESCRIPTION = 'redminedescription'
+    TRACKER = 'redminetracker'
+    STATUS = 'redminestatus'
+    AUTHOR = 'redmineauthor'
+    CATEGORY = 'redminecategory'
+    START_DATE = 'redminestartdate'
+    SPENT_HOURS = 'redminespenthours'
+    ESTIMATED_HOURS = 'redmineestimatedhours'
+    CREATED_ON = 'redminecreatedon'
+    UPDATED_ON = 'redmineupdatedon'
 
     UDAS = {
         URL: {
@@ -58,8 +67,45 @@ class RedMineIssue(Issue):
             'type': 'string',
             'label': 'Redmine Description',
         },
+        TRACKER: {
+            'type': 'string',
+            'label': 'Redmine Tracker',
+        },
+        STATUS: {
+            'type': 'string',
+            'label': 'Redmine Status',
+        },
+        AUTHOR: {
+            'type': 'string',
+            'label': 'Redmine Author',
+        },
+        CATEGORY: {
+            'type': 'string',
+            'label': 'Redmine Category',
+        },
+        START_DATE: {
+            'type': 'date',
+            'label': 'Redmine Start Date',
+        },
+        SPENT_HOURS: {
+            'type': 'duration',
+            'label': 'Redmine Spent Hours',
+        },
+        ESTIMATED_HOURS: {
+            'type': 'duration',
+            'label': 'Redmine Estimated Hours',
+        },
+        CREATED_ON: {
+            'type': 'date',
+            'label': 'Redmine Created On',
+        },
+        UPDATED_ON: {
+            'type': 'date',
+            'label': 'Redmine Updated On',
+        },
+
     }
-    UNIQUE_KEY = (URL, )
+    UNIQUE_KEY = (ID, )
 
     PRIORITY_MAP = {
         'Low': 'L',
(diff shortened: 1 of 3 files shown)


ryneeverett:
I rebased. I'm a little horrified that this new github feature encourages this kind of thing though. Anyone with commit access can anonymously screw with PR's and the modifications are unlikely to be detected if they're careful. In other words, you can no longer count on the submitter to have reviewed their own PR!

kostajh:
@ryneeverett this all works, but I could use help figuring out what to do with the tests. Also, is it best practice to map the Redmine due date directly to a TW due date? I'm not sure the other services do this.

bugwarrior/services/redmine.py
```diff
@@ -1,23 +1,30 @@
...
-        if user_id is not None:
-            args["assigned_to_id"] = user_id
+    def find_issues(self, issue_limit=100):
+        args = {}
+        # TODO: if issue_limit is greater than 100, implement pagination to return all issues.
+        # Leave the implementation of this to the unlucky soul with >100 issues assigned to them.
+        if issue_limit is not None:
+            args["limit"] = issue_limit
+
+        args["assigned_to_id"] = 'me'
```
ryneeverett: Did you mean to leave this this way?

kostajh: Yes, but, should it be conditional based on whether the user has selected `only_if_assigned = True` in their config?

ryneeverett: Ah, I hadn't realized that `"me"` is a special value in the redmine API.

ryneeverett: > Yes, but, should it be conditional based on whether the user has selected only_if_assigned = True in their config?

Yes, I believe that's correct.

ryneeverett:
Reactions in no particular order:

1. Do we know why redmine's time zones are different than other services? In the tests, the other services return a datetime with a `pytz.UTC` timezone while redmine returns a datetime with a `dateutil.tz.tz.tzutc()`. This may well be an inconsistency in the way I wrote the tests, but it is curious.
2. The UDA `ASSIGNED_TO` appears to be equivalent to gitlab's `ASSIGNEE` and bz's `ASSIGNED`. Would it make sense to standardize? I'd be inclined towards `ASSIGNEE`.
3. The following UDA's appear to be completely unique across services:
- TRACKER
- CATEGORY
- START_DATE
- SPENT_HOURS
- ESTIMATED_HOURS

I haven't developed an opinion on their merits, but this may be worth considering and soliciting feedback on.

kostajh:
@ryneeverett thank you for fixing the tests!

> Do we know why redmine's time zones are different than other services? In the tests, the other services return a datetime with a pytz.UTC timezone while redmine returns a datetime with a dateutil.tz.tz.tzutc(). This may well be an inconsistency in the way I wrote the tests, but it is curious.

I'm not sure about this, sorry. I do know that what's in `redmine.py` in this PR works for keeping the dates consistent after running `bugwarrior-pull`, when I first started working on this PR, Taskwarrior thought that the date field was changing across each sync and would update each task on every pull.

> The UDA ASSIGNED_TO appears to be equivalent to gitlab's ASSIGNEE and bz's ASSIGNED. Would it make sense to standardize? I'd be inclined towards ASSIGNEE.

Up to you. `assigned_to` and `assigned` are how these fields are defined in Redmine, so IMO it makes sense to keep the same format for this service, but I don't have a strong feeling about it one way or the other.

> The following UDA's appear to be completely unique across services

Same as above, these are core fields provided by Redmine for issues. I use all of them in my Taskwarrior reports, so I'd like to keep them :) 

A related point, in Redmine one can have custom fields. I'm not sure if it's within scope to allow a user to define custom fields to pull from or how we could achieve these in `.bugwarriorrc`. But if there is a solution for custom fields, then I could use that for ESTIMATED_HOURS and SPENT_HOURS, etc.

ryneeverett:
> Also, is it best practice to map the Redmine due date directly to a TW due date?

No. See https://github.com/ralphbean/bugwarrior/issues/385.

ryneeverett:
Ah, the timezones are probably different because they're retrieved from `task calc`. I don't think that's a problem though.

bugwarrior/services/redmine.py
```diff
@@ -82,10 +191,21 @@ def get_issue_url(self):
             self.origin['url'] + "/issues/" + six.text_type(self.record["id"])
         )
 
+    def get_converted_hours(self, estimated_hours):
+        tw = TaskWarriorShellout()
+        calc = tw._execute('calc', estimated_hours)
+        return (
+            calc[0].rstrip()
+        )
```
ryneeverett: None of the other services convert to taskwarrior format so I'm doubtful we should be here. You want to use the redmine duedate as your taskwarrior duedate though, and I'm not sure how, or if it's possible, to do this if the due dates are stored in integer format.

kostajh: The `estimated_hours` field is different from due date. Estimated hours comes from Redmine in the format `2`, `2.9`, `3.25`, etc. Here I am converting this integer value to a taskwarrior duration field.

ryneeverett: Wow, I have no idea what I was thinking an hour ago.

kostajh:
thanks very much for your help @ryneeverett! I'll follow up later with annotations support. Also, do you have any thoughts on how custom field support might be implemented?

ryneeverett:
> I'll follow up later with annotations support.

In a separate PR? I think this one's big enough and pretty well ready to merge.

> Also, do you have any thoughts on how custom field support might best be implemented?

I'm assuming you know about [field templates](https://bugwarrior.readthedocs.io/en/latest/common_configuration.html#field-templates). Can you be a bit more specific about what you're trying to do?

kostajh:
> In a separate PR? I think this one's big enough and pretty well ready to merge.

Yes in a separate PR.

> I'm assuming you know about field templates. Can you be a bit more specific about what you're trying to do?

Sure. And this is also for a future PR. My redmine instance has custom fields on issues, for example we use a field called "GitHub Branch / PR" where one can paste in the URL to the relevant branch/PR on GitHub that relates to a particular Redmine issue.

This comes across in the `issues.json` like so:

```
      "custom_fields": [
        {
          "id": 12,
          "name": "GitHub Branch / PR",
          "value": "https://github.com/ralphbean/bugwarrior/pull/395"
        }
      ],
```

I'm not sure I see from the field templates how I could set up my configuration to write the value of this field to a new UDA.

ryneeverett:
Gotcha, I wasn't sure what you meant by "fields". I would think the "custom fields" functionality you're talking about should be implemented as a new [Common Service Configuration Option](https://bugwarrior.readthedocs.io/en/latest/common_configuration.html#common-service-configuration-options), right?

PR #397, opened 2016-10-13
gitlab: match types when including all todo items

mathstuf:
---
When fetching all TODO items, the branch did not match the type used when filtering todo items.


ryneeverett:
Wouldn't it be better to have `get_todos` return `todos.values()`?

mathstuf:
Indeed. Updated.

ryneeverett:
Ah, even better.

PR #399, opened 2016-10-13
Filter involved issues

mathstuf:


bugwarrior/services/github.py
```diff
@@ -352,9 +360,8 @@ def issues(self):
 
         issues = {}
         if self.involved_issues:
-            issues.update(
-                self.get_involved_issues(self.username)
-            )
+            involved = self.get_involved_issues(self.username)
+            issues.update(dict(filter(self.filter_issues, involved.items())))
```
mathstuf: Ah, appears that `update` can take an iterator. Will fix.

bugwarrior/services/github.py
```diff
@@ -324,16 +325,23 @@ def get_owner(self, issue):
         if issue[1]['assignee']:
             return issue[1]['assignee']['login']
 
+    def filter_issues(self, issue):
+        _, (_, issue) = issue
+        return self.filter_repo_name(os.path.basename(issue['repository_url']))
```
mathstuf: Unfortunately, search results are really sparse and this seems to be the best way to get the name of the repo :( .

bugwarrior/services/github.py
```diff
@@ -352,9 +360,8 @@ def issues(self):
 
         issues = {}
         if self.involved_issues:
-            issues.update(
-                self.get_involved_issues(self.username)
-            )
+            involved = self.get_involved_issues(self.username)
+            issues.update(filter(self.filter_issues, involved.items()))
```
ryneeverett: Could we use `involved.values()` since we're just throwing the key away anyway?

mathstuf: We're throwing it away in the filter function, but not on `issues` itself. The `involved` intermediate is hanging around unnecessarily though *grumble*Python and function scoping*grumble* ;) .

mathstuf: It appears that the reason the key is important here is that an issue can appear for multiple reasons (assignment, in a repo I care about, mentioned, etc.). The key is the URL of the issue so that any duplicates get dropped this way.

ryneeverett:
I'm not confident whether this is the expected behaviour of this setting. Do you know if it's implemented this way for other services?

ryneeverett:
That is, someone might plausibly want involved issues to be included, even if they're excluded by filter rules.

mathstuf:
Another boolean configuration I guess?

gdetrez:
I think it makes sens for the filters to apply to everything but I would like to argue against an other boolean setting. If you want to pull all involved issues, you can add a separate section in your bugwarriorrc for that.

mathstuf:
Now also filters directly assigned issues as well.

bugwarrior/services/github.py
```diff
@@ -324,16 +324,23 @@ def get_owner(self, issue):
         if issue[1]['assignee']:
             return issue[1]['assignee']['login']
 
+    def filter_issues(self, issue):
+        repo, _ = issue
+        return self.filter_repo_name(repo.split('/')[-3])
```
ryneeverett: Nice catch, though I wonder if `repo.split('/')[1]` would be more understandable.

mathstuf: There is a part of me that hopes github will support nested repos under orgs someday :) .

ryneeverett: Fair enough.

ryneeverett:
Thanks for the judgement @gdetrez, I just wanted to make sure folks thought this through before merging.

Issue #385, opened 2016-09-18
Set tasks due to github milestone due date?


vantreeseba:
Should we set the tasks due date on import to the milestones date?

I'm willing to work on this, but just thought I would ask what people thought before I dug in too far.

ryneeverett:
Makes sense to me. There may be some ambiguity added by github features down the road -- this new "Projects" thing I haven't looked at may have it's own concept of due dates, github may add a duedate feature to issues, etc -- but those can always be resolved by giving precedence to the most immediate due date.

irl:
Do we talk to anyone at @GitHub? Maybe they would have input or could at least point us at a roadmap of what they think their projects thing is. It would be in their interest if people building tools to work with GitHub weren't fighting against their new features.

gdetrez:
It could be annoying if someone use personal due dates to organize their tasks and bugwarrior systematically reset them to milestone dates they way it resets task descriptions or projects.

IMO this feature should be optional but I think there is no need to introduce yet an other option, instead pull the milestone date in a UDA (e.g. `githubmilestonedate`) and make sure that you can set the due date with a template:
```
due_template = {{ githubmilestonedate }}
```

Bonus: add timedelta to the rendering context and you can do cool and useful stuff. Like if your goal is to process all issues in the week after they are created:
```
due_template = {{ githubcreatedon + timedelta(weeks=1) }}
```
or if you want to finish coding a month before the milestone is released to let people test stuff:
```
due_template = {{ githubmilestonedate - timedelta(months=1) }}
```

vantreeseba:
I think @gdetrez is right about going the UDA path.  Especially if more due dates are added to other parts of the issues/projects, this will keep it from becoming tied to any single point of data.

I think that, for now, we could set the default due_template to the milestone date.

This, to me looks like a good path foreward, so I'm going to proceed on it.

ryneeverett:
The conclusion is that this use case would be better served by a milestone duedate UDA. Adding UDA's isn't a lot of work and we're fairly lenient about merging those pull requests so I'm going to close this since nobody has wanted it that seriously for the last 5+ years.

PR #405, opened 2016-10-20
[WIP] optionally set bug to "started" when assigned

### LINKED ISSUE #396: [RFE] add start date to tasks if corresponding bug is in 'assigned' state
Currently, it's hard to tell which bugs are in progress via `task list`. It would be helpful if there was a configuration option on the bugzilla importer to set the task's start datetime to the current datetime if the bug is in `ASSIGNED` and there's no start datetime set currently. If the bug leaves `ASSIGNED` for whatever reason, the start datetime can be removed.

beav:
This patch creates a new UDA for when a bug was set to ASSIGNED state. It is
useful for setting the start time on a bug via configuration.

Fixes #396


beav:
this still needs some cleanup, but I wanted to get the PR in while it's being worked on for any initial comment

gdetrez:
I would like to suggest a similar thing as in #385: instead of adding yet an other boolean option, use the template mechanism: pull the assignation date in an UDA and make sure that you can set the start date with a template:
```
start_template = {{ bzassignedon }}
```

beav:
@gdetrez ah that is cool, I think I see what you mean. I will refactor.

beav:
@gdetrez patch updated to use a UDA

gdetrez:
Thanks @beav!

Note that I have never tested to set the start date with a template so you might want to make sure that it works. But even in case it doesn't I think we can merge this first and take the template problem in an other issue/pr. Or not, whichever you prefer ;-)

I've added WIP to the title because you wrote that you opened the PR for early feedback. Feel free to remove it whenever you feel ready for merging :-) (We will most probably ask for test and documentation of the new UDA.)

ralphbean:
This looks good to me.  @beav, can we merge it?

bugwarrior/services/bz.py
```diff
@@ -65,6 +73,14 @@ def to_taskwarrior(self):
         if self.extra.get('needinfo_since', None) is not None:
             task[self.NEEDINFO] = self.extra.get('needinfo_since')
 
+        # iff field is defined, use it, converting None to empty string.
+        if 'assigned_on' in self.extra:
+            if self.extra['assigned_on']:
+                task[self.ASSIGNED_ON] = self.extra.get('assigned_on')
+            else:
+                task[self.ASSIGNED_ON] = ''
```
ryneeverett: Seems more straightforward to me: `task[self.ASSIGNED_ON] = self.extra.get('assigned_on') or ''`

bugwarrior/services/bz.py
```diff
@@ -260,9 +276,36 @@ def issues(self):
...
+                extra['assigned_on'] = None
+
             issue_obj.update_extra(extra)
             yield issue_obj
 
+    def _get_assigned_date(self, issue):
+        assigned_date = None
+
+        if issue['status'] != 'ASSIGNED':
+            return assigned_date
```
ryneeverett: This condition will never be met since the method is only called `if issue['status'] == 'ASSIGNED'`.

bugwarrior/services/bz.py
```diff
@@ -260,9 +276,36 @@ def issues(self):
...
+
+        bug = self.bz.getbug(issue['id'])
+        history = bug.get_history()['bugs'][0]['history']
+
+        # this is already in chronological order, so the last change is the one we want
+        for h in history:
+            for change in h['changes']:
+                if change['field_name'] == 'status' and change['added'] == 'ASSIGNED':
+                  assigned_date = h['when']
```
ryneeverett: Why can't we just say `h = history[-1]` rather than looping?

ralphbean:
Closed via #648.  Thanks all!

bugwarrior/services/bz.py
djmitche: It looks like this comment was taken into account, but now the code only looks at the last update to the bug, and if *that* changed it to 'ASSIGNED', then records assigned_date.  If the last update is something else, for example adding a comment, then I think it won't find it.

PR #408, opened 2016-10-24
Document how to use systemd timers to run bugwarrior-pull.

bowlofeggs:


bugwarrior/docs/using.rst
```diff
@@ -15,6 +18,41 @@ adding the following to your crontab::
...
+    [Service]
+    Environment="DISPLAY=:0"
+    ExecStart=/usr/bin/bugwarrior-pull
+    Type=oneshot
+
+    [Install]
+    WantedBy=default.target
+    $ cat ~/.config/systemd/user/bugwarrior-pull.timer 
+    [Unit]
+    Description=Run bugwarrior-pull weekly and on boot
```
gdetrez: s/weekly/hourly/ ?

bowlofeggs: Good call - on my own machine (where I copied this from) I run it weekly, but I wanted it to be hourly to match the cron example. I'll fix that right up!

bowlofeggs: This is now fixed and ready for re-review!

Issue #409, opened 2016-10-31
Pivotal Tracker integration


hauleth:

ralphbean:
@hauleth @bruno-, can you give #666 a try and report on how it is working?

Issue #400, opened 2016-10-14
GLPI Support


humboldtux:
I know that [GLPI](https://github.com/glpi-project/glpi) is not a Forge software, but i would like to add support for GLPI tickets in bugwarrior.

ralphbean:
Going to close this (cleaning up the queue). See also #410 for anyone who wants to contribute this as a patch.

Issue #388, opened 2016-09-25
Support for rally.


iNode:
Is there any plans to add support for Rally as a remote source?

https://rally1.rallydev.com

python client library (python2 only for now):
https://github.com/RallyTools/RallyRestToolkitForPython

ralphbean:
Going to close this (cleaning up the queue). See also #410 for anyone who wants to contribute this as a patch.

Issue #356, opened 2016-07-19
Support for youtrack


mozey:
Is there any plans to add support for jetbrains youtrack as a remote source?

coddingtonbear:
There would be if you were planning to add such support, @mozey ;-); we really, really rely on people who actually use each supported service for writing the integration and helping out with testing during future releases.

They [do appear to have python bindings available](https://confluence.jetbrains.com/display/YTD6/Python+Client+Library), so adding a new service wouldn't be very difficult.  Check out [the services listed in the services directory](https://github.com/ralphbean/bugwarrior/tree/develop/bugwarrior/services) (the simplest of them is probably our [megaplan](https://github.com/ralphbean/bugwarrior/blob/develop/bugwarrior/services/mplan.py) integration) to get an idea of what's required; it's not as complex as it might sound, and I'm sure I can speak for all of us when I say that we'd be glad to help if you bump into any questions!

mozey:
@coddingtonbear I'll take a stab at it when I can squeeze in the time :)

I've setup the development environment (on OSX 10.10.5 with Python 2.7.11 virtualenv) as per the docs with some additional steps required:

tests_require in `setup.py` lists some of these not sure why `python setup.py develop` didn't take care of it?

    pip install nose
    pip install responses
    pip install jira
    pip install megaplan
    pip install pypandoc
    pip install pyac
    
Running the tests
    
    cd bugwarrior/tests 
    nosetests
    
    # Ran 39 tests in 2.080s
    # OK

ryneeverett:
@mozey When you invoke `python setup.py nosetests` those dependencies are installed into a separate testing environment so you don't need to install them into your development environment. The way you did it is fine too.

ryneeverett:
Resolved via #425.

PR #410, opened 2016-10-31
Add an FAQ

gdetrez:
There are regularly new requests to support other ITS (#87, #204, #311, #356, #388, #400, #409...) so I tried to compile old answer to such issues and put the result in a new FAQ in the docs.

### DIFF
--- a/bugwarrior/docs/faq.rst
+++ b/bugwarrior/docs/faq.rst
@@ -1,0 +1,15 @@
+FAQ
+===
+
+Can bugarrior support <some issue tracking system>?
+---------------------------------------------------
+
+Sure! But our general rule here is that we won't write a backend for a service
+unless we use it personally, otherwise it's hard to be sure that it really
+works.
+
+We also try to rely on people to become maintainers of the different backend
+plugins they use so that they don't suffer bit rot over time.
+
+In summary, we need someone who 1) uses <some issue tracking system> and 2) can
+develop the plugin.  Could it be you? :)
--- a/bugwarrior/docs/index.rst
+++ b/bugwarrior/docs/index.rst
@@ -40,6 +40,7 @@ Contents
    services
    configuration
    contributing
+   faq
 
 
 Indices and tables


ryneeverett:
What if we listed all the unimplemented services we need people to contribute in this FAQ section, and closed all those issues? When new issues like this are opened, we close them and add them to the list.

ryneeverett:
Or say something like "If you think a service is missing from this list, please open a pull request adding it."

gdetrez:
@ryneeverett I would rather keep the issues as they contain more metadata (who created it, when...), they provide an easy way to contact the person who requested the backend (for testing, contribution or just to know if it is still something they would want to use) and they provide a place for discussion.

On the other hand, I think it would be resonable to have a policy that such issue will be closed after X amount of time without activity.

gdetrez:
If I understand correctly, you want to keep a list, outside of the issue tracker, of all backends that have been requested by someone but not implemented? What would that be for?

ryneeverett:
There are two problems I'd like to see addressed: clutter in the issue tracker and clarity to potential contributors. I don't care what approach we take to addressing them.

**clutter**
To me, new service requests aren't really actionable tickets in the same sense that smaller feature or bug reports are. Implementing a new service is a pretty big job and it will get done when somebody wants it done enough to do it themselves. The only thing we need to discuss in the issue tracker is whether or not we will maintain support for the service. Usually this is pretty obvious, but there's also a huge grey area that we haven't really gotten into yet (#297, #398). It seems like the proportion of issues that are just "implement this new service" has been growing, which is presumably the catalyst of this PR.

**clarity**
If we're going to close some of the old issues, we should be unambiguous about whether we would maintain support for the service should someone contribute it. Closed issues are often interpreted as "wontfix". One approach would be a canned response ("Closing for inactivity, but we'd happily merge a PR...") but I'm not sure we can count on people actually reading the bottom comment. I would suggest that we could make ourselves perfectly clear and solicit new contributors by giving a similar level of visibility to "Services you can Contribute" as "Supported Services".

ryneeverett:
I'm going to go ahead and merge this, but I'd suggest you open an RFC so the team can hash out a policy before closing issues for "inactivity".

gdetrez:
Thanks for merging @ryneeverett.

I don't plan on closing issues for inactivity. I personally don't mind those issues actually, I was only suggesting it as an alternative way to clear the issue tracker from less-actionable items.

ryneeverett:
I think we're basically on the same page here -- "inactivity" isn't really a property of actionable tickets. If we had a mailing list, that would be a more appropriate place to discuss service requests. Adding them to a list is a hack to make unactionable items actionable.

PR #406, opened 2016-10-23
Add installation and test support for python3.

ryneeverett:
This is a rebase and update of @wfranzini's work in #387. We had a remarkable number of merge conflicts in such a short time.

This is great progress on #25, but I think we ought to do significant manual testing before closing that issue and declaring python3 compatibility 'done'.

PR #416, opened 2016-11-09
Fix broken tests.  tox, travis, and py2/py3

ralphbean:
A number of changes here.  See discussion in #406 as well as #384.


ralphbean:
A couple other changes got in here.  Trying to reconcile a couple different changes.

ralphbean:
And... they're passing again.  :green_apple:

ryneeverett:
Ah, so I broke the build because the latest release of `future` is buggy?

bugwarrior/db.py
```diff
@@ -25,8 +26,9 @@
 MARKUP = "(bw)"
 
 
-DOGPILE_CACHE_PATH = os.path.expanduser(
-    os.getenv('XDG_CACHE_HOME', '~/.cache') + '/dagd.dbm')
+PYVER = '%i.%i' % sys.version_info[:2]
+DOGPILE_CACHE_PATH = os.path.expanduser(''.join([
+    os.getenv('XDG_CACHE_HOME', '~/.cache'), '/dagd-py', PYVER, '.dbm']))
```
ryneeverett: Is the cache database really incompatible between python2/python3?

ralphbean: As far as I can tell, yes.  Will verify.

ryneeverett: The actual database format is different? That doesn't make much sense to me.

Or is it just a bytes-vs-strings issue where we could just convert bytes to unicode in python3?

ralphbean: Well, here's some info:


```python
# in python2
>>> import dogpile.cache
>>> cache = dogpile.cache.make_region().configure('dogpile.cache.dbm', arguments=dict(filename='foo.db'))
>>> cache.get_or_create('key', lambda: 'value')
'value'
```

and then...

```python
# in python3
>>> import dogpile.cache
>>> cache = dogpile.cache.make_region().configure('dogpile.cache.dbm', arguments=dict(filename='foo.db'))
>>> cache.get_or_create('key', lambda: 'value')
    cache.get_or_create('key', lambda: 'value')
  File "/usr/lib/python3.5/site-packages/dogpile/cache/region.py", line 825, in get_or_create
    async_creator) as value:
  File "/usr/lib/python3.5/site-packages/dogpile/lock.py", line 154, in __enter__
    return self._enter()
  File "/usr/lib/python3.5/site-packages/dogpile/lock.py", line 87, in _enter
    value = value_fn()
  File "/usr/lib/python3.5/site-packages/dogpile/cache/region.py", line 780, in get_value
    value = self.backend.get(key)
  File "/usr/lib/python3.5/site-packages/dogpile/cache/backends/file.py", line 220, in get
    with self._dbm_file(False) as dbm:
  File "/usr/lib64/python3.5/contextlib.py", line 59, in __enter__
    return next(self.gen)
  File "/usr/lib/python3.5/site-packages/dogpile/cache/backends/file.py", line 215, in _dbm_file
    "w" if write else "r")
  File "/usr/lib64/python3.5/dbm/__init__.py", line 88, in open
    raise error[0]("db type could not be determined")
dbm.error: db type could not be determined
```

```bash
# Inspecting the file created by python2
~❯ file foo.db
foo.db: Berkeley DB (Hash, version 9, native byte-order)
# inspecting the file created by python3
~❯ file foo.db
foo.db: GNU dbm 1.x or ndbm database, little endian, 64-bit
```

ralphbean: Ah ha, the [python 3 docs](https://docs.python.org/3/library/dbm.html) point to this [third party lib](https://www.jcea.es/programacion/pybsddb.htm) which explains that the berkeley db interface was included in py2.3 through py2.7 but is not included in python3.  I'll see if just installing it will do as a workaround.

ralphbean: Yeah, just installing `pybsddb` doesn't have any effect.  `dogpile.cache` would have to invoke it explicitly and it doesn't look like it has a flag to tell it to do so.

ralphbean:
Thanks @ryneeverett!

PR #417, opened 2016-11-09
Fix GitHub oracle in python3.

ryneeverett:
Based off #416.

### DIFF
--- a/bugwarrior/services/github.py
+++ b/bugwarrior/services/github.py
@@ -17,8 +17,10 @@ class GithubClient(ServiceClient):
         self.auth = auth
         self.session = requests.Session()
         if 'token' in self.auth:
-            authorization = 'token ' + self.auth['token']
-            self.session.headers['Authorization'] = authorization
+            token = self.auth['token']
+            if isinstance(token, bytes):  # python3
+                token = token.decode('utf-8')
+            self.session.headers['Authorization'] = 'token ' + token
 
     def get_repos(self, username):
         user_repos = self._getter(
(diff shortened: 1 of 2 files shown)


gdetrez:
This doesn't not work: if the token is `b"foobar"` the generated header value will be `token b'foobar'` (should be `token foobar`).

I would also suggest to move the fix "upstream": this might be an issue in other services so it might be safer to make sure that the oracle returns a `str` rather than making an conversion in the github service. It might be good to make the conversion explicit and make sure that it supports non ascii characters. (using system locale?)

It seem that this is based on a PR that wasn't merged, maybe rebase on develop for a cleaner pull request?

ryneeverett:
Rebased.

ryneeverett:
Oh, I see what you're saying.

```py
>>> 'token {token}'.format(token=b'token')
"token b'token'"
```

Hmm.

ryneeverett:
I guess that error message wasn't joking: You really "Can't convert 'bytes' object to str implicitly".

gdetrez:
What about moving the conversion to the oracle?

gdetrez:
Also, the included test is not very useful, as demonstrated by the initial version of the PR ;-)
It should probably check the value of the 'Authorization' header

ryneeverett:
Whoops, I overlooked that part of your comment. Agreed on both points, which leaves pretty much nothing of value here. I may not get back to this for a while so it might make sense to open an issue if you don't plan to tackle this yourself.

gdetrez:
I made a quick fix locally, so that I can use bw with my config, I'll try to add some tests and make it a proper PR.

PR #421, opened 2016-12-05
Fix @oracle:eval with Github and python 3

gdetrez:
The oracle now returns the password as a unicode string.  The password
is decoded using the UTF-8 coding as this is the python 3 default.


gdetrez:
I think I found the problem: when initializing BugwarriorData, task hangs because it waits for input.
I refactored the service to get a BugwarriorData instance from the config object (some sort of dependency injection I guess). That should make it easier to write proper unit tests.

Issue #360, opened 2016-08-01
Enterprize github support?


hjri:
Currently only github service available is github.com, but if one'd like to use their company's internal github service it wouldn't be possible.

irl:
It looks that GitHub enterprise supports the same APIs as github.com and all that would be needed is to change the API endpoint used in the existing GitHub service.

https://developer.github.com/v3/enterprise/

irl:
@hjri Do you know of any way that testing can be performed for this without going and paying for a GitHub Enterprise subscription?

hjri:
None that I know. I guess only way is to use user-based testing. I have access to one so i could test it a bit.

irl:
Our general rule here is that we won't write a backend for a service unless we use it personally (it's otherwise hard to be sure that it really works). However, we gladly accept new maintainers for new backends. Would you be interested in doing the work to introduce and maintain this feature?

http://bugwarrior.readthedocs.io/en/latest/contributing.html

hjri:
Thanks for the offer, but I'm not good at Python. I could sit for several evenings and make a PR for this issue, but I doubt code quality would be any good.

irl:
@hjri Ok, we can leave the issue open and hopefully there would be someone who can make this happen. It really doesn't look like it would be too difficult but without someone using this who is happy to maintain it and debug any issues, it could become problematic later.

gdetrez:
@hjri I tryed to work on this but I don't have access to a private github so I have no idea if it works or not.  I'd be happy if you could test it for me if you can.

You can install bugwarrior from [this branch](https://github.com/gdetrez/bugwarrior/tree/360-enterprize-github-support) (I suggest to do in in a virtualenv and to **backup your tasks first**):
```sh
$ pip install git+https://github.com/gdetrez/bugwarrior.git@360-enterprize-github-support 
```

Then there is a new option `github.host` where you should be able to put the hostname of your local github:
```ini
[workproject]
service = gihtub
github.host = github.acme.biz
github.login = john
...
```

hjri:
Seems to be kinda working, after screwing around and reading manual i got it to import PRs and issues, but I get this error after running `bugwarrior-pull`
```
CRITICAL:bugwarrior.services.github: Unrecognized issue URL: https://github.acme.biz/org/repo/pull/1337.
```
(domain/organization/repository/pr number have been replaced)

gdetrez:
[...]

PR #422, opened 2016-12-06
#360: Enterprize GitHub support

gdetrez:
Add support for GitHub Enterprize by adding a `github.host` configuration option (issue #360)

### DIFF
--- a/bugwarrior/services/github.py
+++ b/bugwarrior/services/github.py
@@ -1,6 +1,7 @@
 from builtins import filter
 import re
 import six
+from urllib.parse import urlparse
 
 import requests
 from six.moves.urllib.parse import quote_plus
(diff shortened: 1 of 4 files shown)


bugwarrior/services/github.py
```diff
@@ -278,29 +293,34 @@ def get_involved_issues(self, user):
...
-            if tag is None:
-                log.critical(" Unrecognized issue URL: %s.", url)
-                continue
-            issues[url] = (tag.group(1), issue)
+            try:
+                url = issue['html_url']
+                repo = self.get_repository_from_issue(issue)
+                issues[url] = (repo, issue)
+            except ValueError as e:
+                log.critical(e)
```
ryneeverett: What's causing the `ValueError` here?

gdetrez: The new method `get_repository_from_issue` defined a few lines bellow.

ryneeverett: Ah. I think moving the surrounding lines out of the try/except block would make that clear. Also I consider it best practice to have the bear minimum in the `try` block to avoid catching exceptions you didn't intend to, especially when using built-in exceptions. I'd suggest:

```py
url = issue['html_url']
try:
    repo = self.get_repository_from_issue(issue)
except ValueError as e:
    log.critical(e)
else:
    issues[url] = (repo, issue)
```

bugwarrior/services/github.py
```diff
@@ -278,29 +293,34 @@ def get_involved_issues(self, user):
...
-                '{owner}/{project}'.format(
-                    **match_dict
-                ),
-                issue
-            )
+            repos = self.get_repository_from_issue(issue)
+            issues[issue['url']] = (repos, issue)
         return issues
 
+    @staticmethod
```
ryneeverett: Can't a `staticmethod` be overriden just as easily?

```py
>>> class A(object):
...     @staticmethod
...     def foo():
...         print("bar")
... 
>>> class B(A):
...     @staticmethod
...     def foo():
...         print("baz")
... 
>>> B().foo()
baz
```

coddingtonbear: Can!  But the behavior of what happens when you call other methods in your class is different between the two.  From within a `@classmethod`, you can call other `@classmethods` by using `cls.whatever_method`, and your subclasses can override `whatever_method`.  When using `@staticmethod`, you can only call those methods by using the class name directly -- `MyClass.whatever_method`, and that prevents your subclasses from being able to alter that behavior.

ryneeverett: I see what you're saying now. I agree that calling the class by its explicit name in a static method would almost always be bad, but I disagree that there's no good reason to use them.

A `staticmethod` indicates to developers that the method probably does not access other class methods, class state, or instance state. I find this a helpful part of the method signature in indicating the function's purpose and role in the class.

Generally, `staticmethod` indicates that the method could just as well be a free function, but it is part of the class for organizational purposes. When a class has a lot of `staticmethod`'s, it's usually a sign that a lot of the methods would be better moved into a different module as free functions. With `classmethod`'s, this anti-pattern is much less obvious.

ryneeverett:
@gdetrez No pressure but it looks like this is being held up by trivial inline nitpicks and could be merged as soon as you respond and/or address them.

gdetrez:
You're right, I may have forgot about this one. I'll push some fixes 
toward the end of the week!

gdetrez:
I turned the method into a `classmethod`. `urljoin` didn't work so well because it's for building absolue urls from relative ones, not to build paths from components, i.e. this is what I get:
```python
> urljoin("https://github.example.com/api/v3", "/path/to/something")
'https://github.example.com/path/to/something'
```

gdetrez:
I fixed the last comment (about the try block). I also rebased to try to fix the jira problems on travis...

gdetrez:
It looks like a new version of the jira library is breaking the test (so, not me :+1:). I re-ran the build on `develop` and I get the same result...

ryneeverett:
Looks like @ralphbean fixed jira a few days ago, not that that was really a blocker. I'd recommend a squash but I wont press the issue.

gdetrez:
Rebased and squashed! I'll menge once travic is finished.

ryneeverett:
Thanks, I forgot we would be causing you rebase work.

PR #425, opened 2016-12-22
Add support for YouTrack issue tracker.

babelop:
* Implementation of Feature Proposal #356 (Support for youtrack).
* Additional features include tag importing and query customization.
* PR includes new service, appropriate tests, and documentation.

### DIFF
--- a/bugwarrior/docs/services/youtrack.rst
+++ b/bugwarrior/docs/services/youtrack.rst
@@ -23,29 +23,28 @@ or described in `Service Features`_ below.
 Service Features
 ----------------
 
-Host Options
-+++++++++++++++++++++
+Customize the YouTrack Connection
++++++++++++++++++++++++++++++++++
 
 The ``youtrack.host`` field is used to construct a URL for
 the YouTrack server. It defaults to a secure connection scheme (HTTPS)
 on the standard port (443).
 
-Connecting on a different port::
+To connect on a different port, set::
 
     youtrack.port = 8443
 
-Allow server to use a self-signed SSL certificate::
-
-    youtrack.verify_ssl = False
-
-To use an insecure HTTP connection, and standard port 80::
+If your YouTrack instance is only available over HTTP, set::
 
     youtrack.use_https = False
 
-Query Options
-+++++++++++++++++++++
+If you want to ignore verifying the SSL certificate, set::
 
-The YouTrack service pulls issues based on the results of a customizable query.
+    youtrack.verify_ssl = False
+
+Specify the Query to Use for Gathering Issues
++++++++++++++++++++++++++++++++++++++++++++++
+
 The default option selects unresolved issues assigned to the login user::
 
     youtrack.query = for:me #Unresolved
@@ -58,12 +57,10 @@ Queries are capped at 100 max results by default, but may be adjusted to meet yo
 
     youtrack.query_limit = 100
 
-Import Tags
-+++++++++++++++++++++
+Import Issue Tags
++++++++++++++++++
 
-The YouTrack issue tracker allows you to tag issues. You may
-apply those tags in Taskwarrior by using the ``youtrack.import_tags``
-option::
+The YouTrack issue tracker allows you to tag issues. To apply these tags in Taskwarrior, set::
 
     youtrack.import_tags = True
 
--- a/bugwarrior/docs/services/youtrack.rst
+++ b/bugwarrior/docs/services/youtrack.rst
@@ -1,0 +1,103 @@
+YouTrack
+========
+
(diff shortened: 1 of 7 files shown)


bugwarrior/docs/services/youtrack.rst
```diff
@@ -0,0 +1,103 @@
...
+the YouTrack server. It defaults to a secure connection scheme (HTTPS)
+on the standard port (443).
+
+Connecting on a different port::
+
+    youtrack.port = 8443
+
+Allow server to use a self-signed SSL certificate::
+
+    youtrack.verify_ssl = False
```
ryneeverett: Is adding a self-signed cert to the system trust store too inconvenient? Seems like we should at least mention here in the docs that the server must be in a trusted network for this to be secure. Also, would it make sense for this option to accept a `/path/to/certfile` to pass through to requests?

bugwarrior/services/youtrack.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def _get_record_field(self, field_name):
+        for field in self.record['field']:
+            if field['name'] == field_name:
+                return field
+
+    def _get_record_field_value(self, field_name, field_value='value'):
+        field = self._get_record_field(field_name)
+        if field:
+            return field[field_value]
```
ryneeverett: `_get_record_field` seems redundant.

```py
def _get_record_field_value(self, field_name, field_value='value'):
    for field in self.record['field']:
        if field['name'] == field_name:
            return field[field_value]
```

bugwarrior/docs/services/youtrack.rst
ryneeverett: To be clear, my concern is that a naive user would interpret this as implying that self-signed certs do not need to be verified.

babelop: In terms of consistency within the Bugwarrior project, the option as implemented matches Jira and Gitlab:
  https://github.com/ralphbean/bugwarrior/search?utf8=%E2%9C%93&q=verify_ssl&type=Code

* From the perspective of whether or not the option should exist, vs requiring a trust-store update, I'm deferring to those services precedent.
* Implementation of a custom certfile parameter may be best served as a separate feature request that could then be applied uniformly to all three major issue trackers.

In terms of the documentation, that can be reworded if desired. The [Jira](https://github.com/ralphbean/bugwarrior/blob/develop/bugwarrior/docs/services/jira.rst) docs for `verify_ssl` are perhaps more 'dangerous' in its suggestion to disable it in response to any verification error. The [Gitlab](https://github.com/ralphbean/bugwarrior/blob/develop/bugwarrior/docs/services/gitlab.rst) docs use a more clinical description. I aimed for something in-between with YouTrack.

In the end it comes down to Bugwarrior's target audience ... particularly users of YouTrack/Gitlab/Jira which are traditonally self-hosted solutions. Persons using a local install are unlikely to be using SSL, and in a corporate environment it's unlikely you'd encounter a self-signed certificate unless the workstations were IT managed and had said certificates imported. So in my opinion users that need to use this option would know what they are doing, either because of their technical background, or because they setup the target issue tracker with the self-signed certificate to begin with.

I'm happy to elaborate on the security risks for this option in the docs if you think it's appropriate to the project's scope though.

bugwarrior/services/youtrack.py
babelop: Admittedly the function is redundant for the set of features currently implemented, but if you'll take a look at the sample output from a YouTrack REST endpoint
https://gist.github.com/babelop/d4afacfb0c753c2fc3cd6b7f702e0b78

Complex field types include extra meta-data like valueId and color stored at the same hierarchal level as value, and would require multiple trips through the array if using the "value" convenience method alone. The ability to return the whole record block is ideal for future use in consuming, for example, priority data.

However, I defer to your preference if you'd like to see it pruned at this stage.

bugwarrior/docs/services/youtrack.rst
ryneeverett: Rock solid analysis. What would you think about standardizing the clinical Gitlab description across services?

bugwarrior/services/youtrack.py
ryneeverett: Fair enough.

bugwarrior/docs/services/youtrack.rst
babelop: @ryneeverett I added a new commit to the PR standardizing documentation of the `verify_ssl` option across those three trackers.

ryneeverett:
FYI I'm going to leave this open for a bit for visibility. Somebody else may want to review.

PR #430, opened 2017-01-12
We should verify by default here.

ralphbean:
Right?


ryneeverett:
I don't know what the JIRA client does with `None` -- it may default to `True` anyway. But this certainly seems safer.

ralphbean:
Yeah, plus our use of [asbool](https://github.com/ralphbean/bugwarrior/blob/develop/bugwarrior/config.py#L22-L26) coerces `None` to `False`.  Thanks @ryneeverett!

PR #435, opened 2017-01-18
Include the issue url for jira tasks even if there are no annotations.

ralphbean:
Not sure why this we didn't include these to begin with.

### DIFF
--- a/bugwarrior/services/jira.py
+++ b/bugwarrior/services/jira.py
@@ -288,8 +288,7 @@ class JiraService(IssueService):
         IssueService.validate_config(config, target)
 
     def annotations(self, issue, issue_obj):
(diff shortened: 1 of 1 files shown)


bugwarrior/services/jira.py
```diff
@@ -290,16 +290,13 @@ def validate_config(cls, config, target):
     def annotations(self, issue, issue_obj):
         comments = self.jira.comments(issue.key)
```
jdcasey: Seems like the CI build is failing now when there are no comments, since comments = None. Could we fix it by setting this with:

```
comments = self.jira.comments(issue.key) or []
```

?

ralphbean: Yeah, that should do it.

PR #437, opened 2017-01-29
Allow unlimited description and annotation lengths. Resolve #441.

### LINKED ISSUE #441: Support for unbounded length annotations?
I am *loving* bugwarrior, not least because rather than having to visit 3rd-party tools I can see everything I'm doing right in TaskWarrior. It's seriously made an enormous impact to both my workflow and my enjoyment of getting things done in general. However, I'm having a tricky time finding a clean way to allow unbounded length annotations.

I was hoping that setting `annotation_length = 0` would allow annotations to be of any length; instead it appears to reduce annotations to the shortest length possible, which is the opposite of what I'm after.

Right now I've got `annotation_length` set to a very large number, but that doesn't feel like a very clean way of doing things.

Is there a way to tell bugwarrior to use unbounded length annotations, or if not, may I humbly suggest `annotation_length = 0` be a way of doing this?

ryneeverett:
<s>This is alternative proposal to #436.

The advantage to subclassing ConfigParser rather than using
allow_no_value is that we don't have to worry about handling None for
other value types.</s>

### DIFF
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -202,7 +202,7 @@ def get_config_path():
 
 
 def load_config(main_section, interactive=False):
-    config = ConfigParser({'log.level': "INFO", 'log.file': None})
+    config = BugwarriorConfigParser({'log.level': "INFO", 'log.file': None})
     path = get_config_path()
     config.readfp(codecs.open(path, "r", "utf-8",))
     config.interactive = interactive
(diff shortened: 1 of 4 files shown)


ryneeverett:
@pjf I was actually just about to close this in favor of #438. I wrote this patch as an alternative to #438 because I thought that approach would require setting `allow_no_value=True`. Well we both just found out that that was a false dilemma.

ryneeverett:
Closing in favor of #438.

pjf:
@ryneeverett : I agree with you here! I think I was just so excited to see *test cases* that I clicked the review button out of sheer joy. :)

ryneeverett:
@pjf Yeah, I really should write tests for this feature.

PR #438, opened 2017-01-29
[WIP] Allow limitless description and annotation lengths (option B) Resolve #441.

### LINKED ISSUE #441: Support for unbounded length annotations?

ryneeverett:
This is presented as another alternative to #436.

<s>This could lead to more-difficult-to-debug errors if people pass no
value to other configuration options, because the error will likely be
in bugwarrior failing to handle None rather than ConfigParser raising a
validation error on their config file. However, I don't see a clear
alternative. I did play around with subclassing ConfigParser to just
allow no value for integers, but the extra complication might not be
worth it.</s> See second commit.

- [ ] Tests

### DIFF
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -202,7 +202,7 @@ def get_config_path():
 
 
 def load_config(main_section, interactive=False):
-    config = ConfigParser({'log.level': "INFO", 'log.file': None}, allow_no_value=True)
+    config = ConfigParser({'log.level': "INFO", 'log.file': None})
     path = get_config_path()
     config.readfp(codecs.open(path, "r", "utf-8",))
     config.interactive = interactive
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -202,7 +202,7 @@ def get_config_path():
 
 
 def load_config(main_section, interactive=False):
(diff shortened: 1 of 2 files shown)


ryneeverett:
According to [this comment](https://github.com/ralphbean/bugwarrior/issues/426#issuecomment-281250889) my use of `allow_no_value` is probably misguided.

ryneeverett:
Good thing I tried to test this. I should have realized that I still need the modified version of `getint` if we're going to accept a null value because an empty string is still not an integer. Closing in favor of #437.

PR #440, opened 2017-02-13
Test and fix our usage of different python-jira versions.

ralphbean:
The issue is that they added a new keyword argument to a method that we override.  Our method signature didn't match.

I dropped the details to try and immunize ourself against any future changes.


coddingtonbear:
:heart:

PR #238, opened 2017-02-24
Support Bugzilla API keys (pending support in python-bugzilla)

djmitche:
python-bugzilla refers to these as [tokens](https://github.com/python-bugzilla/python-bugzilla/commit/e1af7160b3162d423fdfeb04e230067a75e9b584).  It would be nice to be able to create an API key and provide it as

```
[bugzilla_mozilla]
bugzilla.username = dustin@mozilla.com
bugzilla.token = e1af7160b3162d423fdfeb04e230067a75e9b584
```

### DIFF
--- a/bugwarrior/services/bz.py
+++ b/bugwarrior/services/bz.py
@@ -125,11 +125,17 @@ class BugzillaService(IssueService):
         # to pass that argument or not.
         self.advanced = asbool(self.config.get('advanced', 'no'))
 
-        self.password = self.get_password('password', self.username)
-
         url = 'https://%s/xmlrpc.cgi' % self.base_uri
-        self.bz = bugzilla.Bugzilla(url=url)
-        self.bz.login(self.username, self.password)
+        api_key = self.config.get('api_key', default=None)
+        if api_key:
+            try:
+                self.bz = bugzilla.Bugzilla(url=url, api_key=api_key)
+            except TypeError:
+                raise Exception("Bugzilla API keys require python-bugzilla>=2.1.0")
+        else:
+            password = self.get_password('password', self.username)
+            self.bz = bugzilla.Bugzilla(url=url)
+            self.bz.login(self.username, password)
 
     @staticmethod
     def get_keyring_service(service_config):
@@ -139,10 +145,12 @@ class BugzillaService(IssueService):
 
     @classmethod
     def validate_config(cls, service_config, target):
-        req = ['username', 'password', 'base_uri']
+        req = ['username', 'base_uri']
         for option in req:
             if option not in service_config:
                 die("[%s] has no 'bugzilla.%s'" % (target, option))
+        if 'password' not in service_config and 'api_key' not in service_config:
+            die("[%s] has neither 'bugzilla.password' nor 'bugzilla.api_key'" % (target,))
 
         super(BugzillaService, cls).validate_config(service_config, target)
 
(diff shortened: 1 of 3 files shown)


djmitche:
Look, ma, now I'm a pull request!

I've added support for api keys in python-bugzilla/python-bugzilla#5, although it is not in a released version yet.  Still, I tested this with a temporary install of the latest git master and it worked OK both with password and api key auth.  Once a new release is made, I would propose requiring that version in `setup.py`, but I'm also OK leaving it unspecified in `setup.py` and keeping the exception when using a too-old version.

djmitche:
I just got word from @crobinso that he'll cut a new release of python-bugzilla in a few weeks, at which point we can depend on that version.

ryneeverett:
Looks like 2.0.0 was released a few days ago.

djmitche:
(rebased onto latest develop)

djmitche:
(rebased again, and a requirement for the newly-released python-bugzilla>=2.1.0 added)

ready for review and merging, I think!

bugwarrior/services/bz.py
```diff
@@ -125,11 +125,17 @@ def __init__(self, *args, **kw):
...
-        self.bz = bugzilla.Bugzilla(url=url)
-        self.bz.login(self.username, self.password)
+        api_key = self.config.get('api_key', default=None)
+        if api_key:
+            try:
+                self.bz = bugzilla.Bugzilla(url=url, api_key=api_key)
+            except TypeError:
+                raise Exception("This version of python-bugzilla does not support api keys")
+        else:
+            password = self.config.get('password', self.username)
```
ryneeverett: Shouldn't this still be `self.get_password`?

bugwarrior/services/bz.py
```diff
@@ -139,10 +145,13 @@ def get_keyring_service(service_config):
...
-        for option in req:
-            if option not in service_config:
-                die("[%s] has no 'bugzilla.%s'" % (target, option))
+        keys = ['username', 'password', 'api_key', 'base_uri']
+        if 'base_uri' not in service_config:
+            die("[%s] has no 'bugzilla.base_uri'" % (target,))
+        if 'username' not in service_config:
+            die("[%s] has no 'bugzilla.username'" % (target,))
+        if not ('password' in service_config or 'api_key' in service_config):
+            die("[%s] has no 'bugzilla.password' or 'bugzilla.api_key'" % (target,))
```
ryneeverett: The `keys` variable appears to be unused. My preference would be to stick with the loop where it works since that makes modifications to the required configuration values easy:

```py
req = ['username', 'base_uri']
for option in req:
    if option not in service_config:
        die("[%s] has no 'bugzilla.%s'" % (target, option))

if not ('password' in service_config or 'api_key' in service_config):
    die("[%s] has no 'bugzilla.password' or 'bugzilla.api_key'" % (target,))
```

setup.py
```diff
@@ -47,7 +47,7 @@
           activecollab=["pypandoc", "pyac>=0.1.5"],
           bts=["PySimpleSOAP","python-debianbts>=2.6.1"],
           trac=["offtrac"],
-          bugzilla=["python-bugzilla"],
+          bugzilla=["python-bugzilla>=2.1.0"],
```
ryneeverett: Do we have to add this constraint? Couldn't one continue using password authentication with older versions?

ryneeverett:
By the way, thanks for persisting with this so long. My comments are regarding diffs that look far different from when you originally wrote the patch.

djmitche:
No worries - it was more than a trivial rebase, so it definitely deserved another look.

djmitche:
Branch updated

PR #448, opened 2017-02-27
Check for duplicated new issues. Fix #447

### LINKED ISSUE #447: Duplicate issues stored to taskwarrior database.
bugwarrior version: 1.4.0

I have multiple github services (one for my personal repos and one for each organization in which I am involved). The issues of my personal repos are duplicated.

My `.bugwarriorrc`:
```
[general]
targets = edigiacomo_github, arpae_simc_github, r_map_github
inline_links = False
annotation_links = False
annotation_comments = False
legacy_matching = False

[edigiacomo_github]
service = github
only_if_assigned = edigiacomo
github.username = edigiacomo
github.login = edigiacomo
github.password = ...
github.include_repos = django-statusboard, arkimet-tools
add_tags = github, edigiacomo_github

[arpae_simc_github]
service = github
only_if_assigned = edigiacomo
github.username = arpa-simc
github.login = edigiacomo
github.password = ...
github.include_repos = meteo-vm2, wreport, dballe, arkimet, simc-opendata-examples, arkimet-postprocessor-suite
add_tags = github, arpae_simc_github

[r_map_github]
service = github
only_if_assigned = edigiacomo
github.username = r-map
github.login = edigiacomo
github.password = ...
github.include_repos = rmap
add_tags = github, r_map_github
```

The first time that I run `bugwarrior-pull`, the issues of the repo `edigiacomo/django-statusboard` are repeated 3 times (one for each github service).

Playing with the code, I found that the "directly assigned issues" (`GithubService.get_directly_assigned_issues`) [are always retrieved](https://github.com/ralphbean/bugwarrior/blob/1.4.0/bugwarrior/services/github.py#L278). Each process has its own copy of these issues (it seems that https://api.github.com/user/issues return issues from my personal repos only). When the program iterates over all the issues found, it doesn't detect that some issues are duplicated and so they are added multiple times.

Maybe am I missing something?

Thank you!

edigiacomo:
The task list is checked before adding the new issue, because two or more
services could have retrieved the same issue.

Fix #447

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -366,6 +366,20 @@ def synchronize(issue_generator, conf, main_section, dry_run=False):
     # Add new issues
     log.info("Adding %i tasks", len(issue_updates['new']))
     for issue in issue_updates['new']:
+        # Check for duplicated new issues
+        try:
+            existing_uuid = find_local_uuid(
+                tw, key_list, issue, legacy_matching=legacy_matching
+            )
+        except NotFound:
+            pass
+        except MultipleMatches:
+            pass
+        else:
+            log.info("Task %s was already adedd, skipping",
+                     issue['description'])
+            continue
+
         log.info("Adding task %s%s",
             issue['description'], notreally)
         if dry_run:
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -126,9 +126,28 @@ class TestSynchronize(ConfigTest):
              'pending': []})
 
         # TEST DUPLICATED NEW ISSUE.
+
+        issue = {
+            'description': 'Twice!',
+            'githubtype': 'issue',
+            'githuburl': 'https://foo.bar',
+            'priority': 'M',
+        }
+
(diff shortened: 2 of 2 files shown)


edigiacomo:
I have tagged the PR as WIP because the Travis test fails...

ryneeverett:
By the way, thanks for the pull. While I'm still not sure what the solution is, your PR explained the problem much better than words.

edigiacomo:
The issues already in the database are not ignored because they aren't in the "new" issues list, but in the "changed" or "existing" list.

The current behaviour in the PR is: if an issue is returned from more than one service and is already in the database, the "changed" list contains all the duplicated issues: the resulting tag list is the union of all the tags (if `merge_tags`) and the last description "wins".

But the problem occurs when an issue is not already in the database and it is returned from more than one service: in this case, the duplicates are ignored and - you're right - this is not good because some updates can be missed.

So, yeah, the PR needs a lot of improvements :)

PR #452, opened 2017-03-02
Implement `github.query` for the github service

### LINKED ISSUE #423: Support GitHub reviewer functionality
Currently, a pull request will show up in one's task list if `filter_pull_requests = True` is set and that user is marked as "Assigned". But if you are marked as "Reviewer", and not "Assigned", the pull request won't show up in the task list. Docs on [Review Requests](https://developer.github.com/v3/pulls/review_requests/).

djmitche:
Fixes #423 and then some by allowing arbitrary github queries.

I'm not sure this is the best way to do the tests -- it repeats a lot of boilerplate just to run the service with slightly different parameters.

### DIFF
--- a/bugwarrior/services/github.py
+++ b/bugwarrior/services/github.py
@@ -3,6 +3,7 @@ import re
 import six
 
 import requests
+import urllib
 from jinja2 import Template
 
 from bugwarrior.config import asbool, aslist, die
@@ -28,9 +29,10 @@ class GithubClient(ServiceClient):
                 username=username))
         return user_repos + public_repos
 
-    def get_involved_issues(self, username):
-        tmpl = "https://api.github.com/search/issues?q=involves%3A{username}%20state%3Aopen&per_page=100"
-        url = tmpl.format(username=username)
+    def get_query(self, query):
+        """Run a generic issue/PR query"""
+        tmpl = "https://api.github.com/search/issues?q={query}&per_page=100"
+        url = tmpl.format(query=urllib.quote_plus(query.encode('utf-8')))
         return self._getter(url, subkey='items')
 
     def get_issues(self, username, repo):
@@ -226,7 +228,6 @@ class GithubService(IssueService):
         super(GithubService, self).__init__(*args, **kw)
 
         self.login = self.config_get('login')
-        self.username = self.config_get('username')
 
         auth = {}
         token = self.config_get_default('token')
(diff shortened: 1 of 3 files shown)


djmitche:
Tested by hand.  The docs update tries to distinguish `github.login` from `github.username` a little.

ryneeverett:
> I'm not sure this is the best way to do the tests -- it repeats a lot of boilerplate just to run the service with slightly different parameters.

I'd suggest moving the `arbitrary_` class properties into global variables.

djmitche:
Good point - updated accordingly.

bugwarrior/services/github.py
```diff
@@ -384,6 +399,13 @@ def issues(self):
         log.debug(" Found %i issues.", len(issues))
         issues = list(filter(self.include, issues.values()))
         log.debug(" Pruned down to %i issues.", len(issues))
+        return issues
+
+    def issues(self):
+        if self.query:
+            issues = self._issues_by_query()
+        else:
+            issues = self._issues_by_filters()
```
ryneeverett: I would think the new `query` option would trigger additional task aggregation rather than act as an alternative to the existing API.

djmitche:
The python-bugzilla was a failure to operate `git` correctly, sorry.

I'm not totally sure what you mean by "additional task aggregation".  I suspect you mean that the results of the query would be added to the results of the other filters.  I really don't want that, since there's no clear, simple way to disable all of those other filters such that i *only* get the results of my query.  In fact, looking at the source, I think the only way to do this is to set `involved_issues` to False (since it is not limited to repos) and then set `include_repos` to an empty list (and I'm not sure how to do that in the config file).

Maybe this gets to the vague unease i have with the current design: I think there's an idea that, basically, Bugwarrior users want to sync assigned things into TaskWarrior and maybe there's a few options to tweak what "assigned" means.  My concept is that users should be able to indicate exactly which things they want to sync into TaskWarrior.  This query functionality is very definitely in the latter category, and not merely a tweak to "assigned things".  My intent (and it's been realized for `bugzilla.query` already) is to allow both: a bunch of common options; *or* a `query` option where the user specifies exactly the things of interest.

ryneeverett:
That is indeed what I meant, and agree with everything you said there.

In fact, I would like to go further in correcting this design problem. Rather than essentially adding an "advanced configuration" setting as an alternative to the "basic configuration" settings, I would rather see some of the existing settings deprecated in favor of a default query that does the same thing. However, this patch is probably doing enough already.

ryneeverett:
My previous comment was vague and not well though out, so I made a [pull request](https://github.com/djmitche/bugwarrior/pull/1) against your branch.

My premise is that we'd be better off making improvements to our existing configuration model than maintaining support for two configuration models.

The changes I made to the test configuration explain how to achieve the behaviour you want. I can understand if you wouldn't be thrilled about having to set those flags to `False`, but to me that's a better trade off than having the `query` option dynamically change the default values to `False`, which is what this PR currently effectively does. I might be receptive to changing those default values to `False`, but dynamic default values seem too confusing.

djmitche:
OK, updated -- thanks for the PR!

djmitche:
Sorry if I'm mixed up on the protocol -- is this waiting for something to merge?  @ralphbean perhaps?

ryneeverett:
There's no particular protocol, but I often leave PR's open for a bit in case somebody feels like providing a second review. I also generally wait a bit longer when development is moving slowly (little risk of merge conflicts), when the changes are major (this is a significant change to the most widely used service), and when I am a contributor to the patch (I shouldn't really be reviewing my own code).

djmitche:
OK, cool.  I'm just eager to use this (and the new bugzilla API key support...)

ralphbean:
Sorry to be MIA.  Let's merge!  :)

PR #453, opened 2017-03-11
Service config refactor.

ryneeverett:
The `IssueService.config_*` methods are part of a concept that wants to be it's own
object. The functionality of these convenience methods is
duplicated wherever an `IssueService` instance is not available
(validate_config, get_keyring_service, etc.), indicating that they're in
the wrong place.

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -8,7 +8,8 @@ from lockfile.pidlockfile import PIDLockFile
 import getpass
 import click
 
-from bugwarrior.config import get_data_path, get_keyring, load_config
+from bugwarrior.config import (
+    get_data_path, get_keyring, load_config, ServiceConfig)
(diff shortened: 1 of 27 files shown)


gdetrez:
I think this is good to merge as it is. :+1: 
For extra pythonic fanciness, you could also add `ServiceConfig.__contains__` as an alias for `has` so you can write `"username" in self.config` instad of `self.config.has("username")`

ralphbean:
Looks like this needs to be rebased on `develop`.

ryneeverett:
Rebased and switched over to the `__contains__` syntax. I found the `has`/`has_option` distinction confusing and that solved it.

However, I'm still not fond of masking the `ConfigParser.get` method with a different API. I thought about implementing `__getitem__` instead but that wouldn't handle cases where kwargs are needed. I suppose I could just rename the method. Any name suggestions? Any better ideas?

Issue #336, opened 2016-06-26
Notifications fail to work with pynotify (on Debian stretch)


irl:
Notifications seem to cause an exception with pynotify. They work fine with gobject though.

This is using Debian stretch, exception and installed package versions below:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/bugwarrior/command.py", line 56, in pull
    synchronize(issue_generator, config, main_section, dry_run)
  File "/usr/lib/python2.7/dist-packages/bugwarrior/db.py", line 366, in synchronize
    send_notification(issue, 'Created', conf)
  File "/usr/lib/python2.7/dist-packages/bugwarrior/notifications.py", line 93, in send_notification
    import pynotify
  File "/usr/lib/python2.7/dist-packages/gtk-2.0/pynotify/__init__.py", line 1, in <module>
    from _pynotify import *
ImportError: could not import gobject (could not find _PyGObject_API object)
```

```
+++-==================-==============-==============-==========================================
ii  python-gobject     3.20.1-1       all            Python 2.x bindings for GObject - transiti
ii  python-gobject-2   2.28.6-12+b1   amd64          deprecated static Python bindings for the 
ii  python-notify      0.1.1-4        amd64          Python bindings for libnotify
```

urzds:
I just ran into the same issue.

@irl: Did you find a workaround for this?

irl:
@urzds looking at my bugwarriorrc, I see I turned off notifications. I did get it to work with the not pynotify method iirc though.

urzds:
I changed my bugwarriorrc to `backend = gobject` and applied PR #419, now it works.

PR #456, opened 2017-03-15
Remove pynotify notifications.

ralphbean:
According to #336, they don't work.  It also looks like they just use
gobject notifications under the hood.. and we support that directly.

No use keeping an option around that doesn't work, right?  ;)

"Fixes" #336.


bugwarrior/notifications.py
```diff
@@ -89,23 +94,7 @@ def send_notification(issue, op, conf):
...
-                issue['description']
-        else:
-            message = "%s task: %s" % (op, issue['description'])
-            metadata = _get_metadata(issue)
-            if metadata is not None:
-                message += metadata
-
-        pynotify.Notification("Bugwarrior", message, logo_path).show()
-    elif notify_backend == 'gobject':
+    elif notify_backend == 'gobject' or notify_backend == 'pynotify':
```
ryneeverett: nitpick: Would it be better to set `notify_backend = 'gobject'` after the warning so this deprecated setting is handled in just one place? Seems simpler to cleanup if and when it's removed and less confusing if references to historical features are all in one place.

ralphbean: Yes, good call.

PR #458, opened 2017-03-22
Decode all byte strings from utf8 before any db actions.

### LINKED ISSUE #350: Unicode in task description causes logging to throw exception
When a task contains non-ascii characters, logging will throw an exception and not display `INFO:bugwarrior.db:Updating task ...`.

```
  File "/usr/lib/python2.7/logging/__init__.py", line 861, in emit
    msg = self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 37: ordinal not in range(128)
```

ralphbean:
Fixes #350.

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -325,6 +325,20 @@ def synchronize(issue_generator, conf, main_section, dry_run=False):
     }
 
     for issue in issue_generator:
+
+        # We received this issue from The Internet, but we're not sure what
+        # kind of encoding the service providers may have handed us. Let's try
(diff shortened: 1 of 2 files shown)


ralphbean:
Eh, this broke the world.  :(

PR #468, opened 2017-04-08
Add product and component as UDA for bugzilla

nblock:
This allows to use a ```project_template``` with a value of
```{{ bugzillaproduct }}.{{ bugzillacomponent }}``` which is nicely rendered
by taskwarrior as:
```
  $ task projects
  Project                      Tasks
  product                         10
    component1                     1
    component2                     7
    componentN                     2
```

Bugzilla has more fields which might be useful as UDA:
* 5.0: https://bugzilla.readthedocs.io/en/5.0/using/understanding.html
* 4.4: https://www.bugzilla.org/docs/4.4/en/html/bug_page.html


nblock:
Rebased onto current develop.

PR #469, opened 2017-04-12
Use service prefix for field templates

nblock:
Field templates should be loaded with the corresponding service prefix:
```bugzilla.description_template``` is the ```description_template``` for the
Bugzilla service.

At least the documentation (code comments and user docs) needs to
updated before this can be merged. Am I on the right track with this PR?

See: #466

### DIFF
--- a/bugwarrior/docs/common_configuration.rst
+++ b/bugwarrior/docs/common_configuration.rst
@@ -63,8 +63,6 @@ service example:
   that are not assigned to anybody.
 * ``default_priority``: Assign this priority ('L', 'M', or 'H') to
   newly-imported issues.
-* ``<fieldname>_template``: Generate the value of a field using a template.
-  See `Field Templates`_ for more details.
 * ``add_tags``: A comma-separated list of tags to add to an issue.  In most
   cases, this will just be a series of strings, but you can also make
   tags by defining one of your tags following the example set in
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ryneeverett** [comment]
I'm not quite sure why this is a WIP. Looks ready to me.

**Comment by nblock** [comment]
OK, I just found a missing service prefix for Trac. Updated the commit accordingly and removed the WIP marker.

**Comment by djmitche** [comment]
This was a substantial, breaking change -- it probably deserved more attention in the release notes than just including it in a list of PRs, and per semver deserved a major version bump.

To be honest, I don't see the motivation -- it made sense to me that in a particular section, the common options were given without a service prefix, while the service-specific options were given with a service prefix:

```
[bugzilla_mozilla_respond]
service = bugzilla
bugzilla.base_uri = bugzilla.mozilla.org
bugzilla.ignore_cc = True
add_tags = bugzilla, respond
project_template = moz
```
the first two are specific to bugzilla, and the last two are common.

ryneeverett:
@djmitche This PR is just documenting a decision that was made long ago in #93: "Standardizes configuration file key names such that they are always prefixed by the service name. (this will
allow people to use [default] if they're feeling particularly clever)." I think @coddingtonbear has given more detailed rationale in subsequent issues but I can't track it down at the moment.

djmitche:
I can understand the feature improvement. But it resulted in a whole bunch
of my tasks disappearing from reports when their project changed. If the
old config continued to work but generated a warning, then a minor version
bump would have been ok, but the sudden change is pretty jarring.

ryneeverett:
@djmitche Fair point. I think we've been fairly careful to raise such warnings for the per-service configuration options in the validation method of most services, but we may have neglected to do so for the `_template` options altogether.

PR #477, opened 2017-05-20
GMail service for bugwarrior

cthulahoops:
Add a service using Google APIs to fetch e-mails as tasks. You
can specify a query using the google search syntex to specify
which e-mails to fetch. Authentication is performed via oauth
and will require the user to set up a production and download
a client_secret file.

The default query is starred threads. No support for pinning and
snoozing yet as I don't use this. I think support for google tasks
could be extended from this easily.

### DIFF
--- a/bugwarrior/docs/configuration.rst
+++ b/bugwarrior/docs/configuration.rst
@@ -192,11 +192,4 @@ Example Configuration
     [my_gmail]
     service = gmail
     gmail.query = label:action OR label:readme
-    # You need a get a client_secret.json. Follow the instructions on: https://developers.google.com/gmail/api/quickstart/python
-    gmail.client_secret_path = ~/.config/bugwarrior/client_secret.json
-    # This file will be created when you authenticate.
-    gmail.credentials_path = ~/.config/bugwarrior/credentials.json
-    # Restrict to a particular e-mail address. This is optional and will cause the pull to fail if the credentials don't match
-    # the specified user. When working with multiple accounts it's a bit a tricky to make sure you always authorise the correct
-    # account.
-    gmail.login_name = name@example.com
+    gmail.login_name = you@example.com
--- a/bugwarrior/docs/configuration.rst
+++ b/bugwarrior/docs/configuration.rst
@@ -188,3 +188,15 @@ Example Configuration
     activecollab2.key = your-api-key
     activecollab2.user_id = 15
     activecollab2.projects = 1:first_project, 5:another_project
+
+    [my_gmail]
+    service = gmail
+    gmail.query = label:action OR label:readme
+    # You need a get a client_secret.json. Follow the instructions on: https://developers.google.com/gmail/api/quickstart/python
+    gmail.client_secret_path = ~/.config/bugwarrior/client_secret.json
+    # This file will be created when you authenticate.
+    gmail.credentials_path = ~/.config/bugwarrior/credentials.json
+    # Restrict to a particular e-mail address. This is optional and will cause the pull to fail if the credentials don't match
+    # the specified user. When working with multiple accounts it's a bit a tricky to make sure you always authorise the correct
+    # account.
+    gmail.login_name = name@example.com
--- a/bugwarrior/docs/services/gmail.rst
+++ b/bugwarrior/docs/services/gmail.rst
@@ -1,0 +1,64 @@
+Gmail
+=====
+
+You can create tasks from e-mails in your Gmail account using the ``gmail``
+service name.
+
+Additional Dependencies
+-----------------------
+
+Install packages needed for Gmail support with:
+
+::
+   pip install bugwarrior[gmail]
+
+Client Secret
+-------------
+
+In order to use this service, you need to create a product and download a
+client secret file. Do this by following the instructions on:
+``https://developers.google.com/gmail/api/quickstart/python``. You should save
+the resulting secret in your home directory as ``.gmail_client_secret.json``.
+You can override this location by setting the ``client_secret_path`` option.
+
+Example Service
+---------------
+
+Here's an example of a gmail target:
+
+::
+    [my_gmail]
+    service = gmail
+    gmail.query = label:action OR label:readme
+    gmail.login_name = you@example.com
+
+The specified query can be any gmail search term. By default it will select
+starred threads. One task is created per selected thread, not per e-mail.
+
+You do not need to specify the ``login_name``, but it can be useful to avoid
+accidentally fetching data from the wrong account. (This also allows multiple
+targets with the same login to share the same authentication token.)
+
+Authentication
+--------------
+
+When you first run ``bugwarrior-pull``, a browser will be opened and you'll be
+asked to authorise the application to access your e-mail. Once authorised a
+token will be stored in your bugwarrior data directory.
+
+Provided UDA Fields
+-------------------
+
++---------------------+-----------------------------------+---------------|
+| ``gmailthreadid``   | Thread Id                         | Text (string) |
++---------------------+-----------------------------------+---------------|
+| ``gmailsubject``    | Subject                           | Text (string) |
++---------------------+-----------------------------------+---------------|
+| ``gmailurl``        | URL                               | Text (string) |
++---------------------+-----------------------------------+---------------|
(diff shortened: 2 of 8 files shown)


bugwarrior/docs/configuration.rst
```diff
@@ -188,3 +188,15 @@ Example Configuration
...
+    [my_gmail]
+    service = gmail
+    gmail.query = label:action OR label:readme
+    # You need a get a client_secret.json. Follow the instructions on: https://developers.google.com/gmail/api/quickstart/python
+    gmail.client_secret_path = ~/.config/bugwarrior/client_secret.json
+    # This file will be created when you authenticate.
+    gmail.credentials_path = ~/.config/bugwarrior/credentials.json
+    # Restrict to a particular e-mail address. This is optional and will cause the pull to fail if the credentials don't match
+    # the specified user. When working with multiple accounts it's a bit a tricky to make sure you always authorise the correct
+    # account.
```
ryneeverett: I think most of these comments would be better off in the documentation page at `bugwarrior/docs/services/gmail.rst` which it doesn't look like you've written.

bugwarrior/services/gmail.py
```diff
@@ -0,0 +1,178 @@
...
+    ISSUE_CLASS = GmailIssue
+    CONFIG_PREFIX = 'gmail'
+
+    def __init__(self, *args, **kw):
+        super(GmailService, self).__init__(*args, **kw)
+
+        self.query = self.config.get('query', 'label:Starred')
+        self.login_name = self.config.get('login_name', 'me')
+        self.client_secret_path = self.get_config_path('client_secret_path')
+        self.credentials_path = self.get_config_path('credentials_path')
```
ryneeverett: Would it make sense to have default paths for these? In fact, are settings really needed? I have an especially hard time seeing the use case for `credentials_path` since the user never has any reason to interact with the file.

bugwarrior/docs/configuration.rst
```diff
@@ -188,3 +188,15 @@ Example Configuration
...
     activecollab2.user_id = 15
     activecollab2.projects = 1:first_project, 5:another_project
+
+    [my_gmail]
+    service = gmail
+    gmail.query = label:action OR label:readme
+    # You need a get a client_secret.json. Follow the instructions on: https://developers.google.com/gmail/api/quickstart/python
+    gmail.client_secret_path = ~/.config/bugwarrior/client_secret.json
+    # This file will be created when you authenticate.
+    gmail.credentials_path = ~/.config/bugwarrior/credentials.json
```
ryneeverett: I'm not sure storing machine-written temporary keyfiles in `~/.config` is a good suggestion.

bugwarrior/docs/configuration.rst
cthulahoops: Ah, I missed that. I'll write something.

bugwarrior/services/gmail.py
cthulahoops: I default paths previously, and left them out in the end because I was struggling to come up with reasonable defaults. (I interact with that file because I want to make sure it gets synchronised between various machines.)

Where do you think the credentials should be saved by default? Are you happy with the client secret living in the bugwarrior configuration directory by default?

ryneeverett: I don't think it would be recommended to synchronize any of these keys between machines. I believe the keys are meant to be per-client (device). If I'm wrong in this premise everything below is also wrong.

My suggestion would be to put the long term secret at `$HOME/.gmail_client_secret.json` and the short-term secret in the bugwarrior data path. You could add an attribute `self.path = data_path` in the `bugwarrior.data.BugwarriorData` class and access it by assigning `self.credentials_path = os.path.join(self.config.data.path, 'gmail_credentials.json')`. I'm not at all confident that this is ideal though and am open to better ideas.

If you think the client secret path ought to be configurable I wouldn't oppose a setting for that.

cthulahoops:
The credential files probably should be per machine. Syncing them works for me, but we probably don't need to make any particular effort to support that workflow. The trickier part is the need to store credentials for multiple accounts. I've taken your suggestion of storing them in the bugwarrior data dir, but need to generate a filename containing either the login name or the target name.

The client secret is per product not per machine. I think the right thing is for the user to generate a single secret (for each application) and download/distribute that secret to every machine. If we're going to place this file in the home directory - then it should perhaps be called ``.bugwarrior_gmail_client_secret``. For the moment though, I've gone with your suggestion and left out the bugwarrior, as I'm not sure there's actually any harm in a user reusing a secret for multiple products.

I would like to leave the optional configuration setting in though.

It's confusing when using multiple accounts as there's no indication which browser tab refers to which  service and no way (I can see) to pass through a greeting message. At the moment I'm relying on the fact that the tabs open in the order that the targets are specified in the bugwarrior config. It's not exactly user friendly. I'm considering that using a client secret per account to improve this.

It's not currently possible to specify a particular target when invoking bugwarrior? (This would allow me to run targets one at a time when authenticating.)

bugwarrior/docs/services/gmail.rst
```diff
@@ -0,0 +1,64 @@
...
+You do not need to specify the ``login_name``, but it can be useful to avoid
+accidentally fetching data from the wrong account. (This also allows multiple
+targets with the same login to share the same credentials.)
+
+Authentication
+--------------
+
+When you first run ``bugwarrior-pull``, a browser will be opened and you'll be
+asked to authorise the application to access your e-mail. Once authorised the
+credentials will be stored in your bugwarrior data directory.
```
ryneeverett: I'd recommend changing "the credentials" to "a token". Security-minded users would likely interpret this as meaning that their *password* is stored in the bugwarrior data directory.

bugwarrior/services/gmail.py
```diff
@@ -90,20 +91,18 @@ def __init__(self, *args, **kw):
...
+                'gmail_credentials_%s.json' % (credentials_name,))
         self.gmail_api = self.build_api()
 
-    @classmethod
-    def validate_config(cls, service_config, target):
-        for option in ('client_secret_path', 'credentials_path'):
-            if option not in service_config:
-                die("[%s] has no '%s.%s'" % (target, cls.CONFIG_PREFIX, option))
-
-        IssueService.validate_config(service_config, target)
```
ryneeverett: Did you intend to delete the `validate_config` method?

ryneeverett:
> It's confusing when using multiple accounts as there's no indication which browser tab refers to which service and no way (I can see) to pass through a greeting message. At the moment I'm relying on the fact that the tabs open in the order that the targets are specified in the bugwarrior config. It's not exactly user friendly. I'm considering that using a client secret per account to improve this.

Could we write a message to stdout like "Authorizing <some> service..." and block while the user authorizes each account?

> It's not currently possible to specify a particular target when invoking bugwarrior? (This would allow me to run targets one at a time when authenticating.)

Not that I know of but I would support adding target as a command line argument.

bugwarrior/docs/services/gmail.rst
cthulahoops: Done.

bugwarrior/services/gmail.py
cthulahoops: Yes, there's no longer any validation to do (all parameters are optional). So I'm happy to leave this to ``IssueService``'s implementation.

cthulahoops:
Each of the Services is instantiated in a separate process (via ``multiprocessing.Process``) so we'd need to use  ``multiprocessing.Lock`` to serialise the authentications. That might work quite well though - I could try that.

bugwarrior/services/gmail.py
ryneeverett: Gotcha

ryneeverett:
Looking good. Care to take this for a test drive @pjf?

cthulahoops:
Just squashing the fixups, and reordering so the path attribute gets added before the code that uses it. The code is identical.

pjf:
Thank you so much for this! Apologies I haven't had a chance to properly test this; I was on an international trip and then hit #481 which slowed me down in testing newer features.

Now this is merged I'm ludicrously excited and will be testing soon! Thank you again, @cthulahoops !!

PR #482, opened 2017-06-11
Phab: Add support for tags based on project

OliverUv:
I used the GitHub service's code as reference for implementing this, assuming that it would be the most up to date one. Give me a ping if there's anything I should change.

You can see that I added some `TODO`s. It might look like I've removed functionality where those are located, but I've just removed code that was buggy or doing nothing. Access to the projects that my code is now giving would make it possible to implement those things without much trouble, however.

### DIFF
--- a/bugwarrior/docs/services/phabricator.rst
+++ b/bugwarrior/docs/services/phabricator.rst
@@ -60,6 +60,34 @@ you can find it out by querying Phabricator Conduit
 the methods which return the needed info are ``user.query``, ``project.query``
 and ``repository.query`` respectively.
 
+Import Projects as Tags
++++++++++++++++++++++++
+
+Phabricator tasks may belong to multiple projects, and differentials (PRs)
+belong to a diffusion repository which may belong to multiple projects; to
+use those projects' slugs (short identifying, human readable strings) as
+tags, you can use the ``phabricator.should_set_tags`` option::
+
+    phabricator.should_set_tags = True
+
+Also, if you would like to control how these tags are created, you can
+specify a template used for converting the project slug into a Taskwarrior
+tag.
+
+For example, to prefix all incoming labels with the string 'phab_' (perhaps
+to differentiate them from any existing tags you might have), you could
+add the following configuration option::
+
+    phabricator.tag_template = phab_{{project_slug}}
+
+In addition to the context variable ``{{project_slug}}``, you also have
+access to all fields on the Taskwarrior task if needed.
+
+.. note::
+
+   See :ref:`field_templates` for more details regarding how templates
+   are processed.
+
 
 Provided UDA Fields
 -------------------
--- a/bugwarrior/services/phab.py
+++ b/bugwarrior/services/phab.py
@@ -1,7 +1,10 @@
 from builtins import str
+import re
+
+from jinja2 import Template
 import six
 
-from bugwarrior.config import aslist
+from bugwarrior.config import aslist, asbool
 from bugwarrior.services import IssueService, Issue
 
 # This comes from PyPI
@@ -42,6 +45,7 @@ class PhabricatorIssue(Issue):
             'project': self.extra['project'],
             'priority': self.origin['default_priority'],
             'annotations': self.extra.get('annotations', []),
+            'tags': self.get_tags(),
 
             self.URL: self.record['uri'],
             self.TYPE: self.extra['type'],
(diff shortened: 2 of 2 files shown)


OliverUv:
Adding another pull request which is this one plus some maybe controversial functionality.

OliverUv:
Oh, by the way, I initially tried doing this with TaskWarrior hooks, but whenever I ran `bugwarrior-pull`, it would re-apply the tags I did not want. Added the substitution functionality because I think should be useful to most phab users, and because writing **both** an add- and a modify-hook to ensure tags are managed would be annoying.

bugwarrior/docs/services/phabricator.rst
```diff
@@ -60,6 +60,34 @@ you can find it out by querying Phabricator Conduit
...
 
+Import Projects as Tags
++++++++++++++++++++++++
+
+Phabricator tasks may belong to multiple projects, and differentials (PRs)
+belong to a diffusion repository which may belong to multiple projects; to
+use those projects' slugs (short identifying, human readable strings) as
+tags, you can use the ``phabricator.should_set_tags`` option::
+
+    phabricator.should_set_tags = True
```
ryneeverett: I wonder if `import_projects_as_tags` (in the spirit of github's `import_labels_as_tags`) would be more clear since there are other uses of projects.

bugwarrior/docs/services/phabricator.rst
```diff
@@ -60,6 +60,34 @@ you can find it out by querying Phabricator Conduit
...
+
+Also, if you would like to control how these tags are created, you can
+specify a template used for converting the project slug into a Taskwarrior
+tag.
+
+For example, to prefix all incoming labels with the string 'phab_' (perhaps
+to differentiate them from any existing tags you might have), you could
+add the following configuration option::
+
+    phabricator.tag_template = phab_{{project_slug}}
```
ryneeverett: You might mention that the default value is `{{project_slug}}` somewhere in the docs.

bugwarrior/services/phab.py
```diff
@@ -97,13 +133,13 @@ def issues(self):
...
         for phid, issue in issues:
 
-            project = self.target  # a sensible default
-            try:
-                project = projects.get(issue['projectPHIDs'][0], project)
-            except IndexError:
-                pass
+            # TODO fix this now that we have access to proper projects.
+            # It was bugged before
+            project = self.target
```
ryneeverett: I don't understand this change.

bugwarrior/services/phab.py
```diff
@@ -178,7 +221,89 @@ def issues(self):
...
+
+def fetch_all_related_projects(api, proj_phid_lists):
+    phids = set()
+    for pl in proj_phid_lists:
+        if pl is None:
+            continue
+        for phid in pl:
+            phids.add(phid)
+
+    phids = [p for p in phids] # set to list
```
ryneeverett: Any reason not to use `phids = list(phids)` here?

bugwarrior/docs/services/phabricator.rst
OliverUv: Probably true!

bugwarrior/docs/services/phabricator.rst
OliverUv: That's a good idea. Sorry that the multiple pull requests might be confusing. That is implemented in the most developed of the versions I submitted. See https://github.com/ralphbean/bugwarrior/pull/483/files

bugwarrior/services/phab.py
OliverUv: No! I had just forgotten this was possible, it's been about 2 years since I last wrote any python.

bugwarrior/services/phab.py
OliverUv: The previous author had left in some code that was unused (because the projects list was always empty), anticipating that he could use it after populating the projects list. It should now be possible to implement the feature he planned to implement, though not using his code. So I just deleted his code (which would now break things instead of do nothing) and added a TODO marker.

OliverUv:
@ryneeverett  I'll see if I can get some time on Sunday (China Time) to address your comments. Thanks for taking a look! Also I'd like to know if you think the tag mutation in the other pull request ( https://github.com/ralphbean/bugwarrior/pull/483/files ) is acceptable. It's a feature I rely on now. If so, that branch is where I'll continue my work.

bugwarrior/services/phab.py
ryneeverett: Cool. Could you make the comment more verbose though, to explain what needs fixing?

ryneeverett:
I would suggest cherry-picking 50e334c8aecd5ad765b6ba025e5bb1bad7b1f141 onto this branch, continuing work on this branch, and then rebasing #483 onto this branch. I'm quite comfortable with this feature and implementation but #483 might take a bit longer.

bugwarrior/services/phab.py
devurandom: In b95e24ab0c3876e7b520f4a292594235d0072012 / #566 I populate the `projects` dict as a side-effect from fetching the list of all sub-projects, which also fixes this behaviour.

bugwarrior/services/phab.py
```diff
@@ -169,6 +208,10 @@ def issues(self):
                 except KeyError:
                     pass
 
+                phabricator_projects += projects_from_repo
+
+                # TODO The docs say project PHIDs decide relevancy, but here a
+                # repository PHID is used instead
```
devurandom: Which docs say that and what is "relevancy"?

ralphbean:
@OliverUv, have you had a chance to look into this further?  It would be nice to get it wrapped up and merged (it's been a while since the thread was active).

ryneeverett:
Closing for now since no one has expressed interest in finishing this for some time.

PR #483, opened 2017-06-11
Phab: Add support for mutable tags based on project

OliverUv:
> I used the GitHub service's code as reference for implementing this, assuming that it would be the most up to date one. Give me a ping if there's anything I should change.
> 
> You can see that I added some `TODO`s. It might look like I've removed functionality where those are located, but I've just removed code that was buggy or doing nothing. Access to the projects that my code is now giving would make it possible to implement those things without much trouble, however.

Adding this as a separate pull request because it includes a relatively hacky substitution method (see docs below). I would prefer if this one is merged rather than #482, but I would understand if you didn't want to.

### DIFF
--- a/bugwarrior/services/phab.py
+++ b/bugwarrior/services/phab.py
@@ -74,13 +74,30 @@ class PhabricatorIssue(Issue):
 
         template = Template(self.origin['tag_template'])
         for p in projects:
+            slug = p['fields']['slug']
+            slug = self.apply_tag_substitutions(slug)
             context = self.record.copy()
             context.update({
-                'project_slug': self._normalize_to_tag(p['fields']['slug'])
+                'project_slug': self._normalize_to_tag(slug),
             })
             rendered = template.render(context)
             tags.append(rendered)
         return tags
+
+    def apply_tag_substitutions(self, tag):
+        """Applies a user's regexes and replacement strings to a tag
+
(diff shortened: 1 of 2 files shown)


OliverUv:
Oh, by the way, I initially tried doing this with TaskWarrior hooks, but whenever I ran bugwarrior-pull, it would re-apply the tags I did not want. Added the substitution functionality because I think should be useful to most phab users, and because writing **both** an add- and a modify-hook to ensure tags are managed would be annoying.

bugwarrior/services/phab.py
```diff
@@ -57,6 +61,46 @@ def get_default_description(self):
...
+        exprs = self.origin['tag_substitutions']
+        if len(exprs) == 0:
+            return tag
+        exprs = [e.split('#', 1) for e in exprs]
+
+        def reducer(acc, upd):
+            [pattern, sub] = upd
+            return re.sub(pattern, sub, acc)
+
+        return reduce(reducer, exprs, tag)
```
ryneeverett: This will need to be changed for python3. You could `try` to `import reduce from functools` or you could switch to a for-loop [as suggested by the python maintainers](https://docs.python.org/3.0/whatsnew/3.0.html#builtins). Your call.

ryneeverett:
I'm unsure about this one. Couldn't we change bugwarrrior's behavior to not clobber the hook modifications?

ryneeverett:
Closing for now since no one has expressed interest in finishing this for some time.

Issue #488, opened 2017-06-20
`bugwarrior-pull --dry-run` fails with not-yet-authed gmail integration


pjf:
I'm just getting started with gmail integration. If I run `bugwarrior-pull`, it bounces me over to a Google Accounts sign-in after reading my credentials file. However if I run `bugwarrior-pull --dry-run` I get:

```
INFO:bugwarrior.services:Working on [gmail]
INFO:bugwarrior.services.gmail:Starting authentication for gmail
INFO:bugwarrior.services.gmail:No valid login. Starting OAUTH flow.
usage: bugwarrior-pull [--auth_host_name AUTH_HOST_NAME]
                       [--noauth_local_webserver]
                       [--auth_host_port [AUTH_HOST_PORT [AUTH_HOST_PORT ...]]]
                       [--logging_level {DEBUG,INFO,WARNING,ERROR,CRITICAL}]
bugwarrior-pull: error: unrecognized arguments: --dry-run
CRITICAL:bugwarrior.services:2
INFO:bugwarrior.services:Done with [gmail] in 0.016221s
INFO:bugwarrior.services:Terminating workers
CRITICAL:bugwarrior.command:Aborted (critical error in target 'gmail') 
```

<s>I don't know if dry-runs will work after I've finished the oauth dance (testing that next), but in an unauthenticated state the error isn't what one would expect. :)</s>

`--dry-run` works fine if authentication has been completed. :)

Relates to #477, so cc'ing in the fabulous @cthulahoops. Thank you *so much* for implementing this; I feel I may be finally clearing my inbox sometime soon!

pjf:
I'm guessing desired behaviours with `--dry-run` could be either:

- Tell the user it's not possible to use `--dry-run` if we're not authed, or
- Do the auth (even saving it), but not updating/inserting any tasks.

I'd feel the second option would make the most sense, because it allows the user to play with different options before committing in any way.

More exciting, *it's working*. I now have my pinned gmail tasks in TaskWarrior! :D :D :D

cthulahoops:
Yeah, I think the second would be better too. I don't think it should be too difficult. I hadn't tried dry-run. I'll have a look and see what's involved.

Very glad this is useful to you. :)

pjf:
> Very glad this is useful to you. :)

@cthulahoops : It's *amazing*. Seriously, thank you *so* much.

ryneeverett:
[...]

PR #491, opened 2017-06-21
Gmail oauth flow shouldn't attempt to parse the command line.

cthulahoops:
If any command line arguments are passed to bugwarrior, gmail will try
to parse them and fail. It's not possible pass arguments to Oauth this
way anyway (because bugwarrior would reject the unrecognised command
line arguments), so we shouldn't try to do it. (I see no reason to pass
flags to OAuth here, if needed we'd have to find another way.)

Should resolve issue #488.


pjf:
Tested!

- This resolves #488 for me, and now pushes me through the oauth-dance if I'm not authenticated, which is great!
- After running, this creates an appropriate credentials file inside my `~/.task` directory, but of course does not modify any issues. This matches our suggested behaviour on #488.

**tl;dr**: This works for me! Thank you so much!

PR #493, opened 2017-06-22
services/__init__: Refactor handling of defaults

pjf:
Hi there fabulous maintainers! You've probably seen me bouncing around issues for a while, so I'm hoping to give something back to the main codebase. To help familiarise myself with the code I'm working on making very small and easy changes where I can spot them.

I'm still a relative python newbie, and this is my first pull-request to bugwarrior, so feedback on style and code is very much appreciated. If I've done things in a particularly odd way it's probably because I didn't know there was a better alternative. :)

I noticed that `services/__init__` had a lot of repeated code for handling defaults and configuration. In this change I've refactored that int a separate method, which reduces our line-count and I hope also reduces our complexity.

If these sorts of code cleaning changes are appreciated, then let me know, and I can send through more.

Apologies this change doesn't come with a test suite. I have tested it locally on my machine, but I was making my changes in my brief lunch-break. I'm happy to add tests on request.

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -16,7 +16,7 @@ import six
 
 from taskw.task import Task
 
-from bugwarrior.config import asbool, die, get_service_password, ServiceConfig
+from bugwarrior.config import asbool, asint, die, get_service_password, ServiceConfig
 from bugwarrior.db import MARKUP, URLShortener
 
 import logging
@@ -53,13 +53,6 @@ class IssueService(object):
         self.main_section = main_section
         self.main_config = main_config
         self.target = target
-
-        # Our configuration allows integers to also be 'None'. For example, for
-        # unlimited annotation length. Our private 'asint' function respects this.
-        def asint(x):
-            if x == '':
-                return None
-            return int(x)
 
         self.desc_len = self._get_config_or_default('description_length', 35, asint);
         self.anno_len = self._get_config_or_default('annotation_length', 45, asint);
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -60,7 +60,7 @@ class IssueService(object):
         self.annotation_links = self._get_config_or_default('annotation_links', not self.inline_links, asbool)
         self.annotation_comments = self._get_config_or_default('annotation_comments', True, asbool)
         self.shorten = self._get_config_or_default('shorten', False, asbool)
-        self.default_priority = self._get_config_or_default('default_priority','M', lambda x: x)
+        self.default_priority = self._get_config_or_default('default_priority','M')
 
         self.add_tags = []
         if 'add_tags' in self.config:
@@ -72,7 +72,7 @@ class IssueService(object):
         log.info("Working on [%s]", self.target)
 
 
-    def _get_config_or_default(self, key, default, as_type):
+    def _get_config_or_default(self, key, default, as_type=lambda x: x):
         """Return a main config value, or default if it does not exist."""
 
         if self.main_config.has_option(self.main_section, key):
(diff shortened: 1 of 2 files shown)


pjf:
Oops! Failure from travis! I've now figured out how to use `tox`, so stand-by for improvements and adjustments. :)

pjf:
Updates:

- Empty integer values are passed through as `None`, preserving the `getint` behaviour of ConfigParser.
- Now genuinely passes the test suite. :)

bugwarrior/services/__init__.py
```diff
@@ -51,38 +51,23 @@ class IssueService(object):
...
-
-        self.shorten = False
-        if main_config.has_option(self.main_section, 'shorten'):
-            self.shorten = asbool(main_config.get(self.main_section, 'shorten'))
+        # Our configuration allows integers to also be 'None'. For example, for
+        # unlimited annotation length. Our private 'asint' function respects this.
+        def asint(x):
+            if x == '':
+                return None
+            return int(x)
```
ryneeverett: I would move this function to config.py and import it like `asbool`.

bugwarrior/services/__init__.py
```diff
@@ -51,38 +51,23 @@ class IssueService(object):
...
+
+        self.desc_len = self._get_config_or_default('description_length', 35, asint);
+        self.anno_len = self._get_config_or_default('annotation_length', 45, asint);
+        self.inline_links = self._get_config_or_default('inline_links', True, asbool);
+        self.annotation_links = self._get_config_or_default('annotation_links', not self.inline_links, asbool)
+        self.annotation_comments = self._get_config_or_default('annotation_comments', True, asbool)
+        self.shorten = self._get_config_or_default('shorten', False, asbool)
+        self.default_priority = self._get_config_or_default('default_priority','M', lambda x: x)
```
ryneeverett: I would make `as_type` an optional parameter to avoid passing a do-nothing `lambda` (`as_type=lambda x: x`). From my perspective do-nothing lambda's are less confusing in function signatures than in function calls and if more configuration options are added that do not require type coercion the additional do-nothing lambda's would be repetitious.

ryneeverett:
> If these sorts of code cleaning changes are appreciated, then let me know, and I can send through more.

They are much appreciated, please send more.

bugwarrior/services/__init__.py
pjf: I agree! Thank you!

bugwarrior/services/__init__.py
pjf: Oh! You're right, that is *much* nicer!

Expect an updated commit in the next 24 hours. :)

pjf:
@ryneeverett : Thank you for the review! Will be working on updates in the airport later today. :)

Is `_get_config_or_default` too wordy for a method name? And am I correct that starting a method with an underscore is the correct way to signal it's for internal use?

ryneeverett:
> Is _get_config_or_default too wordy for a method name?

Nope, I like it.

> And am I correct that starting a method with an underscore is the correct way to signal it's for internal use?

Yes.

pjf:
> Will be working on updates in the airport later today. :)

Or it turns out I can sneak coding earlier into my day. Updates are now on the branch. Thank you so much for all the feedback; it's super appreciated!

bugwarrior/services/__init__.py
```diff
@@ -91,12 +69,17 @@ def __init__(self, main_config, main_section, target):
...
                     self.add_tags.append(option)
 
-        self.default_priority = 'M'
-        if 'default_priority' in self.config:
-            self.default_priority = self.config.get('default_priority')
-
         log.info("Working on [%s]", self.target)
 
+
+    def _get_config_or_default(self, key, default, as_type = lambda x: x):
```
ryneeverett: FYI, python convention is to have no spaces around the equals sign in keyword arguments. See the fourth bullet at https://www.python.org/dev/peps/pep-0008/#other-recommendations.

pjf: @ryneeverett : Fixed! Thank you!

ryneeverett:
Merged. Thanks for putting up with the nitpicking!

PR #494, opened 2017-06-22
Common configuration options: update documentation + complete validation

fakenmc:
Update documentation for common service configuration options, namely the new requirement of prefixing these options with the service name.

Also add validation for `default_priority` and `add_tags` options (previously only done for `only_if_assigned` and `also_unassigned`).


ryneeverett:
Thanks!

PR #495, opened 2017-06-24
Reduce redundant code in services/gitlab.py

pjf:
After @ryneeverett said that code reduction is appreciated in #493, here's another refactor.

Both issues and pull-requests used the same processing code in `gitlab.py`, so this factors out a single function to do both. I went with the function returning a list, which is then used to `yield` by the calling code, because:

- It means we're not yielding from inside a function, where it can be missed.
- I'm still a Python n00b, and couldn't figure out to yield from inside a function. :)

This passes our test suite, but I'm not a gitlab user, so I have not tested this on live data. I stumbled upon the duplicated code with the help of [code climate](https://codeclimate.com/github/ralphbean/bugwarrior/issues), which I'm happy to set up more fully for the project if desired.

I'm still learning my way around bugwarrior and python, so feedback *very* much appreciated. :)

### DIFF
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -397,6 +397,27 @@ class GitlabService(IssueService, ServiceClient):
             return project is None or project['id'] in ids
         return include_todo
 
+    def _get_issue_objs(self, issues, issue_type, repo_map):
+        type_plural = issue_type + 's'
+        result = []
+
+        for rid, issue in issues:
+            repo = repo_map[rid]
+            issue['repo'] = repo['path']
+
+            issue_obj = self.get_issue_for_record(issue)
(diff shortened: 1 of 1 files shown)


ryneeverett:
> - It means we're not yielding from inside a function, where it can be missed.

I'm not sure what you mean by "missed".

> - I'm still a Python n00b, and couldn't figure out to yield from inside a function. :)

You should be able to `yield issue_obj` from within the for-loop of `_get_issue_objs`.

```py
>>> def inner_generator():
...     for word in ["foo", "bar", "baz", "qux"]:
...         yield word


>>> def outer_generator():
...     for word in inner_generator():
...         yield word


>>> def print_words():
...     for word in outer_generator():
...         print(word)


>>> print_words()
foo
bar
baz
qux
```

ralphbean:
Merged in #650.

Issue #490, opened 2017-06-21
Description can't be a static field?


pjf:
Hello again fabulous bugwarrior devs!

I'm hitting a stumbling block with my workflow, which might just be me misunderstanding the documentation, but it looks like I can't make description into a static field. This is something I'd *really* love, as it means that while bugwarrior can set my task names to begin with, I can modify them to be more descriptive as needed. This is especially useful with the new gmail service, as subject lines don't always reflect the thing I need to be doing.

I have the following config fragment:

```
[general]
static_fields = description
```

But when I run this, bugwarrior-pull gives me:

```
Traceback (most recent call last):
  File "/usr/local/bin/bugwarrior-pull", line 11, in <module>
    load_entry_point('bugwarrior==1.5.1', 'console_scripts', 'bugwarrior-pull')()
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/bugwarrior-1.5.1-py2.7.egg/bugwarrior/command.py", line 73, in pull
    synchronize(issue_generator, config, main_section, dry_run)
  File "/usr/local/lib/python2.7/dist-packages/bugwarrior-1.5.1-py2.7.egg/bugwarrior/db.py", line 351, in synchronize
    del issue_dict[field]
KeyError: u' description'
```

I'm using bugwarrior ddc60f6.

Is there a way to have description be a static field, so bugwarrior can set it on task creation, but not modify it after the task exists? (Right now bugwarrior helpfully reverts my hand-crafted task descriptions back to their defaults whenever it runs.)

Many, many thanks again for an amazing tool!

~ pjf

ryneeverett:
The `issue_dict` ought to have a `description` key at that point in the code. I suspect this is a bug in the gmail service which is somehow failing to add a `description` field.

cthulahoops:
I'm struggling to reproduce this. It seems to just work for me.

[...]

PR #496, opened 2017-06-24
Use aslist everywhere when parsing list config options.

cthulahoops:
Code to split config options on commas was duplicated in a number of
places, not all of which remembered to strip leading and trailing
whitespace.
Resolves issue #490.


pjf:
Did a local rebase and tested with these changes. This absolutely resolves #490 for me!

Thank you so much, @cthulahoops!

ryneeverett:
Thanks all!

PR #497, opened 2017-07-10
Unpin futures to avoid dependency conflicts

### LINKED ISSUE #427: bugwarrior conflicts with version of future in fedora
I removed "future!=0.16.0" from setup.py and it seems to be working fine for me after that. This is on F24.

ssbarnea:
Fix #427

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(name='bugwarrior',
           "dogpile.cache>=0.5.3",
           "lockfile>=0.9.1",
           "click",
-          "future!=0.16.0",
+          "future",
       ],
       extras_require=dict(
           keyring=["keyring", "dbus-python"],


ryneeverett:
The build failure you're encountering is the reason the version was pinned in the first place. I wouldn't say pinning is the ideal solution but it seems like a better trade-off until somebody does a deeper investigation into the underlying problem.

ryneeverett:
Since the errors are only encountered in python2 though, an easy patch would limit the scope of #427 by only pinning the version in python2.

ssbarnea:
Lets be more specific: my impression is that the build error is specific to an incompatibility introduced on Fedora distribution of that package, meaning that if we pin something it should look like "xxx!=1.2.3; platform!=xxx". Full details at https://www.python.org/dev/peps/pep-0508/#examples

ryneeverett:
@ssbarnea Did ubuntu make the same packaging mistake? That's the base distribution travis uses and this patch breaks python2 jobs.

ralphbean:
Triggering a re-run of the travis build to see if this is resolved in that context.

ralphbean:
And.... looks like we're good now.  :)

PR #498, opened 2017-07-11
WIP Issue 487 gtask

### LINKED ISSUE #487: Google Tasks Support
The following is copied from [a comment](https://github.com/ralphbean/bugwarrior/issues/442#issuecomment-282004477) by @emmajane:

> Hello! I am reading this and I use Gmail for task management! I specifically use their keyboard shortcut to send an email to their "tasks".
>
> My workflow currently looks like this:
>
> - Triage Gmail inbox using [ and ] shortcuts (send to archive; move to next message in the inbox).
> - If there is an email that requires follow-up, I first send the message to Gmail tasks using shift+t.
>
> This action adds the task to a pop-up window which appears in the bottom right of my screen (see attached).
>
> attached
>
> This also appears as a stand-alone screen at: https://mail.google.com/tasks/canvas?pli=1
>
> There are Gmail help pages at: https://support.google.com/mail/answer/106237?hl=en I believe this is the relevant API documentation: https://developers.google.com/google-apps/tasks/
>
> I'm happy to provide more details if they would be useful.

cthulahoops:
This has been hanging around for a while and I haven't had a chance to finish it off, so here it is in case anyone wants to try it out, and in case there are any comments on the refactor of the split to a GoogleService base class.

I think it only needs some documentation.

Should nearly resolve #487 .

### DIFF
--- a/bugwarrior/docs/configuration.rst
+++ b/bugwarrior/docs/configuration.rst
@@ -193,3 +193,8 @@ Example Configuration
     service = gmail
     gmail.query = label:action OR label:readme
     gmail.login_name = you@example.com
+
+    [my_gtask]
+    service = gtask
+    gtask.lists = task list 1, task lisk 2
+    gtask.login_name = you@example.com
--- a/bugwarrior/services/gmail.py
+++ b/bugwarrior/services/gmail.py
@@ -1,15 +1,6 @@
-import httplib2
-import os
 import email
-import re
-import multiprocessing
-
-import googleapiclient.discovery
-import oauth2client.client
-import oauth2client.tools
-import oauth2client.file
-
-from bugwarrior.services import IssueService, Issue
+from .google import GoogleService
+from bugwarrior.services import Issue
 
 import logging
 log = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


cthulahoops:
(My real target is google calendar support.)

ryneeverett:
Nice. @emmajane might be interested in trying it out.

I wonder if it wouldn't be a simpler user experience to roll the two into one "google" service with `google.gmail_query` and `google.task_lists` options, especially if more google sub-services are on the horizon. The separation of concerns could be maintained in separate classes though, with the `GoogleService` class instantiating sub-service classes as properties and calling their methods depending on the configuration options. I'm just throwing ideas out there though. What do you think?

ralphbean:
This is nice @cthulahoops - have you had a chance to work on it any more?

Looks like it will need to be rebased on `develop` now that #667 is merged.

cthulahoops:
Hi. No, I switched away from using taskwarrior, so I haven't been giving this any attention. The comments above seem reasonable, but I think it's unlikely I'll get back to this, sorry.

ralphbean:
Ok.  I'm going to close this then.  Cheers!

(If anyone wants to pick this up, grab the code from @cthulahoops's branch and open a new PR, please.)

PR #499, opened 2017-07-18
WIP: Also include tasks for taiga

fmauch:
By default tasks are not included into the taiga pull.
I hacked access to assigned tasks into the taiga service, but probably this should be configurable at some point?

Before I dive deep into making this really clean I wanted to know whether this has a chance of getting merged into the main project.

Cheers
Felix

### DIFF
--- a/bugwarrior/services/taiga.py
+++ b/bugwarrior/services/taiga.py
@@ -123,7 +123,7 @@ class TaigaService(IssueService, ServiceClient):
             yield issue
 
         if self.include_tasks:
-            for issue in  self._issues(userid, 'task', 'tasks', 'task'):
+            for issue in self._issues(userid, 'task', 'tasks', 'task'):
                 yield issue
 
     @cache.cache_on_arguments()
--- a/bugwarrior/services/taiga.py
+++ b/bugwarrior/services/taiga.py
@@ -7,6 +7,9 @@ from bugwarrior.db import CACHE_REGION as cache
 from bugwarrior.config import die
 from bugwarrior.services import IssueService, Issue, ServiceClient
 
+import logging
+log = logging.getLogger(__name__)
+
 
 class TaigaIssue(Issue):
     SUMMARY = 'taigasummary'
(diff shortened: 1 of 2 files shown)


ryneeverett:
I would merge this as an optional feature in the absence of any compelling objections from others. I think we should add a configuration option like `taiga.include_tasks` defaulting to `False`.

fmauch:
I added a config flag, defaulting to False and also updated the documentation accordingly.

bugwarrior/services/taiga.py
```diff
@@ -131,3 +150,15 @@ def annotations(self, story, project):
...
+        url = '%s/api/v1/history/task/%i' % (self.url, task['id'])
+        response = self.session.get(url)
+        history = response.json()
+        return self.build_annotations(
+            ((
+                item['user']['username'],
+                item['comment'],
+            ) for item in history if item['comment']),
+            self.build_url_tasks(task, project)
+        )
```
ryneeverett: Couldn't we keep the single `annotations` and `build_url` methods but add a `task_type` argument? E.g.:

```py
def build_url(self, story, project, task_type):
    return '%s/project/%s/%s/%i' % (self.url, project['slug'], task_type, task['ref'])

def annotations(self, task, project, task_type):
    url = '%s/api/v1/history/%s/%i' % (self.url, task_type, task['id'])
    ...
        self.build_url(task, project, task_type)
    ...
```

bugwarrior/services/taiga.py
```diff
@@ -107,11 +108,26 @@ def issues(self):
...
+
+            for task in tasks:
+                project = self.get_project(task['project'])
+                extra = {
+                    'project': project['slug'],
+                    'annotations': self.annotations_tasks(task, project),
+                    'url': self.build_url_tasks(task, project),
+                }
+                yield self.get_issue_for_record(task, extra)
+
```
ryneeverett: You might abstract this into a method, e.g.:

```py
def issues(self):
    ...
    yield self._issues('userstory', 'userstories')

    if self.include_tasks:
        yield self._issues('task', 'tasks')

def _issues(task_type, task_type_plural):
    response = self.session.get(
        self.url + '/api/v1/' + task_type_plural,
        params={'assigned_to': userid, 'status__is_closed': "false"})
    tasks = response.json()

    for task in tasks:
        project = self.get_project(task['project'])
        extra = {
            'project': project['slug'],
            'annotations': self.annotations(task, project, task_type),
            'url': self.build_url(task, project, task_type),
        }
        yield self.get_issue_for_record(task, extra)

PR #501, opened 2017-07-23
Globally exclude __pycache__ and py[co] from sdist

koobs:
The current (1.5.1) source distribution contains compiled python files/artifacts in the tests/ directory.

This change excludes them globally from source distributions.


ryneeverett:
Thanks.

PR #514, opened 2017-09-09
doc: rework section Common Service Configuration Options

rohieb:
The last paragraph probably gets overread a lot (like in my case), so make the service-prefixing more clear and put the paragraph to the top so it doesn't get lost under the list.

Mention default values and make the documentation reflect the implementation better, and try to add more clarity.


rohieb:
Anything to add? :)

ryneeverett:
Nope, just leaving a week for input from other contributors. Thanks!

rohieb:
Ah, makes sense. Thank you!

PR #515, opened 2017-09-10
gitlab: filter repositories by regex

rohieb:
This introduces the service options `gitlab.include_regex` and `gitlab.exclude_regex`, which can be used in addition to the existing lists in `gitlab.include_repos` and `gitlab.exclude_repos`.

For my use case: I would like to import all issues of the repos in my own namespace; I feel responsible of those even when I'm not assigned. Also I want the issues that are assigned to me in organization repos (I don't care about the rest, someone else can grab them :P)

So basically I want two different services, one with `gitlab.only_if_assigned` seti to my username, matching only the organization repos, and one without that option matching my own repos. However, the GitLab service (or even the API? I'm not sure) lists all accessible repos, even in organizations. So I would have to maintain an exclude list of organization repos in the one service definition, and a list of my own repos in the other service definition. Without regexes, listing them all and keeping the lists in sync with reality seems like a tedious nightmare.

### DIFF
--- a/bugwarrior/docs/services/gitlab.rst
+++ b/bugwarrior/docs/services/gitlab.rst
@@ -60,6 +60,33 @@ In this example, ``noisy/repository`` is the repository you would
        gitlab.login = foo
        gitlab.include_repos = foo/bar
 
+Filtering Repositories with Regular Expressions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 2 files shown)


rohieb:
Hmm, the additional debug output seems to throw off the tests. I have removed it again.

ryneeverett:
Any thoughts @mathstuf ?

mathstuf:
LGTM (mobile app, so no review available)

PR #522, opened 2017-11-05
Modify drop static field logic to prevent KeyError

Fongshway:
A KeyError is thrown when I have a field in the static_fields config that is not implemented in a specific service. Rather than have the KeyError get thrown, I figured it makes sense to add a check to confirm whether or not the field even exists in the issue.


irl:
+1

I have this exact patch already prepared, but hadn't got to making a PR yet.

ryneeverett:
Thanks for the patch and review!

Issue #484, opened 2017-06-15
Include organization in project name


paralin:
It would be useful to have project names like so: `ralphbean.bugwarrior`, that way for an organization like `golang/protobuf` we would have `golang.Protobuf`, for Go we would have `golang.Go`, etc. Then you can classify every Go project with just the selector `project:golang` for example.

ryneeverett:
I was unfamiliar with taskwarrior's "project hierarchy" support until I just now found it in the [examples](https://taskwarrior.org/docs/examples.html).

I'd like to see support for this feature across all our services. That doesn't preclude merging a pull request that only adds github support but it may require a more abstract implementation than your pull request provides (I haven't reviewed it yet).

ryneeverett:
Reopening since I'd like to see this for all services, not just github.

ryneeverett:
I think there's been sufficient implementation to establish precedent and there's no need to keep this open. Further implementation across services is, of course, welcome.

PR #523, opened 2017-11-07
Prefix project name with git service namespace

tjaartvdwalt:
I implemented #484  for Gitlab and Bitbucket.


ryneeverett:
Thanks!

PR #524, opened 2017-11-09
TypeError: not enough arguments for format string

aliceinwire:
python is complaining about not enough arguments.

I couldn't test because of retro-incompatibility with my Redmine (3.2.0.stable.15005)
I got NameError: name 'DESCRIPTION' is not defined


ryneeverett:
Thanks!

PR #531, opened 2017-12-04
Preserve PERL5LIB environment when calling out to taskw

pjf:
When we call out to TaskWarrior, we provide a very restricted
environment. However TaskWarrior may have hooks installed which rely
upon certain environment variables to be set.

In particular, any hooks written in Perl, or loaded using the
Taskwarrior::Kusarigama hook framework, will depend upon the `PERL5LIB`
environment variable being set.

This change passes through the `PERL5LIB` environment variable if it
exists. Since neither BugWarrior nor the TaskWarrior core is written in
Perl, this is essentially a no-op on systems that aren't using Perl
hooks.

There's also potentially scope to discuss what environment cleaning
should happen when running TaskWarrior (possibly with `.bugwarriorrc`
config settings), but this change fixes the conflict that currently
exists between BugWarrior and Kusarigama (which I both use heavily).

### DIFF
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -245,6 +245,11 @@ def get_data_path(config, main_section):
     if taskdata:
         env['TASKDATA'] = taskdata
 
+    # Let's not break TaskWarrior's Kusarigama
+    perl5lib = os.getenv('PERL5LIB')
+    if perl5lib:
+        env['PERL5LIB'] = perl5lib
+
     tw_show = subprocess.Popen(
         ('task', '_show'), stdout=subprocess.PIPE, env=env)
     data_location = subprocess.check_output(


ryneeverett:
Is there any harm in `PERL5LIB` being set to an empty string? It looks like there is an issue with `TASKDATA` being empty which is why I conditionally added the variable to the environment. If there's no issue might as well add it directly to the `env` dict. (`TASKDATA: os.getenv('TASKDATA'),`)

However, maybe there's a better approach...

> There's also potentially scope to discuss what environment cleaning
should happen when running TaskWarrior

I agree. It doesn't look like the environment cleaning was at all intentional in this case, it's just the default `Popen` behavior when you use the `env` argument. 

> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of inheriting the current process’ environment, which is the default behavior.

See https://github.com/ralphbean/bugwarrior/commit/9199e9aaee6b3e286f9359a295b6150defe02de4#diff-7df37b661601d0ca7a7a621f8c1c8d1cR215. I simply needed to override the `TASKRC` variable and didn't even think about other variables that might come into play.

Maybe a better approach would be to grab the whole `os.environ`, set the `TASKRC` variable, and pass the whole thing through?

On the other hand, why is `task _show` triggering user hooks?

pjf:
>  Is there any harm in PERL5LIB being set to an empty string?

It means that any user-installed modules arne't going to be found, and it's common (and arguably best) practice to install modules locally rather than overwriting system libraries.

> Maybe a better approach would be to grab the whole `os.environ`, set the `TASKRC` variable, and pass the whole thing through?

That sounds absolutely perfect! And is a much, much better solution than what I've proposed.

> On the other hand, why is `task _show` triggering user hooks?

I would expect on-launch and on-exit hooks to trigger, and I have both of these.

pjf:
Closing this as we've a better way to do this.

Opening a new issue I can reference in commits so folks can see the conversation.

PR in the next hour if my coding skills don't fail me.

PR #534, opened 2017-12-13
Preserve existing environment variables when calling task

### LINKED ISSUE #533: Preserve existing environments when calling `task`
As described in #531, we currently run taskwarrior with a *very* restricted environment. It's confirmed in those discussions that this is a bug, and we only really want to modify a couple of task-related variables and pass the rest through as-is.

This ticket exists as I've got a new PR coming up in the next hour, and I like to have tickets that PRs can close. :)

pjf:
This change copies our existing environment and then makes changes to it
before calling `task`, rather than creating an empty environment and
populating it.

This is much friendlier to any hooks the user may have set.

Replaces #531.
Closes #533.

### DIFF
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -240,11 +240,6 @@ def get_data_path(config, main_section):
     env = dict(os.environ)
     env['TASKRC'] = taskrc
 
-    # Prune the TASKDATA variable if it's completely empty, since otherwise
-    # taskwarrior tries to open the empty-string directory, which fails.
-    if not env.get('TASKDATA',''):
-        del env['TASKDATA']
-
     tw_show = subprocess.Popen(
         ('task', '_show'), stdout=subprocess.PIPE, env=env)
     data_location = subprocess.check_output(
--- a/bugwarrior/config.py
+++ b/bugwarrior/config.py
@@ -236,14 +236,14 @@ def get_data_path(config, main_section):
     # the `_` subcommands properly (`rc:` can't be used for them).
     line_prefix = 'data.location='
 
-    env={
-        'PATH': os.getenv('PATH'),
-        'TASKRC': taskrc,
-    }
-    # If TASKDATA is set but empty, taskwarrior's data.location is empty.
-    taskdata = os.getenv('TASKDATA')
-    if taskdata:
-        env['TASKDATA'] = taskdata
+    # Take a copy of the environment and add our taskrc to it.
+    env = os.environ
+    env['TASKRC'] = taskrc
+
+    # Prune the TASKDATA variable if it's completely empty, since otherwise
+    # taskwarrior tries to open the empty-string directory, which fails.
+    if not env.get('TASKDATA',''):
+        del env['TASKDATA']
 
     tw_show = subprocess.Popen(
         ('task', '_show'), stdout=subprocess.PIPE, env=env)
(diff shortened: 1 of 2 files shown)


bugwarrior/config.py
```diff
@@ -236,14 +236,14 @@ def get_data_path(config, main_section):
...
-    env={
-        'PATH': os.getenv('PATH'),
-        'TASKRC': taskrc,
-    }
-    # If TASKDATA is set but empty, taskwarrior's data.location is empty.
-    taskdata = os.getenv('TASKDATA')
-    if taskdata:
-        env['TASKDATA'] = taskdata
+    # Take a copy of the environment and add our taskrc to it.
+    env = os.environ
```
coddingtonbear: Just an FYI, but on some platforms, changing the values of `os.environ` will actually change the calling environment's environment variables.  You probably want to create a copy of this by doing something like `env = dict(os.environ)`

pjf: Oh, oops! I thought this *did* make a copy, but I'm still new to Python and learning when things are assigning-by-value and assigning-by-reference. Thank you so much for spotting this!

Updated changes landing soon. :)

bugwarrior/config.py
```diff
@@ -236,14 +236,14 @@ def get_data_path(config, main_section):
...
-    if taskdata:
-        env['TASKDATA'] = taskdata
+    # Take a copy of the environment and add our taskrc to it.
+    env = dict(os.environ)
+    env['TASKRC'] = taskrc
+
+    # Prune the TASKDATA variable if it's completely empty, since otherwise
+    # taskwarrior tries to open the empty-string directory, which fails.
+    if not env.get('TASKDATA',''):
+        del env['TASKDATA']
```
ryneeverett: I would delete this `TASKDATA` part altogether. This patch changes its nature from working around bugwarrior-caused misconfiguration to working around user-caused misconfiguration.

The reason we needed to have this conditional before is that `os.getenv('TASKDATA')` returns an empty string when the `TASKDATA` environment variable *is not set*.

After your revision, its purpose is to silently work around a mis-configured system in which the user has assigned `TASKDATA` an empty string. I don't think this is going to be a very frequent scenario. If it were, a more appropriate response would be to raise a warning or an exception. Also bugwarrior would probably break down at a lot of other points if `TASKDATA` were set to empty since this is far from our only invocation of `task`.

pjf:
My apologies for my extraordinarily slow response on this PR. Moving house ends up taking more time than I thought!

I've adjusted the changes to no longer touch an empty TASKDATA env variable, and have rebased to the tip of develop.

We still retain the rest of the user's environment, so anything needing that (like taskwarrior hooks) can use it.

pjf:
Done! Although I used a slightly different snippet of code since Python does not look fondly upon looking up a key that's not there. :)

There seems to be a transient travis failure where it failed to upgrade `pip`. I don't think I've got permissions to restart the job, but the other travis instances passing makes me feel good about the test changes.

tests/test_config.py
```diff
@@ -1,4 +1,4 @@
-# coding: utf-8
```
ryneeverett: What does this do?

pjf: I'm guessing that would be me typing into the wrong window. :fearful:  Removing now.

coddingtonbear: That's how you tell Python 2.x that the file you're in might contain UTF8 bytes.  By default, it parses files with ascii.  https://www.python.org/dev/peps/pep-0263/

ryneeverett:
Thanks for sticking with it!

pjf:
Thank *you* for sticking with it! ♥

PR #539, opened 2018-01-05
Use comma instead of space separated list for [service].add_tags in example configuration

ubuntudroid:
add_tags requires a comma separated list of tags, not a space separated list as currently in the example configuration. This changes the example accordingly.


ryneeverett:
Thanks!

PR #553, opened 2018-04-08
redmine: add UDA to export .project.name

towo:
Since the default filter for some spurious reason does a regexp replace
on the project name, I at least want to export it to use it in my own
`project_template`.

This is only the top-level project, though, as the parent/child
relationship for projects is done outside of the issue object.

### DIFF
--- a/bugwarrior/services/redmine.py
+++ b/bugwarrior/services/redmine.py
@@ -56,6 +56,7 @@ class RedMineIssue(Issue):
     UPDATED_ON = 'redmineupdatedon'
     DUEDATE = 'redmineduedate'
     ASSIGNED_TO = 'redmineassignedto'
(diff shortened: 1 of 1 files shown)


towo:
FWIW, the build fails because of some other peculiarity (tz changing from local to utc) which should be present in the base already.

ralphbean:
@umonkey, @kostajh - can either of you give this one a review.  I think you may be the resident redmine experts.  :)

ralphbean:
Well, no response from @kostajh and @umonkey.

@towo - can you rebase this on `develop` to see if that resolves the utc issue for travis?

ralphbean:
Nevermind.  Rebased in as 10e14eb.  :)  Cheers, and thanks!

PR #554, opened 2018-04-08
Use get_password to evaluate API key

towo:
This allows you to put the API key into @oracle mode so you don't have
to store clear text API keys in your configuration.

This should not have any effect on existing configuration.


ryneeverett:
Makes sense to me. Seems like we got the semantics wrong originally and the method that does keychain lookups should be called *get_secret*.

PR #555, opened 2018-04-16
Store GMail labels in a new gmaillabels UDA

pjf:
While user-defined labels get imported as tags using the gmail service, automatic labels (IMPORTANT, SENT, CATEGORY_*, etc) get pruned from the tags to reduce clutter. However these automatic labels are still useful to have, especially for tools that may wish to act on mailbox state. This PR places all the labels into a UDA.

It's worth noting that while I'm using a simple space-separated list, this isn't necessarily ideal as it looks like labels can contain *anything*. However none of the automatic labels contain spaces (which are the only things we *really* need the UDA for), and something in our pipeline (possibly TaskWarrior itself) munged my attempts to store JSON data in a UDA (resulting in ugly `&dquot;` strings).

I'm currently using this code as part of my code that synchronises pinned/snooze inbox status to taskwarrior. Knowing that my gmail query is `label:snoozed OR label:pinned`, I can use the following queries to translate that information to TaskWarrior.

```
yes | task gmailurl.any: gmaillabels.hasnt:INBOX +PENDING modify -INBOX wait:+7d

yes | task gmaillabels.has:INBOX +WAITING modify +INBOX wait:
```


ryneeverett:
Makes sense to me. Just need to add the labels to the test data to get the build passing.

pjf:
- Code updated to always sort labels.
- Tests updated to expect sorted labels.
- Travis is happy.

pjf:
Just checking in to see if there's anything else needed from me here. :)

Many thanks!

ryneeverett:
Not by me. I just like to leave UDA-changing pull requests open a little longer for community feedback.

pjf:
Thank you for the merge! I'm showing off how I use BugWarrior and Gmail today at CraftConf, so the timing is perfect! ♥

PR #557, opened 2018-04-20
gitlab: make the gitlabnumber UDA a string (fixes #552)

### LINKED ISSUE #552: deletes and recreates todos every run (gitlab.com)
Every time I run bugwarrior-pull, it marks my todos as completed and creates new ones.

Curiously, this only happens with gitlab.com but not with gitlab.gnome.org.

abderrahim:
The gitlabnumber's of todos on gitlab.com have grown beyond the limits of the
numeric type, and that may happen on other instances too.

### DIFF
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -75,7 +75,7 @@ class GitlabIssue(Issue):
             'label': 'Gitlab Type',
         },
         NUMBER: {
-            'type': 'numeric',
+            'type': 'string',
             'label': 'Gitlab Issue/MR #',
         },
         STATE: {
@@ -168,7 +168,7 @@ class GitlabIssue(Issue):
             self.TITLE: title,
             self.DESCRIPTION: description,
             self.MILESTONE: milestone,
-            self.NUMBER: number,
+            self.NUMBER: str(number),
             self.CREATED_AT: created,
             self.UPDATED_AT: updated,
             self.DUEDATE: duedate,


coddingtonbear:
Out of curiosity, which numeric type is in use, and in which system is that limit being reached?

ryneeverett:
Pardon @coddingtonbear, but I don't understand your questions. By my reading of @abderrahim's analysis, `numeric` is the taskwarrior UDA type and the limitation is in the underlying data type taskwarrior uses to store `numeric`. Or are you asking which C numeric type is in use by taskwarrior and on which operating system the limit is being reached?

coddingtonbear:
Yep!  I wasn't totally clear on whether this was attempting to workaround a python numeric type limitation (which shouldn't ever be necessary given Python having support for arbitrarily large numbers), or if this was an attempt at working around a limitation of JSON's numeric type (2**53), or if this was attempting to work around some kind of limitation in Taskwarrior itself.  Even a `uint32_t` can fit up numbers higher than 4 trillion.

chikei:
It likely due to this upstream bug https://github.com/GothenburgBitFactory/taskwarrior/issues/2011.
However when it reaches 9 digits, taskwarrior seems start to lose precision, consider gitlab use global id for todo and recent migration event, this MR might still relevant even taskwarrior fix the bug.

ralphbean:
OK - I'm inclined to include this if it unblocks people who are running over that limit.

ralphbean:
Rebased and included as 0c139f3.

PR #562, opened 2018-05-08
Fix comparing with projects/repositories for Phabricator diffs

urzds:


ryneeverett:
Could you elaborate on what this fixes?

devurandom:
Before this fix, you were comparing PHIDs of type "PROJ" (from the issue) with PHIDs of type "USER" (from the config), which would never match. After this commit, you compare the correct types.

PR #563, opened 2018-05-08
Allow "strict matching" mode where user AND project has to match for Phabricator

ghost:


ryneeverett:
Seems like a good feature but needs a rebase and documentation.

Also -- and this is just a thought and not an objection or change request at this point -- I'm wondering if "strict matching" is the right semantics. I know its conceptually similar to regex "strict matching" but I don't think I've seen the phrase applied to "AND matching".

ralphbean:
@ghost - have you had a chance to come back and look at this?

ralphbean:
Oh.  Nevermind.  @ghost is deleted...

PR #565, opened 2018-05-09
 Add an option to always make the target/section name the project for Phabricator issues/diffs

ghost:
This depends on #561

### DIFF
--- a/bugwarrior/services/phab.py
+++ b/bugwarrior/services/phab.py
@@ -1,7 +1,7 @@
 from builtins import str
 import six
 
-from bugwarrior.config import aslist
+from bugwarrior.config import aslist, asbool
 from bugwarrior.services import IssueService, Issue
 
 # This comes from PyPI
@@ -93,6 +93,9 @@ class PhabricatorService(IssueService):
         self.shown_project_phids = (
             self.config.get("project_phids", None, aslist))
 
+        self.target_is_project = self.config.get(
+            "target_is_project", None, asbool)
+
     def issues(self):
 
         # TODO -- get a list of these from the api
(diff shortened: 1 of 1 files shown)


ryneeverett:
Needs a rebase and documentation

I'd prefer isolating this feature rather than adding more nested if-statements in the loops. That is, just override the `project` variable at the end of the function:
```py
if self.target_is_project:
    project = self.target
```

Also, is this a feature you actually want? Seems like a strange use case to me.

devurandom:
Initially I thought taking the taskwarrior projects attribute from the Phabricator project tags would work, but it turned out that I'd need a way to prioritise them, in case there is more than one tag. So I came up with having multiple sections for the same `phabricator.host` instead. Each section has a different name and `phabricator.project_phids`, so the result is that my taskwarrior tasks get a deterministic project attribute.

ryneeverett:
Would it be better to use ‘myphabtarget.add_tags=myphabtarget’ and prioritize with tags instead?

devurandom:
`myphabtarget.add_tags=myphabtarget` -- what does that do?

When you say "prioritise with tags" you mean [taskwarrior UDA value weights / coefficients](https://taskwarrior.org/docs/priority.html)? In that case: (So far) I did not use projects to prioritise tasks, but to allow me to filter them efficiently (and have this information immediately visible when skimming through the task list) using the `project` column in taskwarrior. Even if tasks in Phabricator have multiple project tags, usually only one of them is related to an actual project (as in area of responsibility), while the others are just tags, like "user story" or "bug" or relations like "team X/Y/Z".

ryneeverett:
Adding a tag with the same name as the target would allow you to filter by that tag the same way you would if it were the project name.

ralphbean:
WDYT, @urzds?   Are you still using this feature in your personal branch?

urzds:
> WDYT, @urzds? Are you still using this feature in your personal branch?

I am still using this. Thanks for reminding me -- I will need to invest some time in the coming weeks and properly re-design my PRs, incorporating the feedback that I received.

ralphbean:
> I am still using this. Thanks for reminding me -- I will need to invest some time in the coming weeks and properly re-design my PRs, incorporating the feedback that I received.

Thanks.  :)  We'll stand by!

ralphbean:
@ghost has been deleted.  :(

Closing this.

PR #567, opened 2018-05-09
Phabricator: Add an additional PHID field, make hashable based on PHID, simplify ID deduction

ghost:


ryneeverett:
What is the benefit of this change?

ralphbean:
@ghost is deleted.  Closing PR.

PR #571, opened 2018-05-15
Fix default_priority option

alyssais:
This has been broken since 02e9bf615000ed11a92f368b85932a5ba2357819, which incorrectly assumed like, all the other options here, it came from the main section rather than the target section. I left a blank line to show that it's different.

I also removed hard coded priorities in the Trello and GitHub services (the only ones I use so can reasonably test). Overriding like the GitHub service currently does makes it _impossible_ to set any other default priority for pull requests, while it's still possible to have them default to high priority by using a template or another target.


ryneeverett:
Makes sense to me. Just need to fix the test data and make the change across all services. We'd never get things fixed if we waited for a user of every service to manually test.

ralphbean:
Rebased and merged.  :)  Thanks @alyssais!

PR #576, opened 2018-05-18
[phabricator] Connect to phabricator without Differential

genofire:
phabricator could be runned without `Differential` or Review.

```
phabricator.APIError: ERR-CONDUIT-CALL: Method 'differential.query' belongs to application 'Differential', which is not installed.
```

---
A solution is to put this part in a try block

### DIFF
--- a/bugwarrior/services/phab.py
+++ b/bugwarrior/services/phab.py
@@ -140,11 +140,13 @@ class PhabricatorService(IssueService):
             }
 
             yield self.get_issue_for_record(issue, extra)
+
         try:
-                diffs = self.api.differential.query(status='status-open')
+            diffs = self.api.differential.query(status='status-open')
         except phabricator.APIError as err:
-                log.warn("Could not read reviews: %s" % err)
-                return
+            log.warn("Could not read reviews: %s" % err)
+            return
+
         diffs = list(diffs)
 
         log.info("Found %i differentials" % len(diffs))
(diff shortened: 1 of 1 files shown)


genofire:
@ryneeverett hi, that is no reason to deny this PR -> a "request for changes" ...
Without a solution it is not possible to use the phabricator service ... so it is a bug, which is needed to fixed on one or a other way.

devurandom:
@genofire The PR was not reject in principle, but in its current form. @ryneeverett proposed very specific changes that need to be made for this PR to be accepted.

genofire:
Sry for the wrong interpretation of deny.
Have changed the logoutput from:
```
WARNING:bugwarrior.services.phab:Could not read reviews
```
to:
```
WARNING:bugwarrior.services.phab:Could not read reviews: ERR-CONDUIT-CALL: Method 'differential.query' belongs to application 'Differential', which is not installed. 
```

Better?

ryneeverett:
Better error message but we need to catch the specific exception that’s being raised.  ‘Exception’ is the base class of all exceptions so catching it is functionally no different than the previous form.

genofire:
Is it really nessasary? If there any else exception thrown, it will still crash ...

---

Done

ryneeverett:
Looks good.
> Is it really nessasary? If there any else exception thrown, it will still crash ...
> 
True, but it would crash for a different reason and you’d get a stack trace to debug with rather than a vague or inaccurate debug message. I suppose in this case it might be tolerable since this is an optional feature but I’d rather people come here and submit bugs and patches when bugwarrior crashes rather than bugwarrior perform incorrectly.

genofire:
Done, i have not seen it - would be a CI job for syntax checks not great.

genofire:
Thank you - i know i was not easy.

ryneeverett:
Thank you for sticking with it. Anyone willing to accept feedback is a welcome contributor here.

PR #581, opened 2018-05-30
gitlab WIP status is True or False. This does not convert to numerics.

fmauch:
I always ran into errors when pulling in gitlab MRs which changed their WIP status. Today I found some time to debug this and it seems to be a type error. The WIP status seems to be either 'False' or 'True' coming from the server, but the UDA is set to numeric. I changed it to string and now my issues drop in without problems. 

As far as I can see, this value is used nowhere else, so this MR shouldn't do any harm.

Oh, btw, I'm using Gitlab 10.8.2. I don't know if they ever changed their API on that, just in case...

### DIFF
--- a/tests/test_gitlab.py
+++ b/tests/test_gitlab.py
@@ -54,66 +54,66 @@ class TestGitlabIssue(AbstractServiceTest, ServiceTest):
         'gitlab.login': 'arbitrary_login',
         'gitlab.token': 'arbitrary_token',
     }
-    arbitrary_created = (
-        datetime.datetime.utcnow() - datetime.timedelta(hours=1)
-    ).replace(tzinfo=pytz.UTC, microsecond=0)
-    arbitrary_updated = datetime.datetime.utcnow().replace(
-        tzinfo=pytz.UTC, microsecond=0)
-    arbitrary_duedate = (
-        datetime.datetime.combine(datetime.date.today(),
-                                  datetime.datetime.min.time())
-    ).replace(tzinfo=pytz.UTC)
-    arbitrary_issue = {
-        "id": 42,
-        "iid": 3,
-        "project_id": 8,
-        "title": "Add user settings",
-        "description": "",
-        "labels": [
-            "feature"
-        ],
-        "milestone": {
-            "id": 1,
-            "title": "v1.0",
-            "description": "",
-            "due_date": arbitrary_duedate.date().isoformat(),
-            "state": "closed",
-            "updated_at": "2012-07-04T13:42:48Z",
-            "created_at": "2012-07-04T13:42:48Z"
-        },
-        "assignee": {
-            "id": 2,
-            "username": "jack_smith",
-            "email": "jack@example.com",
-            "name": "Jack Smith",
-            "state": "active",
-            "created_at": "2012-05-23T08:01:01Z"
-        },
-        "author": {
-            "id": 1,
-            "username": "john_smith",
-            "email": "john@example.com",
-            "name": "John Smith",
-            "state": "active",
-            "created_at": "2012-05-23T08:00:58Z"
-        },
-        "state": "opened",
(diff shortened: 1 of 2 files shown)


fmauch:
This seems like a good solution! Unfortunately I cannot test it currently, as the gitlab v3 API is no longer supported by gitlab. So currently, my bugwarrior doesn't play with our gitlab instance anymore...

ryneeverett:
Gitlab api v4 support was just merged into the develop branch about a week ago.

fmauch:
I see my mistake... was fetching from the wrong remote. What is the workflow you prefer for diverged branches in MRs? Merge in current dev branch or rebase + force push on my side?

ryneeverett:
Rebase + force push.

fmauch:
alright, it came down to be a one-liner then :)

ryneeverett:
Sweet, we should make sure this is covered by the automated test suite too.

Would be good to add a work_in_progress field to the arbitrary issue and then add another test case and flip the work_in_progress field to make sure both true and false values are tested.

fmauch:
I've added the field to the default testcase and added another testcase for wip=false

tests/test_gitlab.py
```diff
@@ -129,6 +130,43 @@ def test_to_taskwarrior(self):
...
+            issue.ASSIGNEE: 'jack_smith',
+            issue.NAMESPACE: 'arbitrary_namespace',
+            issue.WEIGHT: 3,
+        }
+        actual_output = issue.to_taskwarrior()
+
+        self.assertEqual(actual_output, expected_output)
+
+    def test_work_in_progress(self):
+        arbitrary_issue_2 = self.arbitrary_issue
```
ryneeverett: Assignment like this just adds another label for `self.arbitrary_issue`, it does not copy it.

You could copy it with `arbitrary_issue_2 = copy.deepcopy(self.arbitrary_issue)`, but it would probably be better to skip the copying/relabeling altogether and move the `arbitrary_*` variables from the class scope to the `setUp` method to guarantee testcase isolation.

tests/test_gitlab.py
```diff
@@ -129,6 +130,43 @@ def test_to_taskwarrior(self):
...
+        arbitrary_issue_2['work_in_progress'] = 'false'
```
ryneeverett: Because this line mutates the class attribute `arbitrary_issue`, it will affect subsequent testcases.

tests/test_gitlab.py
fmauch: I understand the problem here...

Do you suggest to define all ```arbitrary_*``` members in the ```setUp``` method instead directily in the class definition? So they would still be members, but defined in the ```setUp``` method which will be called before each test function, is this correct? (Sorry, I am quite new to python unittests).

So I would define ```arbitrary_issue``` as before, but inside the ```setUp``` function and then create ```arbitrary_issue_no_wip``` by creating a deepcopy from ```arbitrary_issue``` and modifying the respective field. I would still do it inside the test function, as this is only relevant for that scope, right?

ryneeverett:
> Do you suggest to define all arbitrary_* members in the setUp method instead directily in the class definition? So they would still be members, but defined in the setUp method which will be called before each test function, is this correct? (Sorry, I am quite new to python unittests).
> 
Correct, but you’ll have to assign to self.arbitrary_* as the setUp methods local variables are not automatically assigned to the class instance.
> So I would define arbitrary_issue as before, but inside the setUp function and then create arbitrary_issue_no_wip by creating a deepcopy from arbitrary_issue and modifying the respective field. I would still do it inside the test function, as this is only relevant for that scope, right?
> 
If you go this route there is no need to copy or rename arbitrary_issue because you can mutate it freely in isolation from the other test cases. That’s the advantage of assigning it in setUp before each test method.

fmauch:
done

ryneeverett:
Thanks!

Issue #551, opened 2018-04-04
Pulling from gitlab is very slow


abderrahim:
I've set it with
```
gitlab.only_if_assigned = True
gitlab.include_todos = True
```
and it takes a very long time
```
INFO:bugwarrior.services:Done with [gnome_gitlab] in 9.695463s
INFO:bugwarrior.services:Done with [gitlab] in 216.326536s
```
to get only 3 and 8 todo items respectively (I have no assigned issue).

Looking at the code, it makes a request for every repository (I'm in three groups with about 50 repositories in total on gitlab.com). This explains the delay.

I don't know if this is a limitation of the v3 api, but looking at the documentation, it is possible (with v4 api) to get all assigned issues with `/issues?scope=assigned-to-me`

This should speed up at least this common case.

ryneeverett:
Closing in favor of #624.

PR #592, opened 2018-07-05
gitlab: fetch annotations only if necessary

wookietreiber:
The title says it all: we don't need to fetch notes if **annotation_comments** is set to **False**.


wookietreiber:
This also helps #551, depending on the **annotation_comments** setting and the number of issues that are synchronized.

PR #593, opened 2018-07-05
gitlab: reduces issue and merge request api calls

### LINKED ISSUE #590: gitlab: reduce number of api requests
The number of API requests for GitLab can be largely reduced when no repository filters are configured **and** an assignee is configured. With that information, we do not need to iterate over all projects but can do only the basic API requests:

```
/issues?assignee_id=X
/merge_requests?assignee_id=X
```

People usually do not have thousands of issues/merge requests assigned, thus each of the requests above would probably result in less than 100 per type (which is below the pagination limit), thus only a single API request per type.

---

In pseudo-code:

```
if include_repos is not defined and
   include_regex is not defined and
   exclude_repos is not defined and
   exclude_regex is not defined and
   only_if_assigned is defined
then
   use /issues?assignee_id=X
else
   iterate over all projects to try and filter
fi
```

---

This would also fix #551. I will prepare a pull request later this week.

wookietreiber:
Probably not the prettiest of codes yet, but it does the job: with 49 issues and 1 merge request, I'm down to 5 API requests (including #592) from about 150.

Note, that the improvement only works if you have:

```
gitlab.only_if_assigned = username
gitlab.also_unassigned = False
```

---

fixes #590

### DIFF
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -250,6 +250,18 @@ class GitlabService(IssueService, ServiceClient):
         self.include_regex = re.compile(self.include_regex) if self.include_regex else None
         self.exclude_regex = re.compile(self.exclude_regex) if self.exclude_regex else None
 
+        self.assignee = self.config.get('only_if_assigned', None)
+
+        if self.assignee:
+            tmpl = '{scheme}://{host}/api/v4/users?username=%s' % self.assignee
+            users = self._fetch_paged(tmpl)
+            for user in users:
+                self.assignee_id = user['id']
+        else:
+            self.assignee_id = None
+
+        self.also_unassigned = self.config.get('also_unassigned', None)
+
         self.import_labels_as_tags = self.config.get(
             'import_labels_as_tags', default=False, to_type=asbool
         )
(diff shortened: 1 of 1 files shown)


ryneeverett:
Would there be a cleaner way to add a couple `if`-statements to the `get_repo_issues` and `get_repo_merge_requests` methods? E.g.:

```py
    def get_repo_issues(self, rid):
        tmpl = '{scheme}://{host}/api/v4/projects/%d/issues?state=opened' % rid
        if self.assignee and not self.also_unassigned:
            tmpl+='&assignee_id=%s' % self.assignee_id
        issues = {}
        try:
            repo_issues = self._fetch_paged(tmpl)
        except IOError:
            # Projects may have issues disabled.
            return {}
        for issue in repo_issues:
            if self.assignee and not self.also_unassigned:
                rid = issue['project_id']
            issues[issue['id']] = (rid, issue)
        return issues
```

wookietreiber:
I just realized, while reviewing #605, that the `assignee_id`-based template I'm using more or less ignores all of the include and exclude filters. This PR will most likely fail in the `for rid, issue in issues:` loop, if either one of the repos is excluded or not included because `repo_map` would not have that repo ID.

ryneeverett:
> I just realized, while reviewing #605, that the `assignee_id`-based template I'm using more or less ignores all of the include and exclude filters. This PR will most likely fail in the `for rid, issue in issues:` loop, if either one of the repos is excluded or not included because `repo_map` would not have that repo ID.

So this is in WIP status, right?

wookietreiber:
Yes it is WIP.

ralphbean:
Have you had any luck with this one @wookietreiber?

wookietreiber:
@ralphbean I've been using this patch since I proposed it and it works really well. I just didn't merge because it's not the prettiest of codes and also because of the codecov miss. I don't have the capacity to fix this ATM, so feel free to take over.

ralphbean:
Good to know!  I'll give it a look the next chance I get.  (Will be a little while still.   Time comes in spurts.)

ryneeverett:
Superseded by #903.

Issue #597, opened 2018-08-07
Python 3.7: configparser: TypeError: option values must be strings


wookietreiber:
On a fresh (2018-08-07) Arch Linux with Python 3.7:

```
Traceback (most recent call last):
  File "/home/wookietreiber/.apps/bugwarrior/bin/bugwarrior-pull", line 11, in <module>
    load_entry_point('bugwarrior==1.5.1', 'console_scripts', 'bugwarrior-pull')()
  File "/home/wookietreiber/.apps/bugwarrior/lib/python3.7/site-packages/click-6.7-py3.7.egg/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/wookietreiber/.apps/bugwarrior/lib/python3.7/site-packages/click-6.7-py3.7.egg/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/wookietreiber/.apps/bugwarrior/lib/python3.7/site-packages/click-6.7-py3.7.egg/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/wookietreiber/.apps/bugwarrior/lib/python3.7/site-packages/click-6.7-py3.7.egg/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/wookietreiber/.apps/bugwarrior/lib/python3.7/site-packages/bugwarrior-1.5.1-py3.7.egg/bugwarrior/command.py", line 62, in pull
    config = _try_load_config(main_section, interactive)
  File "/home/wookietreiber/.apps/bugwarrior/lib/python3.7/site-packages/bugwarrior-1.5.1-py3.7.egg/bugwarrior/command.py", line 35, in _try_load_config
    return load_config(main_section, interactive)
  File "/home/wookietreiber/.apps/bugwarrior/lib/python3.7/site-packages/bugwarrior-1.5.1-py3.7.egg/bugwarrior/config.py", line 214, in load_config
    config = BugwarriorConfigParser({'log.level': "INFO", 'log.file': None})
  File "/usr/lib64/python3.7/configparser.py", line 638, in __init__
    self._read_defaults(defaults)
  File "/usr/lib64/python3.7/configparser.py", line 1216, in _read_defaults
    self.read_dict({self.default_section: defaults})
  File "/usr/lib64/python3.7/configparser.py", line 753, in read_dict
    self.set(section, key, value)
[...]

PR #600, opened 2018-08-10
set allow_no_value=True so that it doesn't break on Python 3.7

archite:
Resolves the issue where one gets the error "configparser: TypeError: option values must be strings" as in #597.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
  - "3.4"
  - "3.5"
  - "3.6"
+ - "3.7"
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install uuid-dev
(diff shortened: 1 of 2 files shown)


wookietreiber:
Please also add Python 3.7 to travis, so this is properly checked.

wookietreiber:
My bad, seems travis does not yet support Python 3.7 :(

The other travis errors seem to be sporadic `apt` issues. I guess, we can either restart or ignore these.

archite:
I assumed it might have had it. I’m not much for Travis as I use GitLab for work and personal.

ryneeverett:
Merged da9221ea673b32fe3f9732b13c818604cf657230.

PR #605, opened 2018-09-04
Make include_repos work the way the documentation suggests

gms8994:
The [documentation](https://bugwarrior.readthedocs.io/en/latest/services/gitlab.html) suggests the following, but actually seems to pull in all of the available projects on the gitlab server - and on gitlab.com, that's a lot of projects.

`If you happen to be working with a large number of projects, you may want to pull issues from only a subset of your repositories. To do that, you can use the gitlab.include_repos option.`

This change performs individual pulls for the projects listed in `include_repos` instead of paginating through the entire project list.

### DIFF
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -4,6 +4,7 @@ standard_library.install_aliases()
 from builtins import map
 from builtins import filter
 
+from urllib import quote_plus
 from configparser import NoOptionError
 import re
 import requests
@@ -433,7 +434,19 @@ class GitlabService(IssueService, ServiceClient):
 
     def issues(self):
         tmpl = '{scheme}://{host}/api/v4/projects'
-        all_repos = self._fetch_paged(tmpl)
+
+        all_repos = []
+        if self.include_repos:
+            for repo in self.include_repos:
+                indiv_tmpl = tmpl + '/' + quote_plus(repo)
+                item = self._fetch(indiv_tmpl)
+                if not item:
+                    break
+
+                all_repos += [ item ]
+        else:
+            all_repos = self._fetch_paged(tmpl)
+
         repos = list(filter(self.filter_repos, all_repos))
 
         repo_map = {}


bugwarrior/services/gitlab.py
```diff
@@ -433,7 +434,19 @@ def include_todo(todo):
...
+                indiv_tmpl = tmpl + '/' + quote_plus(repo)
+                item = self._fetch(indiv_tmpl)
+                if not item:
+                    break
+
+                all_repos += [ item ]
+        else:
+            all_repos = self._fetch_paged(tmpl)
+
         repos = list(filter(self.filter_repos, all_repos))
```
wookietreiber: `include_repos` is used in the `filter_repos` function. You should remove this line here and rename `all_repos` to `repos`.

Edit: ... otherwise, you're filtering twice.

wookietreiber:
FYI: See also #593 for an alternative way to reduce API requests.

gms8994:
@wookietreiber that's actually what sent me down the path of looking at this. I noticed that I'd specified `include_repos` in my config, but that the code still fetches "all" repos, and then filters them after the fact, making an update run take almost 10 minutes.

bugwarrior/services/gitlab.py
wookietreiber: ... or maybe even leave the code here as is and remove the `include_repos` filtering from `filter_repos`, because `filter_repos` also filters the `exclude_repos`.

gms8994: What about renaming the `all_repos` on 446 to repos and moving the `self.filter_repos` bit up in to the else? That way, I'm only filtering if it went through the normal fetch methods?

bugwarrior/services/gitlab.py
```diff
@@ -433,7 +434,19 @@ def include_todo(todo):
 
     def issues(self):
         tmpl = '{scheme}://{host}/api/v4/projects'
-        all_repos = self._fetch_paged(tmpl)
+
+        all_repos = []
+        if self.include_repos:
+            for repo in self.include_repos:
```
wookietreiber: You are using `include_repos` here, however, there is also `include_regex`, which you should check on as well.

bugwarrior/services/gitlab.py
wookietreiber: Then, you would leave out the `exclude_repos` and the `exclude_regex` filters. All of these four configuration settings can be used at the same time, e.g.:

```ini
gitlab.include_regex = foo/bar/*
gitlab.exclude_repos = foo/bar/baz
```

Get everything from the **foo/bar** sub-group except this one **baz** project.

bugwarrior/services/gitlab.py
gms8994: To allow `include_regex` would require fetching all the projects still; is that a correct statement?

bugwarrior/services/gitlab.py
gms8994: Removed include_repos filter from filter_repos check in 3417384

bugwarrior/services/gitlab.py
wookietreiber: Possibly. Probably the best solution would be to find a GitLab APIv4 call that returns a list of repos based on a regex, but I don't know if there is such a call. Might be worth suggesting that to the GitLab API devs, this would save projects like bugwarrior using the API a lot of time.

gms8994: I've added code to handle the `include_regex` functionality in 04f6e2d

gms8994: Gitlab API allows for searching projects, but not by regex; I'll investigate opening an issue with their core functionality to see if they'd be willing to provide that.

bugwarrior/services/gitlab.py
```diff
@@ -433,7 +431,21 @@ def include_todo(todo):
...
+                indiv_tmpl = tmpl + '/' + quote(repo, '')
+                item = self._fetch(indiv_tmpl)
+                if not item:
+                    break
```
ryneeverett: I see that the other invocation of `_fetch` has this `if not`/`break` clause but do we know what circumstances would cause the `item` to be falsey?

bugwarrior/services/gitlab.py
```diff
@@ -433,7 +431,21 @@ def include_todo(todo):
 
     def issues(self):
         tmpl = '{scheme}://{host}/api/v4/projects'
-        all_repos = self._fetch_paged(tmpl)
+
+        all_repos = []
+        if self.include_repos:
```
ryneeverett: I think we should make this `if self.include_repos and not self.include_regex` to avoid doing these extra API calls if we're just going to throw out the results and fetch all the repos anyway.

bugwarrior/services/gitlab.py
```diff
@@ -433,7 +431,21 @@ def include_todo(todo):
...
+            for repo in self.include_repos:
+                indiv_tmpl = tmpl + '/' + quote(repo, '')
+                item = self._fetch(indiv_tmpl)
+                if not item:
+                    break
+
+                all_repos += [ item ]
+
+        if (self.include_regex is not None
+            or not self.include_repos):
```
ryneeverett: I think this should just be an `else` statement (per above change recommendations).

bugwarrior/services/gitlab.py
```diff
@@ -314,12 +318,6 @@ def filter_repos(self, repo):
         # fallback if no filter is set
         is_included = True
 
-        if self.include_repos:
-            if repo['path_with_namespace'] in self.include_repos:
-                return True
-            else:
-                is_included = False
-
```
ryneeverett: I think these lines should not be removed for the circumstance where the `include_repos` and `include_regex` settings are both being used.

bugwarrior/services/gitlab.py
gms8994: Updated in 849494e

bugwarrior/services/gitlab.py
gms8994: Updated in 849494e

bugwarrior/services/gitlab.py
gms8994: My only thought would be if the API decided to return something that isn't a repo as normally expected. I'm happy to adjust to make the logic a little more strict than just a `falsey` check.

bugwarrior/services/gitlab.py
gms8994: Good call - I've added them back in with 849494e

PR #611, opened 2018-09-26
Log exceptions during command execution with their backtrace

urzds:


urzds:
> Suppressing this exception output was a deliberate decision in #285:
> 
> > The problem that I'm trying to solve is that the important error message (the fact that the oracle command failed) is buried over five pages of mostly irrelevant output. First die outputs 4 pages of example configs (and in the case of an error in the oracle, none of it is relevant) and then there are the stack traces of SystemExit and RuntimeError. In [a4e6744](https://github.com/ralphbean/bugwarrior/commit/a4e6744bddf49b714c3d867490a05a89b4468e11), I suggest to catch those exceptions separately mostly to suppress the stack trace. But that still leaves me with four pages to scroll up to see what went wrong.
> 
> That's not to say your patch is wrong but it needs an accompanying justification.

My reason was that some backends contain bugs and will eventually cause the whole synchronisation for all remotes to abort. If we do not catch the exception, there will be no information whatsoever on what went wrong, except that this or that remote failed and aborted.

> Also, the whole point of catching the `RuntimeError` appears to be suppression of the traceback. If we're not going to suppress it or do anything with it is there any point of catching it?

If we do not catch it, it would take down the whole process, would it not? If we `log` it instead, we allow the other parts to finish synchronisation?

Issue #612, opened 2018-10-01
gitlab keeps fetching projects forever


Valodim:
Using this config:

```
[gitlab_valodim]
service = gitlab
gitlab.login = Valodim
gitlab.token = TOKEN
gitlab.host = gitlab.com
gitlab.include_repos = nonpublic/internal

gitlab.only_if_assigned = True
gitlab.include_todos = True
```

I get this debug output:

```
INFO:bugwarrior.services:Starting to aggregate remote issues.
INFO:bugwarrior.services:Spawning 1 workers.
INFO:bugwarrior.services:Working on [gitlab_valodim]
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gitlab.com:443
DEBUG:urllib3.connectionpool:https://gitlab.com:443 "GET /api/v4/projects?page=1&per_page=100 HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gitlab.com:443
DEBUG:urllib3.connectionpool:https://gitlab.com:443 "GET /api/v4/projects?page=2&per_page=100 HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gitlab.com:443
DEBUG:urllib3.connectionpool:https://gitlab.com:443 "GET /api/v4/projects?page=3&per_page=100 HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gitlab.com:443
DEBUG:urllib3.connectionpool:https://gitlab.com:443 "GET /api/v4/projects?page=4&per_page=100 HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gitlab.com:443
DEBUG:urllib3.connectionpool:https://gitlab.com:443 "GET /api/v4/projects?page=5&per_page=100 HTTP/1.1" 200 None
```

I've let it run for a very long time (minutes) and it just keeps incrementing the page counter. I've varied some of the settings, but none of them seem to make a difference, except for `gitlab.token` which causes an unauthorized error if I flip a bit. In my gitlab there are only 10 or so repositories under "Your Projects / All", so it's not that I have ridiculous amounts of repositories either.

This is on bugwarrior-1.6.0, installed from pip.

logankoester:
This is happening for me too, both from pip and also with both `master` and `develop` branches from git.

plup:
[...]

PR #613, opened 2018-10-03
Update gitlab service to correct project request

carlitux:


wookietreiber:
Please explain the reasoning for adding these arguments. Your PR does not include this information yet and neither does your commit message. Please explain what these arguments do individually and in combination. This might not be what everyone wants, and if so, it should be configurable.

carlitux:
@wookietreiber sorry for this. 

Please refer this issue https://github.com/ralphbean/bugwarrior/issues/612

Gitlab Doc about `List All Projects`

> Get a list of all visible projects across GitLab for the authenticated user. When accessed without authentication, only public projects with "simple" fields are returned.

So we don't need to collect all projects, that is why we have this issue https://github.com/ralphbean/bugwarrior/issues/612 never finish

> **owned** - Limit by projects explicitly owned by the current user

We only need issues from my repositores

> **membership** - Limit by projects that the current user is a member of

We only need if I am member of a repository

> **simple** - Return only limited fields for each project. This is a no-op without authentication as then only simple fields are returned.

We only need a few project fields so this is performant

Valodim:
Thanks for working on this! I gave this a shot, but received the following error:

```
  File "/home/valodim/.local/lib/python2.7/site-packages/bugwarrior/services/gitlab.py", line 362, in _fetch
    return self.json_response(response)
  File "/home/valodim/.local/lib/python2.7/site-packages/bugwarrior/services/__init__.py", line 473, in json_response
    response.status_code, response.url, response.text,
IOError: Non-200 status code 400; u'https://gitlab.com/api/v4/projects?owned=true&membership=true&simple=true/$org/$repo'; u'{"error":"simple is invalid"}'
```

carlitux:
@Valodim any setup that has the value `/$org/$repo`?

Valodim:
only tried with one. do you think it could be repo-specific? the URI with "&simple=true/org/repo" looks like it might be malformed, a path following parameters is unusual at least. that would also match with the "simple is invalid" error

carlitux:
@Valodim update and try again please

Valodim:
Seems like that issue is fixed, thanks! There's another problem however:

```
  File "/home/valodim/.local/lib/python2.7/site-packages/bugwarrior/db.py", line 388, in synchronize
    tw.task_add(**issue)
  File "/home/valodim/.local/lib/python2.7/site-packages/taskw/warrior.py", line 683, in task_add
    stdout, stderr = self._execute('add', *args)
  File "/home/valodim/.local/lib/python2.7/site-packages/taskw/warrior.py", line 476, in _execute
    raise TaskwarriorError(command, stderr, stdout, proc.returncode)
TaskwarriorError: ['task', 'rc:/home/valodim/.taskrc', 'rc.uda.gitlabduedate.type=date', 'rc.uda.gitlabduedate.label="Gitlab Due D
...
The value '14141016' is not a valid numeric value.
```

wookietreiber:
@carlitux Thank you for the explanation.

## about `simple`

The `simple` argument should work in all cases. At the time of this writing, we only need the `id` and `path_with_namespace` fields and both are included in the `simple` result. This argument is definitely a worthwhile addition to reduce network traffic. Good catch!

## about `owned` and `membership`

There are different use cases and appropriate filters that reduce time to synchronize. Adding ad-hoc arguments to base queries has the potential to disrupt use cases:

1.  If I really want to fetch all issues, even of projects I do not own or are a member of, the query would exclude these projects.
1.  Using **both** `owned` **and** `membership` also does not work as you expect. You only get projects that you own ***and*** are a member of. This effectively makes the `membership` argument redundant and excludes projects that you do not own.

To make the queries work for everyone, **all query arguments *that filter* should be optional / configurable**. For this reason, we should not accept this pull request ***as is***.

@carlitux You should turn both `owned` and `membership` into options that can be configured individually in `bugwarriorrc`. Also, you should document these options accordingly in the GitLab service documentation.

wookietreiber:
Also, I recommend to split this merge request in three separate ones: one merge request for each new query argument, at least one commit for each new query argument.

bugwarrior/services/gitlab.py
```diff
@@ -441,15 +441,15 @@ def issues(self):
...
+                indiv_tmpl = tmpl + '/' + quote(repo, '') + '?simple=true'
                 item = self._fetch(indiv_tmpl)
                 if not item:
                     break
 
                 all_repos += [ item ]
 
         else:
-            all_repos = self._fetch_paged(tmpl)
+            all_repos = self._fetch_paged(tmpl + '?owned=true&membership=true&simple=true')
```
wookietreiber: The `owned` and `membership` query arguments should be optional / configurable.

wookietreiber: Each new query argument should get its own commit.

doronbehar:
Nice work on the fix for #612. Are we expecting new commits for this PR? I would really like to stop manually adding tasks for every GitLab issues I get or have in my TODO.

Thanks.

carlitux:
@doronbehar I will update soon this. I am with a few time.

Valodim:
Great to hear, and thanks again for your work so far! :+1:

carlitux:
@wookietreiber please let me know your thoughts.

psukys:
@carlitux I think gitlab tests can as well be adjusted, since the mocked addresses are hardcoded and now is missing the "simple=True".

i.e. https://github.com/ralphbean/bugwarrior/blob/develop/tests/test_gitlab.py#L204

carlitux:
@psukys fixed tests

configparser also added to setup.py

bugwarrior/docs/services/gitlab.rst
```diff
@@ -87,6 +87,20 @@ For example, you want to include only the repositories ``foo/node`` and
...
+
+    gitlab.membership = True
+
+Filtering Owned
+^^^^^^^^^^^^^^^^^^^^
+
+If you want to filter repositories that you own.
+
+    gitlab.owned = True
+
```
ryneeverett: We usually make our filtering options a little more clearly named. You may want to change these to something like `gitlab.only_if_member` and `gitlab.only_if_owned`.

bugwarrior/services/gitlab.py
```diff
@@ -441,15 +445,20 @@ def issues(self):
...
                 if not item:
                     break
 
-                all_repos += [ item ]
+                all_repos.append(item)
 
         else:
-            all_repos = self._fetch_paged(tmpl)
+            querystring = { 'simple': True }
+            if (self.membership):
```
ryneeverett: In python, one-line condition strings aren't typically parenthesized.

bugwarrior/services/gitlab.py
```diff
@@ -441,15 +445,20 @@ def issues(self):
...
+                querystring['membership'] = True
+            if (self.owned):
```
ryneeverett: In python, one-line condition strings aren't typically parenthesized.

ralphbean:
@gustavoheinz is going to look at getting this rebased and cleaned up during the PyCon 2019 sprints.

ralphbean:
Done in #649!

PR #615, opened 2018-10-23
Fix jira sprint parsing

campbellr:
Jira doesn't escape commas in the value field of the sprint string so it's not safe to  just split on all commas.

In my case, I ran into a sprint with a goal that contained a bunch of commas.

Splitting on the `=` character and then using `rsplit` to only grab the last `,` is maybe a bit safer.


campbellr:
I took a look at the Travis failures, and they seem to be unrelated. On was a failing Gerrit test, the other was an error downloading from AWS...

ryneeverett:
Thanks!

PR #618, opened 2018-11-10
Fix tests on py27 due to configparser

ssbarnea:
Uses the portable configparser name from six to avoid failure on py27.


ssbarnea:
I wonder how it was possible to endup with this issue while having CI enabled on the repository....direct push?

wookietreiber:
Please merge, other PRs like #621 depend on it / need to be rebased onto it.

PR #621, opened 2018-11-21
[WIP] Add state and closed uda to GitHub

pawamoy:
When pulling closed issues (with custom query), bugwarrior will use the `closed_at` field from GitHub to set the `end` value of the task and mark it as "done".

Two new UDAs in this PR:
- `githubclosedon` (to be consistent with `githubcreatedon`): date
- `githubstate`: string, which is the state of the issue/PR (`open`, `closed` or `merged`)

TODO:
- update docs


pawamoy:
Python 2 jobs fixed by #618

tests/test_github.py
```diff
@@ -15,6 +15,9 @@
 ARBITRARY_CREATED = (
     datetime.datetime.utcnow() - datetime.timedelta(hours=1)
 ).replace(tzinfo=pytz.UTC, microsecond=0)
+ARBITRARY_CLOSED = (
+    datetime.datetime.utcnow() - datetime.timedelta(hours=2)
+).replace(tzinfo=pytz.UTC, microsecond=0)
```
wookietreiber: You are closing this issue 1 hour **before** you create it. It does not effect any tests because there aren't any yet, but I would invert these, so we don't bump our heads later.

wookietreiber:
I'm wondering if that error in travis is spurious or not:

```
Failure: ImportError (No module named configparser) ... ERROR
```

It only effects Python 2.7 ...

pawamoy:
Yeah I don't know why it appears only now and not in previous builds. PR #618 fixes it though.

tests/test_github.py
pawamoy: Ah, good catch! I did not pay enough attention. Thanks!

wookietreiber:
LGTM :+1: 

I suppose you tested this locally: does this also create completed tasks from closed issues or is it just closing issues that were created by bugwarrior before? I would really like this for my burndown charts :)

pawamoy:
> I would really like this for my burndown charts :)

This is exactly why I opened this PR :grinning: !

Yes, it will create completed tasks from closed issues, **but** (and I'm not targeting you particularly but writing for everyone instead) you have to understand two things:
1. how bugwarrior closes tasks by default: the default behavior is that bugwarrior pulls **only** open issues. In subsequent pulls, each and every bugwarrior task that is **missing from the pull** is closed. Something like:
  ```
  pull A B C from service
  create A B C in taskwarrior
  close B online
  pull A C from service
    B is missing -> complete B
  ```
2. this PR adds a new behavior: if the pulled issue has an `end`, then bugwarrior will complete it after adding (it did not existed before) or modifying it (it was already pulled before). But that means you **have** to pull closed issues, and you can only do this by writing a custom `SERVICE.query`. For GitHub, that would be something like `github.query = involves:pawamoy archived:false`. Note the absence of `state:open` in the query. A more concrete example would be:
  ```
  [github_pawamoy]
  service = github

  github.login = pawamoy
  github.token = <token>

  github.include_user_issues = True  # include the issues I opened
  github.include_user_repos = True   # include the issues of my repositories
  github.involved_issues = False     # this is overwritten with a custom query to include closed issues

  github.query = involves:pawamoy archived:false  # I'm not interested in archived issues
  # I could also have played with author: assignee: and mentions:
  # Try it online before, to see what works and what doesn't
  ```

With the example in the second point above, a few questions come to mind:
-   Does `github.include_user_issues = True` include closed issues I opened? I don't think so, but this is fixed with the custom query.
- Does `github.include_user_repos = True` include closed issues opened by others on my repositories? I don't think so, and this is a problem because even the custom query could miss some of them (issue opened and closed by others than me, without me commenting, assigning to it, or being mentioned).

So, my conclusion is that bugwarrior **should offer a way to include closed issues**. A simple flag like `SERVICE.include_closed_issues` would do. Of course the service in question must be able to tell the issue is closed (and when) to set the `end` value. Otherwise bugwarrior creates those closed issues as normal, non-completed tasks.

I'm gonna add this to my PR in the docs somewhere, maybe also in a contributing section :slightly_smiling_face:

wookietreiber:
Both `include_user_repos` and `include_user_issues` fetch only open issues. To change this, you would need to add `state=all` to their respective queries, i.e. [here](https://github.com/ralphbean/bugwarrior/blob/2841f05ae66a0d8f8c91c03ec896cad6b8d2cb0a/bugwarrior/services/github.py#L59) and [here](https://github.com/ralphbean/bugwarrior/blob/2841f05ae66a0d8f8c91c03ec896cad6b8d2cb0a/bugwarrior/services/github.py#L48).

pawamoy:
Thanks, we could start by implementing this for GitHub indeed :slightly_smiling_face: In another PR though.

pawamoy:
Added docs for the two new UDAs. Ready to merge @wookietreiber @ralphbean !

bugwarrior/services/github.py
```diff
@@ -198,16 +208,17 @@ def to_taskwarrior(self):
...
+        updated = self.parse_date(self.record.get('updated_at'))
+        closed = self.parse_date(self.record.get('closed_at'))
 
         return {
             'project': self.extra['project'],
             'priority': priority,
             'annotations': self.extra.get('annotations', []),
             'tags': self.get_tags(),
             'entry': created,
+            'end': closed,
```
ryneeverett: Is there a reason we're calling this `end` and not `closed`?

PR #635, opened 2019-01-23
Add option to specify specific issues in github

### LINKED ISSUE #516: Feqture Request: Adding tasks according to an arbitrary list of issues' URL in `bugwarriorrc`
Sometimes there are issues in projects I generally don't make contributions to, yet I would like to be the one who will fix them. It would be great if one could make `bugwarrior-pull` read a list of urls that points to issues one would like to add to his/her task list. Additionally if he/she fixes those issues in a PR and the PR is accepted, the issues will be marked as completed.

What do you think?

doronbehar:
Fixes #516. Documentation was added as well, I hope it's OK. I've also prepared Gitlab support but it can't be used until #613 is merged.

### DIFF
--- a/bugwarrior/docs/services/github.rst
+++ b/bugwarrior/docs/services/github.rst
@@ -146,14 +146,19 @@ issues matched by the query, set::
 Include Certain Issues
 ++++++++++++++++++++++
 
-Since Github doesn't allow to show on the website or even directly with their
-API issues a user is subscribed to (see [#] and [#]), Bugwarrior allows to
-specify specific issues you'd like to be added to your Taskwarrior database::
+GitHub can show you a list of issues you are subscribed to in their website in the URL:
+
+https://github.com/notifications/subscriptions
+
+This could have been very useful for Bugwarrior but unfortunately there is no
+way to access this list of issues through the API. (See [#] and [#]). However,
+Bugwarrior allows to specify specific issues you'd like to be added to your
+Taskwarrior database::
 
     github.include_these_issues = octokit/octokit.rb/issues/1057, octokit/octokit.rb/issues/957
 
 Note that you need to specify only the part that comes after the host and the
-scheme, usually what's after ``https://github.com/``, including the last ``/``.
+scheme, usually what's after ``https://github.com/``, including the last ``/``
 
 .. [#] https://github.com/dear-github/dear-github/issues/247
 .. [#] https://stackoverflow.com/questions/25438721/how-can-i-see-all-the-issues-im-watching-on-github
--- a/bugwarrior/docs/services/github.rst
+++ b/bugwarrior/docs/services/github.rst
@@ -143,6 +143,21 @@ issues matched by the query, set::
     github.include_user_issues = False
     github.include_user_repos = False
 
+Include Certain Issues
+++++++++++++++++++++++
+
+Since Github doesn't allow to show on the website or even directly with their
+API issues a user is subscribed to (see [#] and [#]), Bugwarrior allows to
+specify specific issues you'd like to be added to your Taskwarrior database::
+
+    github.include_these_issues = octokit/octokit.rb/issues/1057, octokit/octokit.rb/issues/957
+
+Note that you need to specify only the part that comes after the host and the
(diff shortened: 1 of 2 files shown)


ralphbean:
Hey!  I saw in dear-github/dear-github#247 that it looks like this may not be necessary any more.

https://github.com/notifications/subscriptions

WDYT, will that work @doronbehar?

doronbehar:
Wow thanks for giving this update :) It looks great but I couldn't find an API equivalent.. I tried `https://api.github.com/issues?filter=subscribed&state=open&sort=updated`  (according to [this](https://developer.github.com/v3/issues/#parameters)) and it returned only one issue for me while my https://github.com/notifications/subscriptions has 751 issues/PRs

bugwarrior/services/github.py
```diff
@@ -49,6 +49,17 @@ def get_issues(self, username, repo):
             username=username, repo=repo)
         return self._getter(url)
 
+    def get_issue_from_path(self, issue_path):
+        """Get issue from issue_path."""
+        # TODO: test new scheme of issue_path
+        regex = r'(^[^/]+)(/)([^/]+)(/issues/[0-9]+)'
+        if not re.match(regex, issue_path):
```
ralphbean: Can you do this validation step in the ``__init__`` method if `include_these_issues` is set?  That should catch things earlier.

bugwarrior/services/github.py
```diff
@@ -49,6 +49,17 @@ def get_issues(self, username, repo):
...
+            log.error("issue path given in config is invalid")
+            raise ValueError
```
ralphbean: Here, just do ``raise ValueError("issue path %s given in config is invalid" % issue_path)`` which will give the user more information.

bugwarrior/docs/services/github.rst
```diff
@@ -135,6 +135,21 @@ issues matched by the query, set::
...
+Since Github doesn't allow to show on the website or even directly with their
+API issues a user is subscribed to (see [#] and [#]), Bugwarrior allows to
+specify specific issues you'd like to be added to your Taskwarrior database::
+
+    github.include_these_issues = octokit/octokit.rb/issues/1057, octokit/octokit.rb/issues/957
+
+Note that you need to specify only the part that comes after the host and the
+scheme, usually what's after ``https://github.com/``, including the last ``/``.
+
+.. [#] https://github.com/dear-github/dear-github/issues/247
```
ralphbean: Maybe update these links to show the new subscription UI, but mention the lack of an API - which may help people understand what this is for.

ralphbean:
Huh - I'm really surprised that `subscribed` API doesn't work.  :(

bugwarrior/services/github.py
doronbehar: Good idea. Unfortunately, I've tried that by reading these issues from the config with `self.config.get('include_these_issues')` in `__init__` but I get an error:
```
AttributeError: 'GithubClient' object has no attribute 'config'
```

doronbehar:
> Huh - I'm really surprised that `subscribed` API doesn't work. :(

I invite you to check so for yourself.. maybe I've missed something in my queries..

bugwarrior/services/github.py
doronbehar: Good idea. I consider this review resolved, it's just that we need to decide whether to try to check this regex match in `__init__` or not.. See my previous comment.

bugwarrior/services/github.py
ralphbean: Ah - try instead in the `__init__` method of the `GithubService` class instead of the `GithubClient` class.  See where all those `self.client....` calls are concentrated?

doronbehar:
I'm sorry but I'm planning to create a project that will replace yours. No offense, it's just that I would like to create an equivalent of my own with a much much more simpler interface. I hope that with my design, very specific PRs like these won't arise be needed since it will be much more extensible by design.

Therefor, I don't care if you don't merge these changes. Feel free to close.

ralphbean:
Oh, that's all good.  Happy hacking, @doronbehar!

PR #642, opened 2019-04-16
Add the ability to exclude Github pull-requests

danihodovic:
In my workflow I only track issues and not pull requests. Pull requests usually refer to issues that I resolve so tracking both results in duplicate tasks.


ralphbean:
Thanks!

PR #645, opened 2019-05-01
jira: split priority.name for lookup

stashlukj:
The priority.name field returns something more like "1 - Critical". Split the string to get the PRIORITY_MAP key.

### DIFF
--- a/bugwarrior/services/jira.py
+++ b/bugwarrior/services/jira.py
@@ -223,7 +223,9 @@ class JiraIssue(Issue):
             value = value['name']
         except (TypeError, ):
             value = str(value)
-        return self.PRIORITY_MAP.get(value, self.origin['default_priority'])
+        # priority.name format: "1 - Critical"
+        map_key = value.strip().split()[-1]
+        return self.PRIORITY_MAP.get(map_key, self.origin['default_priority'])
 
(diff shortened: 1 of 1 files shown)


bugwarrior/services/jira.py
```diff
@@ -223,7 +223,9 @@ def get_priority(self):
             value = value['name']
         except (TypeError, ):
             value = str(value)
-        return self.PRIORITY_MAP.get(value, self.origin['default_priority'])
+        # priority.name format: "1 - Critical"
+        map_key = value.strip().split()[-1]
```
ralphbean: I'm worried this might screw up some other people if they have custom priorities set to something like "Very Important" which would split their string into just "Important".

Can you think of a way to make this formatting configurable?

stashlukj: This module could query the priority list and map accordingly. That will require a little bit more work. In the meantime, priority imports don't work for the default behavior of several JIRA versions. This module has a fixed mapping, so doesn't support custom priorities anyway.

ralphbean: Ah, fair.  I'm game to merge this as is and take on custom priorities at a later time.  Thanks!

PR #647, opened 2019-05-06
Update python3 versions.

ralphbean:
Just moving the python3 versions here.  No need to test against
those older py3 versions anymore.


ralphbean:
Looks like old versions of python-jira uses the `async` keyword (from py3.7) as a keyword argument in a function call, which breaks the world when we import it.  They've fixed it in python-jira 2.0.0.  That makes the test matrix a little complicated here, but workable.

PR #648, opened 2019-05-06
optionally set bug to "started" when assigned

mathiashls:
Further changes in @beav PR #405


ralphbean:
Thanks @mathiashls!

FYI, @beav, we're merging this over here as part of the PyCon Sprints :)

Thanks for the work on #405!

PR #671, opened 2019-05-30
Store RFC Message-ID from Gmail

marcfargas:
Currently, the gmail service stores the the Thread Id in a UDA. this is fine as long as you live in Gmail API World.

But if you are not always within the boundaries of Gmail API like, say, you access email via IMAP with mutt or whatever, the Thread Id is of no use.

Luckily, there is a standard field that can be used, RFC2822 Message-ID!! The field is avaiable on the API, thus it can easily be stored in an UDA just like last sender.

This PR does just that : add a "Last RFC2822 Message-ID" UDA. This enables workflows like:

```bash
taskalot() {
  MSGID=$(task _get $1.gmaillastmessageid)
  if [ ! -z "$MSGID" ]; then
    alot search id:"$MSGID"
  else
    echo "Task $1 does not define gmaillastmessageid UDA."
  fi
}
```

Hope it's useful and can be merged.


ralphbean:
804dfc29beb6b249bacfac01abb3d9e9fac3aff4 LGTM.  There are some other commits that got in here through some git shenanigans.  I'll cherry pick that one.

ralphbean:
Ah, the tests need updating for this.  :/

ralphbean:
Cherry-picked in 4e3c039 and tests fixed in f8ccd0a.

Issue #658, opened 2019-05-08
Completed tasks are ignored


telenieko:
Hi,

Sample workflow:
- I receive an email, put "TODO" label
- gmail service in bugwarrior pulls that email (query = label:TODO)
- I complete the task
- gmail service pulls the email again

As far as I understand the code, this is due bugwarrior only considering tasks which are either pending or waiting, but not completed (db.py, find_local_uuid also get_managed_task_uuids)

But this means that one has to complete the task and remove the label between two runs of bugwarrior. Am I correct?

If you remove the label first only, the task is deleted. If you mark completed first, the task is duplicated.

Shouldn't bugwarrior consider completed tasks also (only in find_local_uuid I think)?

Or, is something wrong in my workflow? 

I am running from the develop branch.

telenieko:
I have done this:

```
diff --git a/bugwarrior/db.py b/bugwarrior/db.py
index df10c26..2b64a68 100644
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -190,6 +190,7 @@ def find_local_uuid(tw, keys, issue, legacy_matching=False):
                 'or': [
                     ('status', 'pending'),
                     ('status', 'waiting'),
+                    ('status', 'completed')
                 ],
             })
```

As a test, and it works properly for gmail (issues are properly updated if still labelled instead of duplicated).

But I have no idea on what sideeffects that could have on other services ?

ralphbean:
Ah, this is kind of intentional.

Consider the use case where you're working with a bug tracker, instead of gmail.

* The issue is opened.
* It is pulled to taskwarrior by github.
* The issue is closed in the tracker.
* Either bugwarrior closes it for you locally, or you mark it as done yourself.
* Later, the issue is re-opened!
* Bugwarrior opens a new task for you to see this.

I think people (including myself) expect bugwarrior to do this for them, so that they can catch situations like this.

In your case, if you mark a email with TODO and bugwarrior pulls it, you should be able to remove the TODO label after which point bugwarrior will mark it as "done" locally (it shouldn't delete it, to my knowledge).

WDYT?

TheNeuralBit:
It seems that bugwarrior just needs to be opinionated about which list (the tracker, or your local tw db) is the source of truth since it can get into these ambiguous situations, and it's been decided that the external tracker is the SoT (which makes sense!).

[...]

PR #672, opened 2019-05-30
Prevent task duplication

marcfargas:
As introduced in #658, ignoring completed tasks during sync can produce duplicates (ie: you mark a task completed that comes from gmail, it will be pulled again and a new task created).

This patchs makes bugwarrior consider completed tasks just right before creating new tasks. This preventing duplicate entries when a task is being ignored on the "normal" filter.


ralphbean:
Left a comment on #658.

ryneeverett:
The reasons not to go in this direction are discussed in #658.

PR #682, opened 2019-08-03
Allow use of oracle statements with bitbucket.

### LINKED ISSUE #579: Store bitbucket OAuth secret in keyring
Firstly, thanks for this amazing tool!

Currently, it seems impossible to store the `bitbucket.secret` in the keyring and get it by setting `bitbucket.secret = @oracle:use_keyring`. As of now, as far as I can tell this will simply interpret the oracle directive as the literal secret and send it to bitbucket, rather than actually performing the lookup.

Though the OAuth secret (ideally) only has read permissions, it is still a secret, and I would prefer to have the ability to encrypt it in my keyring rather than storing it in plaintext in the config file.

On a related note, the docs should ideally also include a hint to chmod the bugwarriorrc file to 600.

ralphbean:
@he-la, can you confirm if this fixes #579?

### DIFF
--- a/bugwarrior/services/bitbucket.py
+++ b/bugwarrior/services/bitbucket.py
@@ -69,11 +69,13 @@ class BitbucketService(IssueService, ServiceClient):
     def __init__(self, *args, **kw):
         super(BitbucketService, self).__init__(*args, **kw)
 
-        key = self.config.get('key')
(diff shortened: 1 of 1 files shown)


ryneeverett:
Superseded by #902.

PR #700, opened 2019-12-17
Generate UDA documentation automatically (WIP)

gdetrez:
This is not really finished but I wanted to get people's opinion before putting more efforts in it.

The idea is to generate the UDA tables in the services documentation from the `UDAS` dictionary in their respective `Issue` class.

I started using bugwarrior with redmine for work and I discovered that it added more UDAs than documented.  I wanted to make a PR with the missing UDAs but got a bit annoyed by all the copy/pasting between python and rst.

So I implemented a small sphinx extension that generates the table instead.  It is used like this:

```rst
 .. udas:: bugwarrior.services.bz.BugzillaIssue 
```

It already fixes many issues with existing UDA documentation (missing UDAs and erroneous names or types).

There are still a few issues and open questions:
- It requires installing all the dependencies when building the docs.  I don't think this is a big requirement but it can be problematic (It doesn't work with the UDAs for the "version one" service right now because of a dependency problem.)  I added a dedicated `docs` tox environment to make generating docs locally easier.
- It doesn't currently build on readthedocs because of ralphbean/taskw#130.
- It uses the `label` field as description.  In most cases if the description was `Foo Bar`, the label would be `Service Foo Bar` (e.g. `Board name` vs `Trello board name`, see example bellow).  It does look a bit silly to have every description prefixed with the service name but I personally can live with it.  Alternatively I could add an option to the sphinx extension to strip the prefix.
- in some few cases, the description and the labels are different; including at least one case where the description uses rst formatting.  If we want to keep the possibility of having a different/richer description, I wast thinking about either adding an optional `help` keyword to the `UDAS` dictionary or extending the sphinx extension to allow overriding descriptions.
- Some UDAs had their type documented as "Text (numeric)".  I'm not completely sure what that meant... (maybe that's just a mistake?)

Example generated table:

image

Since I cannot link to readthedocs, I'm attaching the generated docs as pdf if you want to look at the other tables: [Bugwarrior.pdf](https://github.com/ralphbean/bugwarrior/files/3973560/Bugwarrior.pdf)


ryneeverett:
Continued in #912.

PR #701, opened 2020-01-05
jira: store subtasks in UDA `jirasubtasks`

aleufroy:
Storing the sub-tasks JIRA ids helps users to set task dependencies in TaskWarrior.

IMHO, it's a good starting point to set task dependencies. Setting the
task dependencies is not a trivial process because we need UUIDs which
are not available when we receive tasks from third-party, especially
because they may arrive in any order. Using a hook may solve this
problem but for now, this patch just stores the dependencies
information in UDA.


ralphbean (approved):
Thanks @aleufroy!

PR #702, opened 2020-01-05
jira: talk about api token as password

aleufroy:
Related to https://github.com/ralphbean/bugwarrior/issues/689 

IMHO, having a documentation entry about api token is enough.


ralphbean (approved):
Works for me.  Thanks!

PR #703, opened 2020-01-08
Allow user to synchronize tasks without closing tasks associated with issues that have been closed.

coddingtonbear:
I might have a fairly unique workflow here, but a problem I run into frequently is that I'm expected to report on closed issues during a meeting a day or so after the task is closed, and new issues may have been created and assigned to me during that period that I'd like to pull into my task list. I'd like to keep those issues in my task list so I have an easy-to-access list of tasks for such meetings.  What this PR does is introduce an option allowing one to pull in and update existing issues without closing their associated tasks.

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -49,9 +49,10 @@ def _try_load_config(main_section, interactive=False):
 @click.option('--dry-run', is_flag=True)
 @click.option('--flavor', default=None, help='The flavor to use')
 @click.option('--interactive', is_flag=True)
+@click.option('--no-close', is_flag=True)
 @click.option('--debug', is_flag=True,
               help='Do not use multiprocessing (which breaks pdb).')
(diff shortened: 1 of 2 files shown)


wookietreiber:
I wonder whether a custom report in taskwarrior would be the more obvious choice to get what you want, using a filter the likes of `task list assignee:johndoe closed:yesterday`. That doable? Would be much easier than this hack on bugwarrior?

wookietreiber:
You might need to able to sync closed tasks/issues as well. But there are other issues/PR doing exactly that.

coddingtonbear:
Unfortunately, that won't work for me given that I actually don't use taskwarrior directly for that sort of thing, and instead rely on some synchronization that takes place between my taskwarrior tasks and a trello board somewhere.

That being said, I'm super comfortable just maintaining this functionality in a private fork that I maintain instead of merging my own idiosyncrasies into this package.

PR #706, opened 2020-01-10
Support phabricator.only_if_assigned.

### LINKED ISSUE #705: Phabricator: does not respect SERVICE.only_if_assigned
It looks like the develop branch has a prevision for `phabricator.ignore_{cc,author,owner,reviewers}` settings which might be enough.

It is still strange behaviour that the service does not respect the general settings.

Maybe a reasonable compromise would be to affect the defaults depending on whether `SERVICE.only_if_assigned` is defined or not as follows:

```
phabricator.only_if_assigned = True | PHAB-ID... | non False
# Imply:
phabricator.ignore_cc = True
phabricator.ignore_author = True

# Where as in the default case
phabricator.only_if_assigned = False
# Imply:
phabricator.ignore_cc = False
phabricator.ignore_author = False
```

evilham:
Fixes #705 by having a non `False` value for `phabricator.only_if_assigned`
affect the defaults for `phabricator.ignore_cc` and `phabricator.ignore_author`.

### DIFF
--- a/bugwarrior/docs/services/phabricator.rst
+++ b/bugwarrior/docs/services/phabricator.rst
@@ -95,6 +95,9 @@ Note that there is no way to filter by the reviewer's response (for example, to
 exclude Revisions you have already reviewed). Phabricator does not provide the
 necessary information in the Conduit API.
 
+Furthermore, setting `phabricator.only_if_assigned` to something other than False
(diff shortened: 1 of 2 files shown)


evilham:
It doesn't look to me like the reason why these tests are failing is related to the PR.

ryneeverett (approved):

PR #708, opened 2020-01-26
Notion support

nesaro:
Hi,

I'm trying to make bugwarrior to work with https://notion.so . Please keep in mind that the current python library uses a scraper and notion doesn't have a public API yet.

Beyond this MR, is there any place I can ask for help with development?

- [ ] Tests
- [ ] Documentation


ryneeverett:
No major objections so far. Was there anything in particular you needed help with?

nesaro:
I got some problems with subprocessing as both `bugwarrior` and `notion-api` rely on it.
However when I tried to revisit this the `notion-api` didn't work as expected and I couldn't retrieve the tasks from there. 

Either my understanding of the notion-api is not good enough or it is just that the unofficial api is not that great, but I don't think I can spend that much time on figuring that out and I'd rather wait to the official api to be available. 

I might revisit this later on, sorry for the inconvenience

PR #717, opened 2020-03-04
Add desc and idShort to trello uda

lukas-reineke:
`idShort` is already used in the default description.
`desc` is really useful to have.


ryneeverett (approved):

ryneeverett (changes requested):
Just realized we need to add the new UDA's to [the docs](https://github.com/ralphbean/bugwarrior/blob/develop/bugwarrior/docs/services/trello.rst#provided-uda-fields) as well.

mattcen:
Documentation updated: https://github.com/mattcen/bugwarrior/tree/trello-description-and-shortid-uda

lukas-reineke:
ah forgot about the docs
thanks @mattcen, you want to make a PR to my branch?

ryneeverett (approved):

ryneeverett:
I understand the motivation for the casting change, but I'm thinking the correct solution is to set `"type": "numeric"` for `idShort` since the API specifies it to be an integer.

mattcen:
@ryneeverett that makes sense, but when I try that, it *still* gives me the same output of trying to change from `'1'` to `1`, after multiple runs of `bugwarrior-pull`:

```
mattcen@inga:tmp$ grep -e idShort -e trellocardidshort /usr/local/lib/python3.7/site-packages/bugwarrior/services/trello.py ~/.taskrc 
/usr/local/lib/python3.7/site-packages/bugwarrior/services/trello.py:    SHORTCARDID = 'trellocardidshort'
/usr/local/lib/python3.7/site-packages/bugwarrior/services/trello.py:            number=self.record['idShort'],
/usr/local/lib/python3.7/site-packages/bugwarrior/services/trello.py:            self.SHORTCARDID: self.record['idShort'],
/usr/local/lib/python3.7/site-packages/bugwarrior/services/trello.py:        params = {'fields': 'name,idShort,shortLink,shortUrl,url,labels,due,desc'}
/Users/mattcen/.taskrc:uda.trellocardidshort.type=numeric
/Users/mattcen/.taskrc:uda.trellocardidshort.label=Trello short card ID
```

```
INFO:bugwarrior.db:Updating task 54e5f87f-6be5-4370-b196-5e48aaa7668c, (bw)#912 - Order kitchen timer .. https://trello.com/c/BMTBnTZ1; trellocardidshort: '912' -> 912
```

This is even after I ran `task trellocardidshort.any: modify trellocardidshort=` to unset that attribute. After running that I ran `bugwarrior-pull` twice, The first time it set it from `Null` to `912`, and the second time it set it from `'912'` to `912`. Am I doing something wrong here?

ryneeverett:
<s>I think the issue is that we need to add these new UDA's to `twdict` in the `to_taskwarrior` method. I believe this is the point where bugwarrior parses the data out of the taskwarrior database and then diff's it with the data pulled from remote sources in order to detect changes.</s>

Nevermind, I see that you did this. Hmm...

ryneeverett:
If I get a chance I think i'm going to try adding some tests to the update logic. This seems like a large hole in our standard service testing and there's likely a bunch of these bugs across the services.

ryneeverett:
I did a little digging and it seems that taskwarrior doesn't make any distinction between string and numeric UDA types in input, output, or persistent storage. The distinction is in validation and runtime storage.

Therefore I think the string coercion is fine but the type should still be numeric.

mattcen:
Cool, thanks for investigating that, @ryneeverett! I've sent a PR to @lukas-reineke (https://github.com/lukas-reineke/bugwarrior/pull/3) for this :)

ryneeverett (approved):

mattcen:
Hey, @lukas-reineke, can you please test whether this latest change I made works  for you? I just tested it again, and was getting a `ValueErrorr: Value must be numeric.` from Python. When I took out my `str()` casts, the problem went away *and* I didn't get continuous updates every time I ran `bugwarrior-pull`. Current setup is that I'm running [this version of the file](https://github.com/mattcen/bugwarrior/commit/15d88de3336e1e575c9eef8f488127e1edf0e525), and have ensured I have the following in my config:
```
mattcen@inga:services(trello-description-and-shortid-uda)$ grep -i idshort ~/.taskrc 
uda.trellocardidshort.type=numeric
uda.trellocardidshort.label=Trello short card ID
```

I reckon the reason things weren't working before was because I'd forgotten to make [this change](https://github.com/mattcen/bugwarrior/commit/31f9fd46802334b097d062413e30ee76d83e599b#diff-650219ece741e89837d808224fc2a019) last time around.

ryneeverett:
Looking at [the numeric class](https://github.com/ralphbean/taskw/blob/develop/taskw/fields/numeric.py) where this error came from, it seems like we should be passing an integer into taskwarrior and getting an integer back. The problem (leading to the incorrect "updating" notices) seems to be that you get a string back even when you pass in an integer. I don't see why the integer coercion in the `deserialize` method wouldn't work though.

mattcen:
Yeah, it now seems like I *am* getting an integer back, probably because earlier I missed setting the UDA to a numeric type at https://github.com/mattcen/bugwarrior/commit/31f9fd46802334b097d062413e30ee76d83e599b#diff-650219ece741e89837d808224fc2a019. Now that's done, the current code (https://github.com/mattcen/bugwarrior/tree/trello-description-and-shortid-uda) seems to all do the right thing.

lukas-reineke:
i had time to test it myself now, the only thing that was wrong was the UDA
i reverted the changes to cast to string and just fixed that. Works fine for me now

ryneeverett (approved):

PR #722, opened 2020-04-11
Fix: SyntaxWarning: "is not" with a literal

jspricke:


ryneeverett (approved):

ryneeverett:
This is probably a good idea because I don't think there's any guarantee short integers will be singletons in the python spec.

jspricke:
Can we get this in and a new release maybe? @ralphbean 
(Debian switched to Python 3.8 where this is is a warning upon install.)

PR #731, opened 2020-05-21
Allow using project IDs instead of names for gitlab

ivan-cukic:


bugwarrior/services/gitlab.py
```diff
@@ -284,7 +284,9 @@ def add_default_namespace(self, repo):
         otherwise, the loggin will be prepended as namespace:
             e.g. "bar" → "<login>/bar"
         """
-        if repo.find('/') < 0:
+        if repo.startswith('id:'):
+            return repo
+        elif repo.find('/') < 0:
```
ryneeverett: Couldn't this be simplified to:

```py
if repo.find('/') < 0 and not repo.startswith('id:'):
    return self.login + '/' + repo
else:
    return repo
```

ryneeverett (commented):
What's the motivation for using id rather than name?

bugwarrior/services/gitlab.py
ivan-cukic: The reason why I decided to separate the conditions is that they belong to different /features/. But I have nothing against merging them.

ivan-cukic:
IDs are more stable than names.

Also, I've encountered some peculiar gitlab setups where name-based project retrieval through the API did not work.

ryneeverett (approved):

ivan-cukic:
Any news on merging this?

Issue #131, opened 2014-05-14
Delete tags before adding new ones


kostajh:
Currently I'm mapping issue labels in the activeCollab service as Taskwarrior tags. Let's say an activeCollab issue comes in with the label `ON_HOLD`. Bugwarrior creates a TW task with this tag. Then the activeCollab issue gets changed to `IN_PROGRESS`. On the next `bugwarrior-pull` we now have a task with `ON_HOLD` and `IN_PROGRESS` tags assigned.

What do you all think about having Bugwarrior delete any prior tags before assigning them on a task update? /cc @coddingtonbear @ralphbean

ralphbean:
Closing this way after the fact.  :)

PR #734, opened 2020-06-01
Add replace_tags option

PtitCaius:
This relate to the old #131.

I added a new option, ``replace_tags`` allowing bugwarrior to delete tags before adding new ones. In my actual workflow, i use labels/tags to filter issues, and keep track of the status of each task.

In gitlab, the state of an issue can be tracked using labels (thanks to the issue board). This option allow me to sync labels (added and deleted ones), in order to have the same labels as the remote issue.

With this option i added ``static_tags``, in order to have a list of labels that bugwarrior will not delete, no matter what.

This option is False by default, and will not change the actual behavior of bugwarrior until checked in.

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -201,6 +201,43 @@ def find_taskwarrior_uuid(tw, keys, issue):
     )
 
 
+def replace_left(field, local_task, remote_issue, keep_items=[]):
+    """ Replace array field from the remote_issue to the local_task
+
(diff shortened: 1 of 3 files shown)


ryneeverett:
I don't have any objections to this feature but I'd like to get some feedback from other maintainers on this one.

ryneeverett (approved):
No other review appears to be forthcoming and I don't want to let this high quality patch rot so I'm going to go ahead and merge.

PR #737, opened 2020-06-06
Test a description beginning with "depends".

ryneeverett:
See https://github.com/ralphbean/bugwarrior/issues/733. I don't foresee this being fixable from bugwarrior but could be worked around in taskw.

I confirmed that this does pass with https://github.com/ralphbean/taskw/pull/135.

Rebased on #913.


ryneeverett:
I rebased on top of #913 and it now passes locally with taskwarrior-2.5.1 and taskw-1.3.1.

PR #739, opened 2020-06-27
Add Review and Owner UDAs for Gerrit

lyarwood:


ryneeverett:
Am I correct to understand that gerritreviews is essentially a space-delimited array of reviews left by various users? If so, what's the reason for omitting "0" reviews?

lyarwood:
> Am I correct to understand that gerritreviews is essentially a space-delimited array of reviews left by various users? If so, what's the reason for omitting "0" reviews?

Yes, I didn't include 0 based reviews as they don't count towards the change itself being approved or denied, at least in the projects I've worked on. Happy to include them if others want to see them listed, I just didn't think they were useful.

Now that I've used this locally I might also revisit the format of the string and stick to something like this:

    -2 (0) | -1 (0) | 0 (0) | +1 (1) | +2 (1)

Thoughts?

ryneeverett:
If possible I'd prefer to avoid creating our own serialized data structures -- they're opinionated and the data is harder to use programmatically. I'm not sure if this is a better idea but what would you think of 5 UDA's for this?

```
REVIEW-2 = 0
REVIEW-1 = 0
REVIEW+0 = 0
REVIEW+1 = 1
REVIEW+2 = 1
```

(I also don't know if those are valid UDA names.)

lyarwood:
Yeah I did think about that approach but had the same naming concerns. I'll use your suggestions and see if anyone objects.

BamBalaam:
@lyarwood 

Hi! Are you planning to revisit this PR? If not, I would gladly attempt to have a go at it.

Do you think it would be interesting to display the reviewer's name in addition to the score?

lyarwood:
On 03-02-21 11:21:37, André Madeira Cortes wrote:
> @lyarwood 
> 
> Hi! Are you planning to revisit this PR? If not, I would gladly attempt to have a go at it.

I'm not planning on working any further on this sorry so please feel
free to move it forward.
 
> Do you think it would be interesting to display the reviewer's name in addition to the score?

Yes that could be useful, thanks!

-- 
Lee Yarwood                 A5D1 9385 88CB 7E5F BE64  6618 BCA6 6E33 F672 2D76

BamBalaam:
@lyarwood ok thanks, I'll try to have a look at it soon-ish

PR #743, opened 2020-07-21
gmail: replace deprecated oauth2client package with google-auth-oauthlib

Fongshway:
## Summary

When following the Gmail setup [instructions](https://bugwarrior.readthedocs.io/en/latest/services/gmail.html), I discovered that [Google's quick start page](https://developers.google.com/gmail/api/quickstart/python) relies on `google-auth-oauthlib` instead of the deprecated `oauth2client` package, which `bugwarrior` currently uses. Since my credentials had expired, I also discovered that the new `credentials.json` created by following the quick start guide is also incompatible with the old credential file I was using. While I'm no OAuth expert, I used the Python code on the quick start page to make updates to `bugwarrior`'s Gmail service so we can use the newer `google-auth-oauthlib` package. Hopefully this saves someone else some trouble! See [oauth2client deprecation](https://google-auth.readthedocs.io/en/latest/oauth2client-deprecation.html) for more info.

Some other clean up items:

- Add `gmail` as supported service in `README` and sort list alphabetically.
- Sort `install_requires` and `extras_require` alphabetically in `setup.py`.
- Also update `extras_require` to use `{}` instead of `dict()`, since the former plays nicer with [pipenv](https://github.com/pypa/pipenv), which I use for `bugwarrior` development. See https://github.com/pypa/pipenv/issues/4309 for the relevant `pipenv` issue. If there's interest, I can open a separate PR with the `Pipfile` and `Pipfile.lock` I use on my fork.


ryneeverett (approved):

ryneeverett:
> If there's interest, I can open a separate PR with the Pipfile and Pipfile.lock I use on my fork.

I'd be especially supportive of adding these files if they were accompanied by an additional section in the [contributing](https://bugwarrior.readthedocs.io/en/latest/contributing.html) docs describing how to use them as an alternative method of setting up a development environment.

tests/test_gmail.py
```diff
@@ -1,59 +1,100 @@
...
+        gmail.GmailService.build_api = mock_api
+        service = GmailService(self.config, "general", "myservice")
+
+        expected = Credentials(**copy(CREDENTIAL_DATA))
+        with open(service.credentials_path, "wb") as token:
+            pickle.dump(expected, token)
+
+        self.assertEqual(service.get_credentials().to_json(), expected.to_json())
+
+    def test_get_credentials_expired(self):
```
ryneeverett: Could you add some comment(s) explaining how this test works? It isn't obvious why mocking `Credentials.refresh` to return `None` would yield the same results as `expected.__dict__["expiry"] = datetime.now()`.

ryneeverett (approved):

tests/test_gmail.py
Fongshway: I'm new to mocking and agree my initial test wasn't very clear, but after sleeping on it I think I have a cleaner/more self explanatory solution for this via 7e4c263. Let me know what you think!

ryneeverett (approved):

tests/test_gmail.py
ryneeverett: Looks good, thanks!

Fongshway:
@ryneeverett can this be merged? I was planning on opening a separate PR for the `Pipfile` afterwards.

ryneeverett:
> @ryneeverett can this be merged? I was planning on opening a separate PR for the `Pipfile` afterwards.

I do plan to merge this but I usually wait a few weeks to give others the opportunity to review. Reviews from others are rare these days but I like to leave a window in hopes that this will change. There isn't much risk of merge conflicts at the current rate of development.

Please do submit the Pipfile additions in a separate PR.

Fongshway:
Sounds good, thanks!

PR #666, opened 2019-05-23
Add support for Pivotal Tracker service

amdemas:
This resolves #409

PR #745, opened 2020-07-29
[WIP] Azure Devops 

cam-barts:
## Azure Devops Integration 
This integrates [Azure Devops](https://azure.microsoft.com/en-us/services/devops/) workitems into Bugwarrior.

So far the service is functional, but there's still some ironing out to do. 

### The general workflow:
- Get workitem ID's from a [wiql](https://docs.microsoft.com/en-us/azure/devops/boards/queries/wiql-syntax?view=azure-devops) query -> list(int)
- For workitem in query, create a taskwarrior issue

### The default query:
```sql
SELECT [System.Id] FROM workitems WHERE [System.AssignedTo] = @me
```
There is a config item where you can add filters to the query. 
For Example: 
```yaml
 ado.wiql_filter = [System.State] not in ('Done', 'Complete') AND [System.WorkItemType] in ('Task', 'Bug')
```

### Framework mods:
- ADO gives options for many work item types that aren't in the standard `cls_markup`
- I use pipenv, so I ignored the Pipfile and Pipfile.lock

### Known Issues:
- With the default query, at least with my actual work, is terribly slow, as it get all previous closed tickets every pull
   - Pulling 185 items takes around a minute to a minute 15 seconds
- It has no logic to get multiple projects/orgs with the same key, so each project/org pair would need a new config target

### Work Still to Be Done
- [ ] Tests
    - [ ] I need to learn nose
- [ ] Documentation
    - [ ] I've not written any docs in restructured text, so I'll have to figure that out too


.gitignore
```diff
@@ -18,3 +18,4 @@ man/*
 *.idea
 .tox/*
 *.bak
+Pipfile*
```
ryneeverett, line 21: FYI, there's been discussion in #743 of adding Pipfile to the git repository as a tracked file.

bugwarrior/services/__init__.py
```diff
@@ -368,6 +368,8 @@ def build_default_description(
             'release': 'Release',
             'chore': 'Chore',
         }
+        if cls not in cls_markup.keys():
+            cls_markup[cls] = cls.lower()
```
ryneeverett, line 372: It's not clear to me why this is necessary since `task` and `bug` are both keys in `cls_markup`.

cam-barts: Task and Bug aren't the only workitem types that can be assigned to users in Azure Devops. It supports different methodologies, which call tracked tasks different things that aren't included in the dict, like `product backlog item` and `impediment` for scrum process and `change request` and `risk` for CMMI.

ryneeverett: In the past we've added items to `cls_markup` for each individual type, even if it's only used by a single service. (See #666 in which we added 5 items for one service.) It would probably be better to move this logic down into the individual services but I don't think we'll want to accept open-ended types like this.

cam-barts: Perhaps I don't understand this portion of code like I thought I did. I was under the impression that all this provides is a work item type in the default description, and gives it some shorthand or removed it in special cases. I'd be fine doing it either way, and even trying to abstract this into the config somehow, but I don't understand why open ended types would be an issue.

ryneeverett: You're right. In general we prefer that bugwarrior breaks when our API assumptions are violated but in this case there don't appear to be consequences for arbitrary classes.

However, for consistency could we make this  `cls.title()`? You could then remove the last 5 items in `cls_markup` which would be effectively no-op.

Also, I think it would be more straightforward to do `cls_markup.get(cls, cls.title())` in the returned string below than to add it to the dict like this.

silvio:
Can I ask what the status of this PR is? This looks like I could make good use of this? Is it still being worked on, @cam-barts ?

cam-barts:
Hiya @silvio, I had to put this down for awhile to work on other things, but I can put some effort into it by next week. I need to learn nose to get the cov back up to snuff and I need to address some of the stuff @ryneeverett mentioned, but the functionality is there on my fork and I've been using it for months. Let me know if that works for you.

ryneeverett:
> I need to learn nose to get the cov back up to snuff

FYI we switched from nose to pytest since you originally posted this PR.

cam-barts:
It seems I've screwed up my git stuff. Working on cleaning up and then getting some tests in.

PR #747, opened 2020-08-17
bugzilla: Treat api_key as a password

pierreprinetti:
This patch enables using a vault or `@oracle` to retrieve a Bugzilla api
key.

Example:

```toml
bugzilla.api_key = @oracle:eval:pass bugzilla/bugwarrior
```

Note that I have only tested `@oracle eval` locally.

### DIFF
--- a/bugwarrior/services/bz.py
+++ b/bugwarrior/services/bz.py
@@ -150,8 +150,8 @@ class BugzillaService(IssueService):
             force_rest_kwargs = {"force_rest": True}
 
         url = 'https://%s' % self.base_uri
-        api_key = self.config.get('api_key', default=None)
-        if api_key:
+        if self.config.get('api_key'):
+            api_key = self.get_password('api_key')
             try:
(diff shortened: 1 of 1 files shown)


bugwarrior/services/bz.py
```diff
@@ -150,9 +150,9 @@ def __init__(self, *args, **kw):
             force_rest_kwargs = {"force_rest": True}
 
         url = 'https://%s' % self.base_uri
-        api_key = self.config.get('api_key', default=None)
-        if api_key:
+        if self.config.get('api_key'):
             try:
+                api_key = self.get_password('api_key')
```
ryneeverett: Is there a reason this is going into the `try` block?

pierreprinetti: No more than the next call to `get_password`, indeed. Good catch, thanks!

ryneeverett (approved):

PR #749, opened 2020-09-09
Use numeric type for youtrack issues

garacio:


ryneeverett:
What's the motivation?

The services are fairly mixed on this point right now. For a little context see: #353 #557.

ryneeverett:
We can reopen this if there's sufficient interest.

PR #754, opened 2020-09-19
Jira fixes

jspricke:


bugwarrior/services/jira.py
```diff
@@ -188,8 +188,11 @@ def __get_sprints(self):
             for key in self.origin['sprint_field_names']
         ], [])
         for sprint in sprints:
-            # Parse this big ugly string.
-            yield _parse_sprint_string(sprint)
+            if isinstance(sprint, dict):
+                yield sprint
+            else:
```
ryneeverett, line 193: I'd like to add a comment here like `else:  # Old version of python-jira without sprint support.`

ryneeverett (approved):

bugwarrior/services/jira.py
jspricke: This is actually not a new feature of python-jira but Atlassian seems to have fixed it in Jira itself. I left the old code as a backup as I don't know if it was fixed everywhere.

ryneeverett: I see. I still think some comment indicating that the else-clause is for backwards compatibility would be nice.

jspricke: Agreed, I added a comment.

ryneeverett (approved):

bugwarrior/services/jira.py
```diff
@@ -159,7 +159,7 @@ def get_due(self):
             return self.parse_date(self.record['fields']['duedate'])
         # Otherwise, if the issue is in a sprint, use the end date of that sprint.
         sprints = self.__get_sprints()
-        for sprint in filter(lambda e: e.get('state') != 'CLOSED', sprints):
+        for sprint in filter(lambda e: e.get('state', '').lower() != 'closed', sprints):
```
StephanMeijer, line 162: Can we make a constant of `closed`?

StephanMeijer (changes requested):

bugwarrior/services/jira.py
jspricke: The string is generated by Jira and exported through it's API. python-jira has nothing about it and this is the only place it is used in the bugwarrior jira.py. Why would you want to make it a constant?

PR #756, opened 2020-09-22
[phab] Use diff variable iterated over

jspricke:
This was reported to Debian in https://bugs.debian.org/886383.

I didn't test this but given that the code for diffs seems to be copy
and paste from issues, I assume this is currect.


jspricke:
Oups, my branch was old and this was already fixed in develop.
Btw. can you please do a new release?

Issue #760, opened 2020-10-06
Add property `gmail.project_name` to assign project to imported tickets


StephanMeijer:
(Feature proposal)

ryneeverett:
Closing per #781.

PR #762, opened 2020-10-06
Issue #760 add property gmail.project_name

StephanMeijer:


StephanMeijer:
This seems unnecessary due to the availability of `default_project`. Rewriting this PR in another PR into basic tests.

bugwarrior/docs/services/gmail.rst
```diff
@@ -33,6 +33,7 @@ Here's an example of a gmail target:
     service = gmail
     gmail.query = label:action OR label:readme
     gmail.login_name = you@example.com
+    gmail.project_name = example_gmail
```
ryneeverett: I know the other services with this feature document it like this but I think it would be better to add a separate section in the docs about it. It seems like we've moved towards trying to keep the initial "Example Service" section as a minimal usage example or at least just the most common options.

ryneeverett (changes requested):

bugwarrior/docs/services/gmail.rst
StephanMeijer: Ok sure, will make the changes.

StephanMeijer:
@ryneeverett Changed!

ryneeverett:
So long as project_name is not in "Common Service Configuration Options" implemented across all services (which it probably should be) it seems like it ought to be documented somewhere in the gmail docs, no?

StephanMeijer:
Where would you document it, if not in the **Example Service**?

ryneeverett:
I'd add a `Service Features` section and a subsection like [in versionone](https://bugwarrior.readthedocs.io/en/latest/services/versionone.html?#set-a-global-project-name). I'd be fine with you just copy/pasting that if you so choose.

StephanMeijer:
@ryneeverett Documentation like I did in #777?

StephanMeijer:
This can be closed, see #778 / #776

PR #767, opened 2020-10-28
Jira: use paginated query to get ALL results

### LINKED ISSUE #764: jira: not paginating?
- Symptom: when I run bugwarrior-pull, old still-open issues randomly get closed.

- `len(self.jira.search_issues(…))` is  returning 100, while the same query run manually returns 365 items. (yeah,, yeah, … my backlog )

I suspect pagination isn't happening somehow.

srl295:
i haven't read the docs, but from reading the code it seems that `None` is the right thing to pass.  `-1` seems to
get 100 results, while omitting the parameter returns the default of `100`.

Fixes: #764

### DIFF
--- a/bugwarrior/services/jira.py
+++ b/bugwarrior/services/jira.py
@@ -358,7 +358,7 @@ class JiraService(IssueService):
         )
 
     def issues(self):
-        cases = self.jira.search_issues(self.query, maxResults=-1)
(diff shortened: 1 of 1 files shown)


ryneeverett (approved):
Makes sense to me looking at the source as well. Pagination only works if maxResults is falsey and (surprising to me) `bool(-1) is True`.

PR #772, opened 2020-11-10
bugzilla: Implement common filters.

### LINKED ISSUE #243: Please implement 'only_if_assigned' option for Bugzilla plugin
Currently it is not possible to filer bugs from Bugzilla based on the assignee. The result is that also bugs not assigned to me, but reported by me are added to my tasks in taskwarrior. It would be nice to be able to filter only bugs assigned to me.

ryneeverett:
This adds support for the common configuration options
`only_if_assigned` and `also_unassigned`, which should be supported by
all services.

Resolve #243 

<s>**Note**: I haven't tested this, I just pieced it together based on [the docs](https://bugzilla.readthedocs.io/en/latest/api/core/v1/bug.html).</s> (see comment below)

### DIFF
--- a/bugwarrior/docs/services/bugzilla.rst
+++ b/bugwarrior/docs/services/bugzilla.rst
@@ -44,15 +44,10 @@ still required in this case, in order to identify bugs belonging to you.
 The above example is the minimum required to import issues from
 Bugzilla.  You can also feel free to use any of the
 configuration options described in :ref:`common_configuration_options`.
-Note, however, that the filtering options, including ``only_if_assigned``
-and ``also_unassigned``, do not work
 
-There is an option to ignore bugs that you are only cc'd on::
+There is also an option to ignore bugs that you are only cc'd on::
 
     bugzilla.ignore_cc = True
-
-But this will continue to include bugs that you reported, regardless of
-whether they are assigned to you.
 
 If your bugzilla "actionable" bugs only include ON_QA, FAILS_QA, PASSES_QA, and POST::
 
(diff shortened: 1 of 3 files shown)


ryneeverett:
@dthadi3 Any idea why we're only getting this error on power9? I take it the power9 image is just out of date?

```
File "/home/travis/virtualenv/python3.7.2/lib/python3.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 183, in _verify_openssl_version

    "You are linking against OpenSSL 1.0.2, which is no longer "

RuntimeError: You are linking against OpenSSL 1.0.2, which is no longer supported by the OpenSSL project. To use this version of cryptography you need to upgrade to a newer version of OpenSSL. For this version only you can also set the environment variable CRYPTOGRAPHY_ALLOW_OPENSSL_102 to allow OpenSSL 1.0.2.
```

svghadi:
@ryneeverett Your assumption is true. The `xenial` distribution for power comes with `openssl 1.0.2`. If we require higher version, we need to switch to `bionic` which has `openssl 1.1.1`. To resolve this failure, we can either update to `bionic` or set the `CRYPTOGRAPHY_ALLOW_OPENSSL_102` env variable as suggested. @dthadi3

ryneeverett:
Tested with the following configuration (rebased on #787):

```ini
[general]
targets = my_bugzilla

[my_bugzilla]
service = bugzilla
bugzilla.base_uri = bugzilla.redhat.com
bugzilla.username = rbean@redhat.com
bugzilla.only_if_assigned = Ralph Bean
```

PR #774, opened 2020-11-11
travis -> github actions

ryneeverett:
I'm not sure how advantageous this is at the present but this move may put us in a better position in the future as the community continues to migrate away from travis.


ryneeverett:
I'm guessing @ralphbean would need to enable actions in settings for this to work. The results can be seen at ryneeverett/bugwarrior#3.

ryneeverett:
@coddingtonbear made the switch to github-actions in taskw and they run, though they don't appear on the PR page. Presumably the same thing would happen here if we merged this without changing administrative settings -- CI would run and we could see the results in the actions tab but not on the PR's. Or perhaps there's just a bug in which the github-actions results don't appear on existing PR's? (I'm drawing these conclusions based on force-pushing to a PR I have open in taskw, so time will tell if this issue persists with new PR's.)

coddingtonbear:
I believe actions run only when you tell them to, and in the case of most linting/testing kinds of actions, they're executed on "push" (even here: https://github.com/ralphbean/bugwarrior/pull/774/files#diff-455560f34eb048659ff4abf3a46016903d1f8616fa681d2482b76be2472e257bR3).  Unfortunately, that does mean you might need to push up an empty commit for causing github to trigger a re-build; luckily there's always `--allow-empty` on the command-line.

ryneeverett:
Yeah, I was force pushing when rebasing my PR on taskw and I received emails about action success/failure but nothing shows up in the UI.

coddingtonbear:
Have you tried just pushing up an empty commit?  (Also: I'm @coddingtonbear:matrix.org
on matrix or coddingtonbear on freenode if you want to chat)

ryneeverett:
I pushed an empty commit and still no github actions on the PR page. Same for the new PR https://github.com/ralphbean/taskw/pull/138.

coddingtonbear:
Oh, you're right!  It looks like you have to also trigger on `pull_request`; here's an example I found:

```yaml
on:
  push:
    branches:
      - master
      - main
  pull_request:
    paths:
      - "**.css"
      - "**.js"
      - "**.json"
      - "**.py"
      - "**.scss"
      - "**.snap"
      - "**.ts"
      - "**.tsx"
```

I doubt the paths or branch lists are necessary, fwiw.

ryneeverett:
@coddingtonbear Then why did the checks show up in https://github.com/ryneeverett/bugwarrior/pull/3? My guess is that it's because I enabled Actions in settings.

coddingtonbear:
Oh, hrm; that's a good question.  That sounds totally plausible -- I even see this setting there in a repo I own:

> -  [ ] Run workflows from fork pull requests
> 
> This tells Actions to run workflows from pull requests originating from repository forks. Note that doing so will give maintainers of those forks the ability to use tokens with read permissions on the source repository.

ryneeverett:
Travis seems to be working better these days.

coddingtonbear:
Just by the way, @ryneeverett, I _believe_ our time with Travis.ci will run out eventually: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing#building-on-a-public-repositories-only

ryneeverett:
Ah thanks, I hadn't read that article closely enough when it first came out. I didn't realize free open source CI was actually going away. :exploding_head:

Issue #776, opened 2020-11-13
Handle `project_name` as a first class property: just like `default_priority`


StephanMeijer:
This'll enable us to use `default_project` globally, just like we can use `default_priority` globally.

ryneeverett:
To be clear, by "first class property" and "globally" we're referring to the fact that you can use it across all services, correct?

Why can't we stick with `project_name` for consistency and backwards compatibility?

StephanMeijer:
We sure can do that.

ryneeverett:
I'd forgotten that [field templates](https://bugwarrior.readthedocs.io/en/latest/common_configuration.html#field-templates) can be used to achieve the same effect. (`SERVICE.project_template = Office`) Would it make more sense to deprecate the existing `project_name` options?

StephanMeijer:
Yep.

PR #778, opened 2020-11-14
#776 default_project property

StephanMeijer:
Maybe rename attributes `project_name` also to `default_project` or make them synonymous?


bugwarrior/services/__init__.py
```diff
@@ -62,7 +62,8 @@ def __init__(self, main_config, main_section, target):
         self.annotation_newlines = self._get_config_or_default('annotation_newlines', False, asbool)
         self.shorten = self._get_config_or_default('shorten', False, asbool)
 
-        self.default_priority = self.config.get('default_priority', 'M')
+        self.default_priority = self.config.get('default_priority')
```
StephanMeijer: It is weird to force the priority `M` when I'd think: when no `default_priority` is given, no default priority should be assigned.

bugwarrior/services/__init__.py
```diff
@@ -165,18 +167,14 @@ def build_annotations(self, annotations, url):
     @classmethod
     def validate_config(cls, service_config, target):
```
StephanMeijer, line 168: Shortened this function just a bit

StephanMeijer:
Not using because of all templating functionality

PR #782, opened 2020-11-29
Add `github.body_length` configuration

### LINKED ISSUE #766: Support not syncing description
I'm copied on repos that see a lot of large issues with lots of comments.  This generates a huge quantity of updates, each of which includes its own copy of all of that traffic, ballooning the task data that must be sync'd.  And, I don't need any of that -- I just need a link to the issue.

Separately, [certain characters in the subject of an issue will tickle](https://github.com/ralphbean/bugwarrior/issues/527) https://github.com/GothenburgBitFactory/taskwarrior/issues/2140 which means that *every* bugwarrior run "modifies" the task.

It'd be great to have a way to disable the sync of comments to annotations.

djmitche:
This controls the maximum length of the githubbody UDA, and can be used
to avoid syncing the body at all (by setting it to 0).

Fixes #766.

### DIFF
--- a/bugwarrior/docs/services/github.rst
+++ b/bugwarrior/docs/services/github.rst
@@ -151,6 +151,17 @@ point bugwarrior to it with the ``github.host`` configuration option. E.g.::
 
     github.host = github.acme.biz
 
+Synchronizing Issue Content
++++++++++++++++++++++++++++
+
+This service synchronizes most GitHub fields to UDAs, as described below.
+Comments are synchronized as annotations.
+
+To limit the amount of content synchronized into TaskWarrior (which can help to avoid issues with synchronization), use
+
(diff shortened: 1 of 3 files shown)


ryneeverett (approved):

ryneeverett:
The reason I suggested `description_length` rather than `body_length` was for consistency with our existing configuration options (such as `annotation_length`) which reference fields in terms of the field names in taskwarrior rather than in their respective services. What do you think?

djmitche:
Description isn't task field, though.  I was keying off the name of the
uda.  I can change it if you'd like, though.

djmitche:
Oh, also, `github.description_length` applies to the *task* description, rather than the issue description, so I think re-using the word `description` in this context would be super-confusing.

ryneeverett:
Nevermind, it was I that was confused. This makes sense.

PR #796, opened 2020-12-29
find UUID in completed only after pending/waiting

jspricke:
9fe6215 added completed to the search to allow reopening tasks but the
issue could have been completed multiple times before resulting in an
MultipleMatches exception. This change only considers completed tasks if
the issue is not in the pending/waiting set and also arbitrarily returns
the first found result.

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -179,7 +179,6 @@ def find_taskwarrior_uuid(tw, keys, issue):
                 'or': [
                     ('status', 'pending'),
                     ('status', 'waiting'),
-                    ('status', 'completed')
                 ],
             })
             possibilities = possibilities | set([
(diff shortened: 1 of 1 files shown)


bugwarrior/db.py
```diff
@@ -197,6 +196,24 @@ def find_taskwarrior_uuid(tw, keys, issue):
...
 
+    # old bugwarrior did not consider completed task so there could be multiple ones matching
+    for service, key_list in six.iteritems(keys):
+        if any([key in issue for key in key_list]):
+            results = tw.filter_tasks({
+                'and': [("%s.is" % key, issue[key]) for key in key_list],
+                'or': [
+                    ('status', 'pending'),
+                    ('status', 'waiting'),
+                    ('status', 'completed')
```
ryneeverett: Shouldn't we only be checking `completed` issues here? If it were `pending` or `waiting` it would have already been returned.

ryneeverett (approved):
Thanks for reporting this oversight!

bugwarrior/db.py
```diff
@@ -197,6 +196,19 @@ def find_taskwarrior_uuid(tw, keys, issue):
             )
         )
 
+    # old bugwarrior did not consider completed task so there could be multiple ones matching
+    for service, key_list in six.iteritems(keys):
+        if any([key in issue for key in key_list]):
+            results = tw.filter_tasks({
+                'and': [[("%s.is" % key, issue[key]) for key in key_list], ('status', 'completed')]
```
ryneeverett: Looks like these lists need to be concatenated, so something like:

```py
               'and': [("%s.is" % key, issue[key]) for key in key_list] + ['status', 'completed'],
```

ryneeverett:
While I find your PR more readable I'm leaning towards #797 for performance reasons. If the extra taskwarrior query only occurred on edge cases I would consider that acceptable but it looks like as it current stands there will be an extra query for every new issue before adding it to the database. Thoughts?

bugwarrior/db.py
jspricke: looks better, thx. pushed.

jspricke:
I agree that #797 is probably the better fix.

PR #797, opened 2020-12-31
Fix duplicate completed task exception.

ryneeverett:
I was trying for a more readable alternative to #796 but I'm not sure I
was successful.


ryneeverett:
I'd say this loses on readability but wins on performance by avoiding an extra taskwarrior query (which is ultimately an additional subprocess).

ryneeverett:
Another tactic would be to do deduplication and store the status of whether this has been done in the `BugwarriorData` file. However this would be less conservative and inappropriate for the point release I'm hoping to push once this is fixed.

PR #800, opened 2021-02-02
Add missing UDA Fields in the Docs

BamBalaam:
I was playing around with Bugwarrior for JIRA tickets and was confused to see on the documentation no UDA field for the JIRA ticket status.

Wondering if it was easy to introduce and publish a PR, I delved in the code. That's where I saw that the UDA field actually existed, but was not documented.

This PR fixes that issue for all services where I found fields missing.


ryneeverett:
Thanks!

PR #816, opened 2021-05-03
Gitlab individual priorities

fmauch:
Allow configuring individual priorities for issues, todos and MRs

Gitlab doesn't offer any native priority definition. In the current implementation, issues are prioritized with the default priority, while all other types are hardcoded to have priority 'H'.

This PR lets users configure individual priorities for each, issues, todos and merge requests.

This implements #814

---
## ToDo's left
 - [x] move config parsing to `__init__` (https://github.com/ralphbean/bugwarrior/pull/816#discussion_r627878020)
 - [x] rename parameters to `default_[whatever]` (https://github.com/ralphbean/bugwarrior/pull/816#discussion_r627878596)
 - [x] Add tests
 - [ ] Add to changelog

PR #818, opened 2021-05-09
[Gitlab] Use due_date from target for todos

fmauch:
As gitlab todos don't have their own due dates, but the targets they are created for (e.g. issues, MRs) might have a due date, I think it does make sense to add this due date to the task.

I'll create this as WIP, as I'd like to also add tests for that, using the todo test introduced in #816


ryneeverett:
Closing as stale -- merge conflicts and never was un-drafted.

PR #826, opened 2021-06-22
feat(services.jira): include jiraassignee as new uda field

jakebrinkmann:
I really wanted this, so I thought I would share and hope someone might think it's a good idea or something 👍

### DIFF
--- a/bugwarrior/services/jira.py
+++ b/bugwarrior/services/jira.py
@@ -69,6 +69,8 @@ class JiraIssue(Issue):
     SUBTASKS = 'jirasubtasks'
     EPIC = 'jiraparent'
     POINTS = 'estimate'
+    ASSIGNEE = 'jiraassignee'
+    REPORTER = 'jirareporter'
 
     UDAS = {
         EPIC: {
@@ -297,11 +299,13 @@ class JiraIssue(Issue):
 
     def get_assignee(self):
         # email address "better" than accountId (hex) or displayName (with spaces)
-        return self.record['fields']['assignee']['emailAddress']
+        record = self.record['fields']['assignee']
+        return record.get('emailAddress') or record.get('displayName')
 
     def get_reporter(self):
         # email address "better" than accountId (hex) or displayName (with spaces)
-        return self.record['fields']['reporter']['emailAddress']
+        record = self.record['fields']['reporter']
+        return record.get('emailAddress') or record.get('displayName')
 
 
 class JiraService(IssueService):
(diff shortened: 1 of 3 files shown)


bugwarrior/services/jira.py
```diff
@@ -269,6 +275,9 @@ def get_subtasks(self):
     def get_issue_type(self):
         return self.record['fields']['issuetype']['name']
 
+    def get_assignee(self):
+        return self.record['fields']['assignee']['emailAddress']
```
ryneeverett: I think we would normally use a username here rather than an email address.

tests/test_jira.py
```diff
@@ -44,6 +44,7 @@ class TestJiraIssue(AbstractServiceTest, ServiceTest):
             'issuetype': {'name': 'Epic'},
             'status': {'name': 'Open'},
             'subtasks': [{'key': 'DONUT-%s' % subtask} for subtask in arbitrary_subtask_ids]
```
ryneeverett: Missing a comma at the end of this line.

bugwarrior/services/jira.py
jakebrinkmann: I agree, a simple user name wasn't available:

```
 'assignee': {'accountId': '5e18fbf9d52...88579',
              'accountType': 'atlassian',
              'active': True,
              'avatarUrls': {'16x16': 'https://avatar-...-48a3-843a-29184114a6e5/16',},
              'displayName': 'Jake Brinkmann',
              'emailAddress': 'jb...@....com',
              'self': 'https://sonifi.atlassian.net/rest/api/2/user?accountId=5e18fbf9d52...88579',
              'timeZone': 'America/Chicago'},
```

maybe I will make two UDAs, one for email and one for display name. 

the same data structure also exists for `'creator'`, `'reporter'` if I should consider bringing those in as UDAs as well

tests/test_jira.py
jakebrinkmann: do'h 🤦 I'll see if I can get the test suite running locally before pushing again

bugwarrior/services/jira.py
ryneeverett: Ah, that's an unfortunate constraint but now I understand your choice.

ryneeverett (approved):

ryneeverett:
Looks like the tests need a bit more work.

ryneeverett:
Closing as stale -- merge conflicts and doesn't look like it was ever completely done.

Issue #304, opened 2016-04-18
Creates shell argument lists that are too long


ryansb:
[If this error would be better handled in `taskw` I can refile it there]

When running against the [ansible/ansible](https://github.com/ansible/ansible) repo, I had bugwarrior crash because it tries to send too many arguments to taskwarrior at once. 

```
CRITICAL:command|oh noes
TRACE Traceback (most recent call last):
TRACE   File "/home/ryansb/code/bugwarrior/bugwarrior/command.py", line 55, in pull
TRACE     synchronize(issue_generator, config, main_section, dry_run)
TRACE   File "/home/ryansb/code/bugwarrior/bugwarrior/db.py", line 387, in synchronize
TRACE     tw.task_add(**issue)
TRACE   File "/home/ryansb/.pyenv/versions/2.7.10/envs/bugwarrior/lib/python2.7/site-packages/taskw/warrior.py", line 666, in task_add
TRACE     stdout, stderr = self._execute('add', *args)
TRACE   File "/home/ryansb/.pyenv/versions/2.7.10/envs/bugwarrior/lib/python2.7/site-packages/taskw/warrior.py", line 462, in _execute
TRACE     stderr=subprocess.PIPE,
TRACE   File "/home/ryansb/.pyenv/versions/2.7.10/lib/python2.7/subprocess.py", line 710, in __init__
TRACE     errread, errwrite)
TRACE   File "/home/ryansb/.pyenv/versions/2.7.10/lib/python2.7/subprocess.py", line 1335, in _execute_child
TRACE     raise child_exception
TRACE OSError: [Errno 7] Argument list too long
```

coddingtonbear:
Could you check to see what the results of `xargs --show-limits` are on your system, @ryansb; it *sounds* like you might be running somewhere where that value isn't as high as it usually is (around 2mb).

ryansb:
```
$ xargs --show-limits
Your environment variables take up 4275 bytes
POSIX upper limit on argument length (this system): 2090829
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 2086554
Size of command buffer we are actually using: 131072
```

ryansb:
Oh goodness. Here's the issue it seems to be choking on. https://github.com/ansible/ansible/issues/15416 

That would explain why it hits the length limit.

ralphbean:
Are you for real real?  How about we apply some truncation (at the bugwarrior level, or at taskw's?)

[...]

PR #837, opened 2021-09-05
Update docs to suggest using static_fields to suppress descriptions

### LINKED ISSUE #835: Allow truncating / eliminating trello descriptions, as with github and jira
Yep, I have to use Trello *and* Jira.  I am truly blessed.

At least with this change, I can avoid being blessed with constantly-changing huge Taskwarrior tasks :)

djmitche:
This is simpler than setting `body_length=0`, and works for all
services.

Fixes #835

### DIFF
--- a/bugwarrior/docs/services/github.rst
+++ b/bugwarrior/docs/services/github.rst
@@ -160,7 +160,7 @@ Comments are synchronized as annotations.
 To limit the amount of content synchronized into TaskWarrior (which can help to avoid issues with synchronization), use
 
  * ``annotation_comments=False`` (a global configuration) to disable synchronizing comments to annotations; and
- * either ``github.body_length=0``` to disable synchronizing the Github Body UDA (or set it to a small value to limit size) or include ``githubbody`` in ``static_fields`` in the ``[general]`` section to eliminate the UDA entirely.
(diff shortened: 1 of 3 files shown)


bugwarrior/docs/services/github.rst
```diff
@@ -160,7 +160,7 @@ Comments are synchronized as annotations.
 To limit the amount of content synchronized into TaskWarrior (which can help to avoid issues with synchronization), use
 
  * ``annotation_comments=False`` (a global configuration) to disable synchronizing comments to annotations; and
- * ``github.body_length=0``` to disable synchronizing the Github Body UDA (or set it to a small value to limit size).
```
ryneeverett, line 163: I don't think we should suggest multiple ways to attain the same ends with no explanation of the benefits of making one choice over the other.

The advantage of `body_length` is that you might want to limit the length to a non-zero value. (For example to fix #304.)

So maybe something more like

    * either set ``github.body_length`` to limit size or include ``githubbody`` in ``static_fields`` in the ``[general]`` section to eliminate the UDA entirely

ryneeverett: To be clear: I'm suggesting that using static_fields is always a better solution than body_length=0. Perhaps we should even raise a warning to that effect in the validation function.

djmitche: Makes sense!

ryneeverett (approved):

PR #842, opened 2021-10-18
Use a bugwarrior mirror to generate documentation.

### LINKED ISSUE #812: Update readthedocs.
The current commit is a few years old (5f296a82) and doesn't correspond with a release so maybe this is configured in the readthedocs or github UI?

ryneeverett:
Fix #812.

See <https://readthedocs.org/projects/bugwarrior/builds/>. I don't have
the permissions needed to set up the webhook integration currently used,
so I created a mirror of the repository at
<https://github.com/ryneeverett/bugwarrior-docs> and a corresponding
documentation site at <https://bugwarrior-docs.readthedocs.io>.

### DIFF
--- a/bugwarrior/README.rst
+++ b/bugwarrior/README.rst
@@ -32,8 +32,8 @@ It currently supports the following remote resources:
 Documentation
 -------------
 
-For information on how to install and use bugwarrior, read `the docs
-<https://bugwarrior.readthedocs.io>`_ on RTFD.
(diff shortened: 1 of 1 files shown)


ryneeverett:
Note that the documentation as set up currently is built on the latest `develop` branch. Ideally it would be built on release branches but until we have an actual release process it would be hard to remember to create such branches.

PR #844, opened 2021-10-29
Drop Megaplan Service

ryneeverett:
We dropped python2 support a couple years ago and the `megaplan` library
still doesn't support python3. The library's homepage listed on pypi is
a dead link.


ryneeverett:
cc @umonkey

srl295 (approved):
FWIW LGTM

PR #849, opened 2021-11-08
Pydantic

ryneeverett:
See commit messages for motivation, etc.


djmitche (approved):
I did not examine each line carefully, but I like this refactor!

Are there any particular guidelines to doc for using pydantic in per-service configs, or is it pretty much by-the-book?

ryneeverett:
Thanks for having a look! I didn't expect anyone to do a thorough line-by-line review but it's nice to have a sanity check on such a large change before I beta test it on develop users.

Good question about the service schemas. The pydantic usage *is* by-the-book but the potential gotcha is that the schemas receive data that is still serialized from configparser. The only practical implication of this so far is that lists need need to be annotated with the custom field I've called `ConfigList` rather than simply `list`. Also the prefix used in configparser (e.g. `github` in `github.login`) needs to be passed to the constructor.

Since there is currently no documentation on contributing a service I'd prefer to leave that for a separate PR and have opened #855 to track it.

djmitche:
Sounds good!

PR #853, opened 2021-11-25
Support Jira Personal Access Tokens (PATs)

twaugh:
This adds a `jira.personal_access_token` configuration field.

Signed-off-by: Tim Waugh <twaugh@redhat.com>

### DIFF
--- a/bugwarrior/services/jira.py
+++ b/bugwarrior/services/jira.py
@@ -283,8 +283,6 @@ class JiraService(IssueService):
         super(JiraService, self).__init__(*args, **kw)
         self.username = self.config.get('username')
         self.url = self.config.get('base_uri')
-        password = self.get_password('password', self.username)
-
         default_query = 'assignee="' + \
             self.username.replace("@", "\\u0040") + \
             '" AND resolution is null'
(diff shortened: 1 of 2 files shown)


ryneeverett (changes requested):
For consistency with our existing services the token field should be a string containing the token rather than a boolean indicating the password is in fact a token. (Note that the validation would need to be updated to make the password OR token required.)

Also, you might consider naming the field PAT for consistency with Azure, but I'm not that opinionated.

twaugh:
Thanks, re-worked.

bugwarrior/services/jira.py
```diff
@@ -347,10 +349,13 @@ def get_service_metadata(self):
 
     @classmethod
     def validate_config(cls, service_config, target):
-        for option in ('username', 'password', 'base_uri'):
+        for option in ('username', 'base_uri'):
             if option not in service_config:
                 die("[%s] has no 'jira.%s'" % (target, option))
 
+        if 'password' in service_config and 'PAT' in service_config:
```
ryneeverett: An exception should be raised if neither option is set.

ryneeverett (changes requested):

bugwarrior/services/jira.py
twaugh: Thanks, updated.

ryneeverett: I'm not sure your updated logic works the way you think it does. It will catch if both are empty or if both are set to the same string, but not if both are set to different strings. I don't think there's a way to do what you want with a single logical operator. You could do:

```py
if ((not 'password' in service_config or 'PAT' in service_config) or
    ('password' in service_config and 'PAT' in service_config)):
```

twaugh: It does also catch the case where they are set to different strings... each `('...' in service_config)` evaluates as a Boolean value, so the full condition is: 'are both Boolean values the same?' i.e. both present or both absent.

I think your suggestion would trigger when 'PAT' is set but 'password' is not.

Is this more readable?:
```
password_set = 'password' in service_config
pat_set = 'pat' in service_config
if (password_set and pat_set) or not (password_set or pat_set):
```

ryneeverett: Ah, yeah the `not` may need to be outside the second parenthesis. Use whichever route you find the most readable.

twaugh: Updated.

ryneeverett (approved):

PR #859, opened 2021-12-12
Fix GMail test

StephanMeijer:


ryneeverett (approved):
I can't reproduce the failure but I see no harm in the change.

PR #861, opened 2022-01-09
Add parent field to Jira issues

d-miketa:
Identifies the parent issue for Jira.

https://github.com/ralphbean/bugwarrior/issues/860

### DIFF
--- a/tests/test_jira.py
+++ b/tests/test_jira.py
@@ -68,6 +68,7 @@ class TestJiraIssue(AbstractServiceTest, ServiceTest):
     arbitrary_estimation = 3600
     arbitrary_id = '10'
     arbitrary_subtask_ids = ['11', '12']
+    arbitrary_parent_id = '13'
     arbitrary_project = 'DONUT'
     arbitrary_summary = 'lkjaldsfjaldf'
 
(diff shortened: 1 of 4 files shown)


tests/test_jira.py
```diff
@@ -80,14 +81,16 @@ class TestJiraIssue(AbstractServiceTest, ServiceTest):
...
+            'parent': {'key': f'DONUT-{arbitrary_parent_id}'}
         },
         'key': '%s-%s' % (arbitrary_project, arbitrary_id, ),
     }
 
     arbitrary_record_with_due = arbitrary_record.copy()
-    arbitrary_record_with_due['fields']=arbitrary_record_with_due['fields'].copy()
-    arbitrary_record_with_due['fields']['Sprint']=['com.atlassian.greenhopper.service.sprint.Sprint@4c9c41a5[id=2322,rapidViewId=1173,\
+    arbitrary_record_with_due['fields'] = arbitrary_record_with_due['fields'].copy(
+    )
```
ryneeverett, line 92: I'm fine with sneaking in minor format improvements in code near the code you're touching, but this doesn't seem like an improvement.

ryneeverett (approved):

tests/test_jira.py
d-miketa: You're right - honestly not sure what happened there, I just let Black take the wheel.

Do you want me to revert this line and submit again?

ryneeverett: Nah, don't worry about it.

I hadn't noticed at the time that you also added the spacing around the equals sign and that improvement offsets the oddity of putting the finally parenthesis on a separate line. I'd probably like to blackify this whole repo eventually anyway.

d-miketa: Ah, yeah I also missed the spacing change. Probably hit the width limit at that point. Once you blackify, perhaps with a slightly wider line (120?) this'll probably go away.

PR #862, opened 2022-01-09
Add support for extra fields

d-miketa:
Jira boards are very customisable. This PR adds the ability to declare extra UDA fields from `bugwarriorrc`.

https://github.com/ralphbean/bugwarrior/issues/860

### DIFF
--- a/bugwarrior/services/jira.py
+++ b/bugwarrior/services/jira.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from builtins import str
 import sys
+from functools import reduce
 
 from jinja2 import Template
 from jira.client import JIRA as BaseJIRA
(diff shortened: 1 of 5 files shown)


bugwarrior/services/jira.py
```diff
@@ -174,10 +237,21 @@ def to_taskwarrior(self):
             self.SUBTASKS: self.get_subtasks(),
         }
 
+        extrafields = self.get_extrafields()
+
+        return {**fixed_fields, **extrafields}
```
ryneeverett: To me this looks weird -- can we please consistently include or omit an underscore in these variables?

Personally I'd just add `**self.get_extrafields(),` to the end of the existing dict, eliminating the need for variables. But I leave that to your discretion.

bugwarrior/services/jira.py
```diff
@@ -16,6 +17,66 @@
...
+
+
+@pydantic.dataclasses.dataclass
+class JiraExtraField:
+    label: str
+    keys: list[str]
+
+    @property
+    def id(self):
+        return self.keys[0]
```
ryneeverett: Shouldn't this be `'.'.join(keys)` since we don't know which key caused the `KeyError`? (I'm assuming that's the only usage of this method.)

bugwarrior/services/jira.py
```diff
@@ -16,6 +17,66 @@
...
+                raise ExtraFieldConfigError(extrafield_raw)
+
+            keys = keys.split('.')
+
+            extrafield = JiraExtraField(label, keys)
+            extrafields.append(extrafield)
+        return extrafields
+
+
+@pydantic.dataclasses.dataclass
```
ryneeverett, line 60: Couldn't we use a standard library `dataclasses.dataclass` here? I guess we still support python3.6, which doesn't include dataclasses, so maybe just add a comment that we can switch once we drop 3.6?

ryneeverett (approved):
I'd like to see my comments addressed before merging but they're fairly trivial in the scope of this patch.

bugwarrior/services/jira.py
d-miketa: Yeah I agree and have fixed it.

The reason for this format was because I originally wanted to implement this for fields of the form `customfield_12345`, so I called the variable `customfields`. Then I realised I can target _any_ field, so switched to the more generic `extrafields`. But of course now that no longer makes sense!

ryneeverett:
Merged in 55db4c4a212896f5807ea61ab95ace434296f080.

bugwarrior/services/jira.py
```diff
@@ -33,7 +94,10 @@ class JiraConfig(config.ServiceConfig, prefix='jira'):
     verify_ssl: bool = True
     version: int = 5
 
-    @pydantic.root_validator
+    extra_fields: JiraExtraFields = pydantic.Field(
+        default_factory=lambda: JiraExtraFields(''))
```
ryneeverett, line 98: Note that in the rebase I changed this to `extra_fields: typing.Optional[JiraExtraFields] = None` for consistency with other non-required fields with empty defaults.

PR #876, opened 2022-02-22
Allow custom query strings for gitlab config

fmauch:
Ever since I started using bugwarrior, using it for our gitlab instance felt suboptimal. Especially since I started using github more and more I discovered the custom search strings for the gitlab service.

This PR aims at providing similar functionality for the Gitlab service.

To motivate my use case a bit more: I wanted to have all issues that are assigned to me on our gitlab instance in my bugwarrior syncs. Unfortunately I can see quite a lot of projects inside that gitlab instance and since filtering by assignee happens **after** every issue from every project is pulled, this took more than half an hour for me.

This is somehow related to #593 but takes a different approach at a similar problem.

While implementing the tests I noticed that the tasks created for ToDos are quite different in some places. I separated the fields that were surprising to me in 5ed972c. Since I copied the ToDo extraction from the existing implementation and adapted it only a little, I assume this is not related to something I implemented.
I noticed that yielding of merge_request and todo items is currently not covered by the tests at all, so this might be something worth fixing at another point. I also might be completely wrong, though.

Being a maintainer of one or two projects myself I know that larger refactory suggestions from external users can feel intrusive. However, I hope I made my motivation clear and my contribution is more than just a code dump in "works for me" quality.

I am also aware, that there are things missing, such as config documentation for this new feature, hence the draft status.

---

**Edit**: Looking back at the todo implementation, it is quite obvious that many of the unexpected task fields are directly hard coded in the code I copied. So, in order to keep things from diverging further I think it would not make sense to change the implementation (at least not my implementation only). It might, however, make sense to call my new function for yielding todos from the `issue` method in case of keeping the code duplication. (**Edit2**: Did in 139abdd)

### DIFF
--- a/bugwarrior/docs/services/gitlab.rst
+++ b/bugwarrior/docs/services/gitlab.rst
@@ -173,6 +173,33 @@ you can configure individual priorities for each::
     gitlab.default_todo_priority = M
     gitlab.default_mr_priority = H
 
+
+Custom query strings
+++++++++++++++++++++
+
+The Gitlab REST API allows many more configuration options than the ones provided
+by the options explained above. If you want to further customize calls, you can set for example::
+
+    gitlab.issue_query = issues?search=foo&in=title
+    gitlab.merge_request_query = merge_requests?state=opened&scope=all&reviewer_username=myusername
+    gitlab.todo_query = todos?state=pending&action=directly_addressed
+
+
+These can be combined with the other configuration options above, but queries are only evaluated if
+the respective category (issue, merge_request, todo) is enabled.
+
+Note: Depending in the scope you are interested in, this query-based approach can be much faster
+than using the "default queries". For example, imagine that you want to query all issues assigned
+to your user.
+
+This can be achieved by leaving the ``gitlab.include_repos`` configuration value empty and
+setting ``gitlab.only_assigned`` to ``True``. This will result in querying all repos your user has
+access to, which might take a very long time.
+
+Alternatively, you could set ``gitlab.issue_query =
+issues?assignee_username=myusername&state=opened&scope=all``, which will fetch the assigned issues
+first and then only fetch the projects for which issues have been found.
+
 Use HTTP
 ++++++++
 
--- a/bugwarrior/docs/services/gitlab.rst
+++ b/bugwarrior/docs/services/gitlab.rst
@@ -133,6 +133,13 @@ to all fields on the Taskwarrior task if needed.
    See :ref:`field_templates` for more details regarding how templates
    are processed.
 
+Include Issues
+++++++++++++++
+
+Issues are included by default, if not configured otherwise. To disable querying of issues, set::
+
+    gitlab.include_issues = False
+
 Include Merge Requests
 ++++++++++++++++++++++
 
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -33,6 +33,7 @@ class GitlabConfig(config.ServiceConfig, prefix='gitlab'):
     import_labels_as_tags: bool = False
     label_template: str = '{{label}}'
     filter_merge_requests: bool = False
+    include_issues: bool = True
     include_todos: bool = False
     include_all_todos: bool = True
     only_if_author: str = ''
@@ -664,18 +665,19 @@ class GitlabService(IssueService):
         # List of repos will only be queried if needed
         repos = []
 
-        # Issues are always queried
-        if self.config.issue_query:
-            issues = self.gitlab_client.get_issues_from_query(self.config.issue_query)
-        else:
-            if not repos:
-                repos = self.get_all_repos()
-            issues = self.get_issues_from_projects(repos)
+        # Issues
+        if self.config.include_issues:
+            if self.config.issue_query:
+                issues = self.gitlab_client.get_issues_from_query(self.config.issue_query)
+            else:
+                if not repos:
+                    repos = self.get_all_repos()
+                issues = self.get_issues_from_projects(repos)
 
-        log.debug("Found %i issues.", len(issues))
-        issues_filtered = list(filter(self.include, issues.values()))
-        log.debug("Pruned down to %i issues.", len(issues_filtered))
-        yield from self._get_issue_objs(issues_filtered, 'issue')
+            log.debug("Found %i issues.", len(issues))
+            issues_filtered = list(filter(self.include, issues.values()))
+            log.debug("Pruned down to %i issues.", len(issues_filtered))
+            yield from self._get_issue_objs(issues_filtered, 'issue')
 
         # Merge requests
         if self.config.filter_merge_requests:
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -42,6 +42,9 @@ class GitlabConfig(config.ServiceConfig, prefix='gitlab'):
     use_https: bool = True
     verify_ssl: bool = True
     project_owner_prefix: bool = False
+    issue_query: str = ''
+    merge_request_query: str = ''
+    todo_query: str = ''
 
     @pydantic.root_validator
     def namespace_repo_lists(cls, values):
@@ -255,12 +258,13 @@ class GitlabClient(ServiceClient):
             issues[issue['id']] = (issue['project_id'], issue)
         return issues
 
-    def get_todos(self) -> list:
+    def get_todos(self, query: str) -> list:
         """Get all todo objects matching a query returned as list of (project_id, todo) tuples
 
+        :param query: API query string that should get sent to the server
+        :type query: str
         :rtype: list
         """
-        query = 'todos?state=pending'
         todos = []
         try:
             fetched_todos = self._fetch_paged(query)
@@ -568,7 +572,7 @@ class GitlabService(IssueService):
             return project is None or project['id'] in ids
         return include_todo
 
-    def _get_issue_objs(self, issues, issue_type, repo_map):
+    def _get_issue_objs(self, issues, issue_type):
         type_plural = issue_type + 's'
 
         for rid, issue in issues:
@@ -588,6 +592,31 @@ class GitlabService(IssueService):
             }
             issue_obj.update_extra(extra)
             yield issue_obj
+
+    def _get_todo_objs(self, todos):
+        for project, todo in todos:
+            if project is not None:
+                repo = project
+            else:
+                repo = {
+                    'path': 'the instance',
+                }
+            todo['repo'] = repo['path']
+
+            todo_obj = self.get_issue_for_record(todo)
+            todo_url = todo['target_url']
+            project_name = repo['path']
+            if self.config.project_owner_prefix:
+                project_name = repo['namespace']['path'] + "." + project_name
+            extra = {
+                'issue_url': todo_url,
+                'project': project_name,
+                'namespace': "todo",
+                'type': 'todo',
+                'annotations': [],
+            }
+            todo_obj.update_extra(extra)
+            yield todo_obj
 
     def include(self, issue):
         """ Return true if the issue in question should be included """
@@ -631,64 +660,56 @@ class GitlabService(IssueService):
         return repos
 
     def issues(self):
-        repos = self.get_all_repos()
 
-        repo_map = {}
-        issues = {}
-        for repo in repos:
-            rid = repo['id']
-            repo_map[rid] = repo
-            issues.update(
-                self.gitlab_client.get_repo_issues(rid)
-            )
-        log.debug(" Found %i issues.", len(issues))
-        issues = list(filter(self.include, issues.values()))
-        log.debug(" Pruned down to %i issues.", len(issues))
+        # List of repos will only be queried if needed
+        repos = []
 
-        yield from self._get_issue_objs(issues, 'issue', repo_map)
+        # Issues are always queried
+        if self.config.issue_query:
+            issues = self.gitlab_client.get_issues_from_query(self.config.issue_query)
+        else:
+            if not repos:
+                repos = self.get_all_repos()
+            issues = self.get_issues_from_projects(repos)
 
-        if not self.config.filter_merge_requests:
-            merge_requests = {}
-            for repo in repos:
-                rid = repo['id']
-                merge_requests.update(
-                    self.gitlab_client.get_repo_merge_requests(rid)
-                )
-            log.debug(" Found %i merge requests.", len(merge_requests))
-            merge_requests = list(filter(self.include, merge_requests.values()))
-            log.debug(" Pruned down to %i merge requests.", len(merge_requests))
+        log.debug("Found %i issues.", len(issues))
+        issues_filtered = list(filter(self.include, issues.values()))
+        log.debug("Pruned down to %i issues.", len(issues_filtered))
+        yield from self._get_issue_objs(issues_filtered, 'issue')
 
-            yield from self._get_issue_objs(merge_requests,
-                                            'merge_request',
-                                            repo_map)
+        # Merge requests
+        if self.config.filter_merge_requests:
+            if self.config.merge_request_query:
+                merge_requests = self.gitlab_client.get_issues_from_query(
+                    self.config.merge_request_query)
+            else:
+                if not repos:
+                    repos = self.get_all_repos()
+                merge_requests = {}
+                for repo in repos:
+                    rid = repo['id']
+                    merge_requests.update(
+                        self.gitlab_client.get_repo_merge_requests(rid)
+                    )
+            log.debug("Found %i merge requests.", len(merge_requests))
+            merge_requests_filtered = list(filter(self.include, merge_requests.values()))
+            log.debug("Pruned down to %i merge requests.", len(merge_requests_filtered))
 
+            yield from self._get_issue_objs(merge_requests_filtered, 'merge_request')
+
+        # ToDos
         if self.config.include_todos:
-            todos = self.gitlab_client.get_todos()
+            query = 'todos?state=pending'
+            if self.config.todo_query:
+                query = self.config.todo_query
+
+            todos = self.gitlab_client.get_todos(query)
             log.debug(" Found %i todo items.", len(todos))
             if not self.config.include_all_todos:
-                todos = list(filter(self.include_todo(repos), todos))
-            log.debug(" Pruned down to %i todos.", len(todos))
-
-            for project, todo in todos:
-                if project is not None:
-                    repo = project
-                else:
-                    repo = {
-                        'path': 'the instance',
-                    }
-                todo['repo'] = repo['path']
-
-                todo_obj = self.get_issue_for_record(todo)
-                todo_url = todo['target_url']
-                projectName = repo['path']
-                if self.config.project_owner_prefix:
-                    projectName = repo['namespace']['path'] + "." + projectName
-                extra = {
-                    'issue_url': todo_url,
-                    'project': projectName,
-                    'namespace': "todo",
-                    'type': 'todo',
-                    'annotations': [],
-                }
-                todo_obj.update_extra(extra)
-                yield todo_obj
+                if not repos:
+                    repos = self.get_all_repos()
+                todos_filtered = list(filter(self.include_todo(repos), todos))
+            else:
+                todos_filtered = todos
+            log.debug(" Pruned down to %i todos.", len(todos_filtered))
+            yield from self._get_todo_objs(todos_filtered)
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -70,6 +70,191 @@ class GitlabConfig(config.ServiceConfig, prefix='gitlab'):
         return values
 
 
+class GitlabClient(ServiceClient):
+    """Abstraction of Gitlab API v4"""
+
+    def __init__(self, host, token, use_https, verify_ssl):
+        if use_https:
+            self.scheme = 'https'
+        else:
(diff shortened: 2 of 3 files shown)


bugwarrior/services/gitlab.py
```diff
@@ -295,6 +296,9 @@ def __init__(self, *args, **kw):
         else:
             self.scheme = 'http'
 
+        self.base_tmpl = f"{self.scheme}://{self.auth[0]}/api/v4/"
```
ryneeverett: Looks like azuredevops and youtrack call this `base_url` and I think that would be more appropriate since it's not really a template.

bugwarrior/services/gitlab.py
```diff
@@ -484,87 +487,150 @@ def include(self, issue):
...
-                extra = {
-                    'issue_url': todo_url,
-                    'project': projectName,
-                    'namespace': "todo",
-                    'type': 'todo',
-                    'annotations': [],
-                }
-                todo_obj.update_extra(extra)
-                yield todo_obj
+            yield from self.yield_todos_from_query('todos?state=pending')
```
ryneeverett: So if the user sets a `query_string` and sets `include_todos` wont the todos be fetched twice -- once at the beginning and once at the end of this method?

bugwarrior/services/gitlab.py
```diff
@@ -42,6 +42,7 @@ class GitlabConfig(config.ServiceConfig, prefix='gitlab'):
     use_https: bool = True
     verify_ssl: bool = True
     project_owner_prefix: bool = False
+    query_string: str = ''
```
ryneeverett: Any reason not to simply call this `query`, as with github?

ryneeverett (commented):
Ideally I would like to see fetching/parsing/task-construction decoupled from filtering so that the error-prone application of our filtering options doesn't have to be implemented multiple times. Is it practical to maintain this property?

For example, in `yield_issues_from_query` you're applying the `include` filter and it's applied to the results of the `projects` api later in `issues`. If practical, I'd like these methods to return tasks so we can do all our filtering in the `issues` method and for the filtering logic to be ignorant of whether the tasks were returned from the `projects` api or one of the api's your new method makes use of.

This is a somewhat arbitrary application of my overall impression, which is that -- as with github -- use of the `query_string` option should not trigger a completely alternate code path. (`if self.config.query_string` / `else`)

bugwarrior/services/gitlab.py
fmauch: absolutely not. Will change that!

bugwarrior/services/gitlab.py
fmauch: seems reasonable, will change.

bugwarrior/services/gitlab.py
fmauch: yes, this should have been inside the `if self.config.query_string` / `else` block. That will be changing anyways, as I agree with the points you've risen.

fmauch: Resolved in 0394934

bugwarrior/services/gitlab.py
fmauch: Resolved in 4046086

bugwarrior/services/gitlab.py
fmauch: Resolved in 24ef4bf

fmauch:
I've addressed a couple of the issues you mentioned. 

Completely getting rid of the split path isn't that easy in my opinion.

- We would have to add a flag whether we want to include issues or not. Imagine I specify a query for MRs. Should I download issues in that case, as well? For ToDo's I have a separate flag, for issues I don't.
- We would probably need to define queries for all three types of items. Otherwise we would end up with weird configs if we don't split the execution paths. What should I do if I define a query for issues and set `include_todos` to true? If I define a query for merge_requests, do I have to set `filter_merge_requests` to `True`?

If we had a config flag for each type of item (issue, mr, todo), we could do something like the following (semi-pseudo-code):
```
issues = {}
merge_requests = {}
todos = {}

if include_issues:
  if config.issue_query:
    issues = issues_from_query()
  else:
    issues = issues_from_projects_api()
  issues_filtered = filter(issues)
  yield from get_issue_obj(issues)

if include_merge_requests:
 if config.merge_request_query:
   merge_requests = merge_requests_from_query()
 else:
    merge_requests = merge_requests_from_projects_api()
  merge_requests_filtered = filter(merge_requests)
  yield from get_issue_obj(merge_requests)

if include_todos:
  if config.todo_query:
    todos = todos_from_query()
  else:
    todos = todos_from_projects_api()
  todos_filtered = filter(todos)
  for todo in todos_filtered:
    transformed = transform(todo)
    yield transformed
```

Without such a refactoring I don't really see a clean way how to implement this. Which is not nice, as I also see the point of leaving users with unclear configuration options as in "which option can I use in combination with query?".

bugwarrior/services/gitlab.py
```diff
@@ -295,6 +296,9 @@ def __init__(self, *args, **kw):
         else:
             self.scheme = 'http'
 
+        self.base_url = f"{self.scheme}://{self.auth[0]}/api/v4/"
```
ryneeverett: On second thought, is this attribute even necessary? Couldn't we just make this a local variable in `_fetch`, pass relative url's to `_fetch`, and append the url to it there since that's where it always ultimately gets passed? (e.g., `absolute_url = base_url + url`)

bugwarrior/services/gitlab.py
```diff
@@ -500,71 +567,58 @@ def issues(self):
...
+                items = self.get_issues_from_query(self.config.query)
+
+                type_plural = item_type + 's'
+                log.debug("Found %i %s.", len(items), type_plural)
+                items = list(filter(self.include, items.values()))
+                log.debug("Pruned down to %i %s.", len(items), type_plural)
+                self.update_repo_map(items)
+                yield from self._get_issue_objs(items, item_type)
+            else:
+                yield from self.yield_todos_from_query('todos?state=pending')
```
ryneeverett: It seems like if the query starts with "todos" then anything else in the query will be ignored in favor of this hard coded query.

bugwarrior/services/gitlab.py
```diff
@@ -500,71 +567,58 @@ def issues(self):
...
+                    )
+                log.debug(" Found %i merge requests.", len(merge_requests))
+                merge_requests = list(filter(self.include, merge_requests.values()))
+                log.debug(" Pruned down to %i merge requests.", len(merge_requests))
+
+                yield from self._get_issue_objs(merge_requests,
+                                                  'merge_request')
+
+            if self.config.include_todos:
+                yield from self.yield_todos_from_query('todos?state=pending')
```
ryneeverett: Sorry, I think I was wrong and todos should be top level here so that this option will still work if the query is for issues or merge requests. Perhaps we should add a validator that throws an error if `item_type == 'todo' and self.config.include_todos`.

ryneeverett (commented):
I agree that eliminating the split path is not practical here but I still think we should look for opportunities to reduce the divergence. I now understand why there's more divergence than with the github service though -- Github's api allows querying for both issues and pull requests with one query, whereas Gitlab has separate api's. There are a few different ways we could handle this:

1. The way you handled it, with `query` overriding issues and merge requests. The problem is that users then have to have two different service configurations if they want both issues and merge requests because there's no way to get both with one query, right?
2. We could go with your example and have separate `issue_query`, `merge_request_query`, and `todo_query` options. (I'd rather not add additional `include` options in this pull request as the scope big enough already.) Adding more configuration options where one would suffice is not ideal.
3. We could make the option `queries` and allow a list of queries so that users can provide queries for both issues and merge requests if desired. The problem is that there is then not a great way to get the default behavior for issues or pull requests while overriding the other.

I think option **2** is the least bad option, which would make your pseudocode a fairly accurate depiction of the control flow.

Additionally, I think the additional complication of your patch justifies factoring the `ServiceClient` out of the `IssueService` and into a separate class. (For an example, see the github service.)

bugwarrior/services/gitlab.py
fmauch: as everything goes through `_fetch` that's actually true.

bugwarrior/services/gitlab.py
fmauch: yes, that's not intentional, thanks!

bugwarrior/services/gitlab.py
fmauch: I'll have another look

fmauch:
I've pushed another version implementing a draft on option 2 from [your comment](https://github.com/ralphbean/bugwarrior/pull/876#pullrequestreview-894422363)

I couple of thoughts on this draft version:
 - I had to move the `update_repo_map` call out if the query-string block, as it operates on the filtered list. This could probably be refactored to get it back included, though.
 - I added a `include_issues` config option as in my humble opinion it should be possible to have a job that **only** fetches the MRs I am assigned as a reviewer. For example, I like to add reviews I am assigned to with a higher priority as this is something other devs are waiting for.
 - With this we have `include_issues`, `include_todos` and `filter_merge_requests` as boolean flags. Obviously, this is confusing. However, deprecating `filter_merge_requests` in favor of `include_merge_requests` would probably be out of scope of this PR.
 - There is a lot of code duplication between issues and merge_requests. However, further refactoring would probably lead to less readable code at this point.
 - ~~ToDos definitely need some more refactoring.~~ (done in 50a5b64)
 - I think the code structure is now quite understandable and the configuration is also rather logical. Each component can be switched on and off and their default behavior can be overwritten.
 - ~~There is (at least) one unlogical piece of configuration: Configuring a custom query string will ignore the the `include_repos` and such.~~ (fixed in 8394db1)

fmauch:
@ryneeverett I think I am happy with the current implementation. If you don't have any larger complaints I will go ahead adding unit tests and updating the documentation next.

bugwarrior/services/gitlab.py
```diff
@@ -287,12 +482,13 @@ def __init__(self, *args, **kw):
...
-            self.scheme = 'https'
-        else:
-            self.scheme = 'http'
+        self.gitlab_client = GitlabClient(
+            host=self.config.host,
+            token=token,
+            use_https=self.config.use_https,
+            verify_ssl=self.config.verify_ssl
+        )
+        self.repo_map = dict()
```
ryneeverett, line 503: You should be able to drop the `ServiceClient` base class from `GitlabService` now.

bugwarrior/services/gitlab.py
```diff
@@ -483,87 +630,86 @@ def include(self, issue):
...
+                self.repo_map[rid] = repo
 
-        repo_map = {}
+    def get_issues_from_projects(self, repos):
         issues = {}
         for repo in repos:
             rid = repo['id']
-            repo_map[rid] = repo
+            self.repo_map[rid] = repo
+            self.repo_map[rid] = repo
```
ryneeverett: duplicate lines

bugwarrior/services/gitlab.py
```diff
@@ -70,6 +74,197 @@ def default_priorities(cls, values):
...
+        """
+        issues = {}
+        try:
+            repo_issues = self._fetch_paged('projects/%d/issues?state=opened' % rid)
+        except OSError:
+            # Projects may have issues disabled.
+            return {}
+        for issue in repo_issues:
+            issues[issue['id']] = (rid, issue)
+        return issues
```
ryneeverett: I'm still thinking there are opportunities to reduce duplication. For example, couldn't `get_repo_issues` be eliminated by simply passing `'projects/%d/issues?state=opened' % rid` into `get_issues_from_query`? (I suspect the structure of `get_repo_issues` is slightly more correct though.)

ryneeverett (commented):
This is really coming along. I agree that the high level structure is pretty solid at this point.

bugwarrior/services/gitlab.py
fmauch: Good point. I've changed the calls accordingly.

bugwarrior/services/gitlab.py
```diff
@@ -70,6 +74,181 @@ def default_priorities(cls, values):
...
+            self.scheme = 'https'
+        else:
+            self.scheme = 'http'
+        self.verify_ssl = verify_ssl
+
+        self.host = host
+        self.token = token
+
+    def _base_url(self):
+        return f"{self.scheme}://{self.host}/api/v4/"
```
ryneeverett, line 93: Can't `base_url` just be a local variable in `_fetch`?

bugwarrior/docs/services/gitlab.rst
```diff
@@ -173,6 +180,28 @@ you can configure individual priorities for each::
...
+    gitlab.todo_query = todos?state=pending&action=directly_addressed
+
+
+These can be combined with the other configuration options above, but queries are only evaluated if
+the respective category (issue, merge_request, todo) is enabled.
+
+Note: Depending in the scope you are interested in, this query-based approach can be much faster
+than using the "default queries". For example, if you want to query all issues assigned to your
+user, without explicitly setting up a query and leaving the ``include_repos`` configuration empty,
+will result in querying all repos your user has access to, which might take a very long time. If,
```
ryneeverett: grammar: This sentence could use some work. It is comprehensible though, so don't worry about it if you don't see a problem. (I have no idea what your English level is.)

bugwarrior/services/gitlab.py
```diff
@@ -70,6 +74,181 @@ def default_priorities(cls, values):
...
+            full += items
+
+            if len(items) < params['per_page']:
+                break
+            params['page'] += 1
+
+        return full
+
+    def get_repos(self, include_repos=None, include_regex=None,
+                  only_membership=False, only_owned=False):
```
ryneeverett: Could we remove the defaults on these arguments? I'd like to avoid double-defining the default values of configuration fields as this could lead to some really confusing situations if the defaults diverged.

I realize that you probably did this to make testing easier, but IMO it's not worth the tradeoff.

bugwarrior/services/gitlab.py
```diff
@@ -70,6 +74,181 @@ def default_priorities(cls, values):
...
+
+        :param query: API query string that should get sent to the server
+        :type query: str
+        :rtype: dict
+        """
+        issues = {}
+        try:
+            result = self._fetch_paged(query)
+            for issue in result:
+                issues[issue['id']] = (issue['project_id'], issue)
```
ryneeverett: This `for` block should be outside the `try` block, as in the original `get_repo_issues`/`get_repo_merge_requests`.

bugwarrior/services/gitlab.py
```diff
@@ -70,6 +74,181 @@ def default_priorities(cls, values):
...
+        except OSError:
+            # Projects may have issues disabled.
```
ryneeverett: Maybe adjust this comment to "Projects may have this API disabled."?

bugwarrior/services/gitlab.py
```diff
@@ -467,8 +568,38 @@ def _get_issue_objs(self, issues, issue_type, repo_map):
...
+                'namespace': "todo",
+                'type': 'todo',
+                'annotations': [],
+            }
+            todo_obj.update_extra(extra)
+            yield todo_obj
+
     def include(self, issue):
         """ Return true if the issue in question should be included """
+        self.update_repo_map([issue])
```
ryneeverett: I think we should avoid side effects in filter functions. Also, it seems like `update_repo_map` would get called twice for issues and merge requests -- once in `_get_issue_objs` and once in `include`. Couldn't we just add a call to `update_repo_map` to `get_todo_objs` and eliminate the one in `include`?

bugwarrior/services/gitlab.py
```diff
@@ -483,87 +614,85 @@ def include(self, issue):
...
+        # Issues
+        if self.config.include_issues:
+            if self.config.issue_query:
+                issues = self.gitlab_client.get_issues_from_query(self.config.issue_query)
+            else:
+                if not repos:
+                    repos = self.get_all_repos()
+                issues = self.get_issues_from_projects(repos)
+
+            log.debug("Found %i %s.", len(issues), "issues")
```
ryneeverett: Please don't fill in a template with a hard-coded string on the same line.

bugwarrior/services/gitlab.py
```diff
@@ -483,87 +614,85 @@ def include(self, issue):
...
+            issues_filtered = list(filter(self.include, issues.values()))
+            log.debug("Pruned down to %i %s.", len(issues_filtered), "issues")
```
ryneeverett: Please don't fill in a template with a hard-coded string on the same line.

ryneeverett (commented):
We're very close! Thanks for your persistence.

bugwarrior/services/gitlab.py
fmauch: sure. I basically made this equivalent to the implementation in the Github service, but I can move that.

bugwarrior/docs/services/gitlab.rst
fmauch: I am not a native speaker, but I will try to improve that sentence :-)

bugwarrior/services/gitlab.py
fmauch: Sure!

bugwarrior/services/gitlab.py
fmauch: Actually, my motivation was so that we can easily test the different configurations, as https for instance is set in the constructor. Do you think that is worth it? Otherwise, I will move it back into `_fetch()`

bugwarrior/services/gitlab.py
fmauch: I've changed that function in 2744b96 a bit. As the `include_regex` parameter was not really used inside the function, I decided to drop it. In the old structure it was used to only use `include_repos` if `include_regex` was not set in the config. So basically, this seems as if it was just used, because it was there. However, in my opinion the decision whether `include_repos` should be used or not belongs into the `Service`, not the `Client`.

bugwarrior/services/gitlab.py
fmauch: woops! That's  a leftover from a refactoring, but looking at this looks really silly, yes.

bugwarrior/services/gitlab.py
fmauch: I'll have a  look at that again, maybe tonight.

bugwarrior/docs/services/gitlab.rst
fmauch: I probably should not write documentation in the middle of the night just to get it done. I've pushed  a version which should be more understandable.

bugwarrior/services/gitlab.py
fmauch: The thing that's giving me a headache is that with the two possible paths we want to filter for `include_repos` and such at two very different places.

* For the project API we want to filter the repositories **before** we query any issues, as otherwise we would be querying issues for projects that we are actually not interested in. As each individual query is very time/intensive we want to reduce the number of necessary calls as much as possible.
* For the issues API we want to first fetch all matching issues and then filter out the projects as we cannot add a repo filter to the issues query.

4f4d188309161ae05033093a870c182facfb3597 and 8394db1c569c86509a945563bdcee3fe9ae64d54 explain my motivation for the way I've implemented this.

Inside `_get_issue_objects` we need the `repo_map` to be filled, as we fetch a lot of info from the project there. The call tp `update_repo_map` in there ensures that the repo information is actually present. The way `update_repo_map` is implemented, it basically works as a caching mechanism. If repo information about the repo with the given idea is already there, then don't do anything. Therefore, calling it twice is not necessarily a bad thing performance-wise.

Moving the `update_repo_map` call out of the functions would effectively add a precondition to the functions, which is not the best thing to introduce. I've added an alternative suggestion in 8ad81b7, but I am not sure whether this approach really carries a benefit over the one discussed here.

Another alternative I could think of is wrapping the `repo_map` into an own cache data structure with something like `get_repo_info_cached` that would either query the info or read it from the cache. This would make the `update_repo_map` function obsolete. Algorithmically not too much would change to my initial approach, but it would be clearer to other developers, that the `repo_map` is acting as a cache. It would not introduce any preconditions to the `include` function or the `_get_issue_objs` function. 

---

Edit: I've implemented the latter in 650ae5b

bugwarrior/services/gitlab.py
ryneeverett: I'll defer to your judgement on this one.

bugwarrior/docs/services/gitlab.rst
ryneeverett: Looks good, thanks!

bugwarrior/services/gitlab.py
ryneeverett: Very nice solution. This makes the repo map a lot easier to understand.

ryneeverett (approved):

fmauch:
@ryneeverett should I cleanup the commit history before merging this? We've done quite a lot of back and forth in this PR. Of course, one can argue, that leaving it this way it reflects the discussion history in this PR. What's your preferred workflow here?

ryneeverett:
The way it is now, I would probably squash the entire PR into one commit. I think that gives a more accurate picture of the commit history in the context of the project as a whole because most of these commits only modify code that's already been added or modified by this PR. However, if you wanted to rewrite the history into a few commits that stand on their own (i.e., bugwarrior should still work, the tests should still pass, etc.) I'd go with that. For example, I could see something like:

1. gitlab: Refactor ServiceClient out of IssueService.
2. gitlab: Implement repo_cache.
3. gitlab: Custom query strings.
4. gitlab: Add include_issues.

fmauch:
@ryneeverett done rewriting history.

PR #881, opened 2022-03-11
filter_merge_requests -> ! include_merge_requests

### LINKED ISSUE #315: Inconsistent behavior of `filter_{merge,pull}_requests`
The option `filter_merge_requests` or `filter_pull_requests` is documented in gitlab, github and bitbucket like this

> Although you can filter issues using Common Service Configuration Options, pull requests are not 
> filtered by default. You can filter pull requests by adding the following configuration option:
> 
> {service}.filter_pull_requests = True

Which I understand as
- filter_pull_requests = True :arrow_right: prune pull requests according to the assignee (like issues)
- filter_pull_requests = False :arrow_right: import all pull requests

This correspond to the behavior implemented in the github service.

But in both bitbucket and gitlab the implementation seems to do this instead:
- filter_pull_requests = True :arrow_right: don't import pull requests
- filter_pull_requests = False :arrow_right: import pull requests, pruned according to assignee

That's a tad confusing :confused:

ryneeverett:
Deprecate the confusing and inconsistent filter_merge_requests options
in gitlab and bitbucket in favor of an inverse include_merge_requests
field.

Fix #315. I've basically implemented @mathstuf's suggestion.

cc @fmauch

### DIFF
--- a/bugwarrior/config/schema.py
+++ b/bugwarrior/config/schema.py
@@ -313,3 +313,17 @@ class ServiceConfig(_ServiceConfig,  # type: ignore  # (dynamic base class)
         typing_extensions.Literal['L', 'M', 'H']] = 'M'
     add_tags: ConfigList = ConfigList([])
     description_template: str = ''
+
+    @pydantic.root_validator
+    def deprecate_filter_merge_requests(cls, values):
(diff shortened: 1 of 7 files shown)


fmauch:
Awesome, that's a good addition IMHO.

PR #886, opened 2022-03-23
Redmine query parameters

CaptainQuirk:
:dart: Goals
---

- Provide a mechanism to filter issues

:panda_face: Implementation
---

- Support a query configuration value to filter issues

:notebook_with_decorative_cover: Documentation
---

- document the use of the query setting and how it overrides other configuration value
- link to redmine issues api documentation

### DIFF
--- a/bugwarrior/docs/services/redmine.rst
+++ b/bugwarrior/docs/services/redmine.rst
@@ -28,6 +28,21 @@ If you want to ignore verifying the SSL certificate, set::
 
     redmine.verify_ssl = False
 
+Specify the parameters to filter issues
++++++++++++++++++++++++++++++++++++++++
+
+If you want a finer control over the filtering of issues, you may combine
+`service.only_if_assigned` and `service.issue_limit` with the `redmine.query`
+setting.
+
+    redmine.query = project_id=10&status_id=15
+
+Note that no `&` or `?` character is needed at the beginning of the setting
+value.
+
+You may find documentation about the redmine issue api
+`on this page <https://www.redmine.org/projects/redmine/wiki/Rest_Issues>`_:
+
 Provided UDA Fields
 -------------------
 
--- a/bugwarrior/services/redmine.py
+++ b/bugwarrior/services/redmine.py
@@ -18,6 +18,7 @@ class RedMineConfig(config.ServiceConfig, prefix='redmine'):
 
     project_name: str = ''
     issue_limit: int = 100
+    query: str = ''
     login: str = ''
     password: str = ''
     verify_ssl: bool = True
@@ -31,16 +32,20 @@ class RedMineClient(ServiceClient):
         self.issue_limit = issue_limit
         self.verify_ssl = verify_ssl
 
-    def find_issues(self, issue_limit, only_if_assigned=False):
+    def find_issues(self, issue_limit, query, only_if_assigned=False):
         args = {}
-        # TODO: if issue_limit is greater than 100, implement pagination to return all issues.
-        # Leave the implementation of this to the unlucky soul with >100 issues assigned to them.
+        url = "/issues.json" + "?" + query
+
+        # TODO: if issue_limit is greater than 100, implement pagination
+        # to return all issues. Leave the implementation of this to
+        # the unlucky soul with >100 issues assigned to them.
         if issue_limit is not None:
             args["limit"] = issue_limit
 
         if only_if_assigned:
             args["assigned_to_id"] = 'me'
-        return self.call_api("/issues.json", args)["issues"]
+
+        return self.call_api(url, args)["issues"]
 
     def call_api(self, uri, params):
         url = self.url.rstrip("/") + uri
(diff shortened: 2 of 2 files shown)


bugwarrior/services/redmine.py
```diff
@@ -31,16 +32,21 @@ def __init__(self, url, key, auth, issue_limit, verify_ssl):
...
 
-    def find_issues(self, issue_limit, only_if_assigned=False):
+    def find_issues(self, issue_limit, query, only_if_assigned=False):
         args = {}
         # TODO: if issue_limit is greater than 100, implement pagination to return all issues.
         # Leave the implementation of this to the unlucky soul with >100 issues assigned to them.
-        if issue_limit is not None:
-            args["limit"] = issue_limit
+        url = "/issues.json"
+        if query is not None:
```
ryneeverett: Isn't the default value an empty string, rather than None?

bugwarrior/services/redmine.py
```diff
@@ -31,16 +32,21 @@ def __init__(self, url, key, auth, issue_limit, verify_ssl):
...
+            url = url + "?" + query
+        else:
+            if issue_limit is not None:
```
ryneeverett: The comments regarding `issue_limit` should remain above this line.

bugwarrior/docs/services/redmine.rst
```diff
@@ -28,6 +28,22 @@ If you want to ignore verifying the SSL certificate, set::
...
+
+If you need to take control over the filtering of issues, you may use the
+`redmine.query` setting.
+
+Be aware though that the `redmine.issue_limit` and `redmine.only_if_assigned`
+settings will be then be ignored.
+
+You'll have to include their redmine equivalent in the query::
+
+    redmine.query = project_id=10&status_id=15&limit=100&assigned_to_id=me
```
ryneeverett: Sorry if I misled you regarding the extent to which `query` should override default behavior -- I hadn't realized how much simpler the Redmine service is than the others in which we've implemented this option. In this case I don't think it should override either of these configuration options and therefore you could just add it to the conventional querystring as I believe you first suggested.

bugwarrior/services/redmine.py
CaptainQuirk: You're right ! Solved it

bugwarrior/services/redmine.py
CaptainQuirk: Done !

bugwarrior/docs/services/redmine.rst
CaptainQuirk: I handled it the way I first imagined it.

CaptainQuirk:
I also saw that a suite of github actions failed on my branch : https://github.com/ralphbean/bugwarrior/actions/runs/2031120214. I used pytest locally and got a 0 exit code (and a bunch of warnings about distutils version classes being deprecated, due to the fact that python is installed with homebrew on my Linux Mint)

ryneeverett:
It's not clear to me that there is any need for an `if`/`else` clause here. Couldn't we continue passing the limit and assigned_to_id parameters in `args` even with additional query parameters in the url?

CaptainQuirk:
How should I handle the `query` setting which is a partial query string, then ?
Due to its partial nature, I'm not sure I could parse it into a dict easily.

CaptainQuirk:
Maybe your thought was to both append the string value of the query setting to the url variable while also filling the arg dict ?

ryneeverett:
> Maybe your thought was to both append the string value of the query setting to the url variable while also filling the arg dict ?

Yes, that was my thought and I think it's the best route unless there is some reason it doesn't work.

> How should I handle the query setting which is a partial query string, then ?
Due to its partial nature, I'm not sure I could parse it into a dict easily.

I actually think it is fairly easy to parse a query string into a dict with [urlparse.parse_qs](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs).

However, if for some reason we're forced to choose between strings and dicts, I think it would be more reliable to serialize the limit and and assigned_to_id like you did than to parse the user's query string which is just another opportunity for error. However, I'd want to serialize it every time, not just when `query != ''`, so we still wouldn't need the `else` clause. We want to minimize logical branching. Make sense?

CaptainQuirk:
> I actually think it is fairly easy to parse a query string into a dict with urlparse.parse_qs.

A full blown query string, maybe, but something like `project_id=10&status_id=15` with no leading `&` or `?`, I'm not sure.

So to wrap it up  : 

- Try to add the append the given query setting to the url, keeping the `assigned_to_id` and `limit` parameters in the arg dict
- If it does not work, serialize `limit`, serialize `assigned_to_id` if `only_if_assigned` is set and append the query, if it is not the empty string

That should indeed reduce the branching !

ryneeverett:
I think we could completely remove the branching by appending `query` even if it is an empty string because I believe a trailing ampersand is acceptable.

CaptainQuirk:
That should do it now !

ryneeverett (approved):

PR #887, opened 2022-03-29
Allow empty default_priority. Resolve #426.

### LINKED ISSUE #426: Allow setting default priority to empty
Hi, I want to set my default priority to none/empty, but this is currently not possible because ConfigParser always reads a string and taskw only accepts ```'L'```, ```'M'```, ```'H'``` or the ```None``` literal.

I checked out the project and got it fixed, but before I can do a pull request, I am asking how do you prefer it to be fixed. This are the alternatives I can think of, from best to worse:

1. After reading the config file check if the default_priority is ```'None'``` as a string and change it to a literal. (Cons: doesn't feel like a clean solution)
2. Make ```None``` the ```default_priority``` instead of ```'M'```. (Cons: It will change the current expected behavior of bugwarrior)
3. Set the ```allow_no_value``` flag in ConfigParser. (Cons: it could make it difficult to spot and debug errors in the _bugwarriorrc_ file)

P.S. Thanks for bugwarrior it is really useful

ryneeverett:
Contrary to my apparent previous assumptions, configparser returns an
empty string rather than None when encountering an empty assignment,
making this a fairly trivial solution.

### DIFF
--- a/bugwarrior/config/schema.py
+++ b/bugwarrior/config/schema.py
@@ -308,9 +308,7 @@ class ServiceConfig(_ServiceConfig,  # type: ignore  # (dynamic base class)
     # Optional fields shared by all services.
     only_if_assigned: str = ''
     also_unassigned: bool = False
-    # Although upstream supports it, pydantic has problems with Literal[None].
-    default_priority: typing.Optional[
-        typing_extensions.Literal['L', 'M', 'H']] = 'M'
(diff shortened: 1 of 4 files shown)


fmauch:
I just gave this a test run and this seems to work perfectly fine (at least for my use case)

ryneeverett:
Thanks for the review!

PR #889, opened 2022-03-30
Fix priority property from redmine issues

hlobit:
Priorities could not be pulled from a redmine server until I lowercased the `name` property.
I then changed to a non-default priority in tests to ensure the result is as expected.

### DIFF
--- a/bugwarrior/services/redmine.py
+++ b/bugwarrior/services/redmine.py
@@ -203,7 +203,7 @@ class RedMineIssue(Issue):
 
     def get_priority(self):
         return self.PRIORITY_MAP.get(
-            self.record.get('priority', {}).get('Name'),
+            self.record.get('priority', {}).get('name'),
             self.origin['default_priority']
         )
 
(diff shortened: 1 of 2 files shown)


ryneeverett:
I'm not a redmine user myself. Care to test this out @CaptainQuirk?

ryneeverett (approved):

CaptainQuirk:
The fix definitively makes sense but unfortunately, I seem to hit several blocks here : 
- My redmine instance is in French, so the `PRIORITY_MAP` will not work as is
- If I change the mapping just to test it out, although debugging shows me that the proper taskwarrior priority is guessed from the redmine priority, the task priority remains incorrect

hlobit:
Hi @CaptainQuirk,

* yes ideally this should not rely on priority names or priority map should be configurable
* `priority` is by default a static field (I guess in most use cases the users want to manage it independently in taskwarrior)

With this config in `.config/bugwarrior/bugwarriorrc` I got it imported:

```
[general]
static_fields =
```

PR #891, opened 2022-04-02
Merge the union of tags.

### LINKED ISSUE #449: When multiple services match the same bug, add the union of their tags
Stealing an example from #447,
```
[general]
targets = edigiacomo_github, arpae_simc_github, r_map_github
inline_links = False
annotation_links = False
annotation_comments = False
legacy_matching = False
add_tags = github, personal

[arpae_simc_github]
service = github
only_if_assigned = edigiacomo
github.username = arpa-simc
github.login = edigiacomo
github.password = ...
github.include_repos = meteo-vm2, wreport, dballe, arkimet, simc-opendata-examples, arkimet-postprocessor-suite
add_tags = github, arpae_simc_github
```

An issue which matches both [general] and [arpae_simc_github] should get tagged with `github personal arpae_simc_github` (the union of the tags for the two services).

The descriptions will also differ; some stable method of selecting among the descriptions will need to be employed.

ryneeverett:
Resolve #449. 

When multiple targets pull in the same issue, the tags added to
taskwarrior should be a union of those of each target.

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -337,31 +337,37 @@ def synchronize(issue_generator, conf, main_section, dry_run=False):
                 continue
             else:
                 raise
+
+        # We received this issue from The Internet, but we're not sure what
+        # kind of encoding the service providers may have handed us. Let's try
+        # and decode all byte strings from UTF8 off the bat.  If we encounter
+        # other encodings in the wild in the future, we can revise the handling
+        # here. https://github.com/ralphbean/bugwarrior/issues/350
+        for key in issue_dict.keys():
+            if isinstance(issue_dict[key], bytes):
+                try:
+                    issue_dict[key] = issue_dict[key].decode('utf-8')
+                except UnicodeDecodeError:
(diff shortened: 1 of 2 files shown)


djmitche (approved):
This looks good, but doesn't have much context.  Is it addressing a specific issue?  Why are tags merged when other issue attributes are not?  What are the situations under which a service would generate multiple duplicate issues?

ryneeverett:
> Is it addressing a specific issue?

Yes, #449.You opened it which is why I requested your review.

> Why are tags merged when other issue attributes are not?

As I understand it the only reason merging is relevant with tags is that you can add tags with the `add_tags` configuration option which is why they could differ for the same issue amongst different targets. I'm not sure there are other attributes for which this issue applies.

> What are the situations under which a service would generate multiple duplicate issues?

Multiple targets utilizing the same service.

ryneeverett:
I copied the primary commit message into the description for context.

djmitche:
Ah, thanks!  I had forgotten I filed that!

PR #896, opened 2022-04-11
gitlab: When failing to connect do not pass

fmauch:
With query-based fetching it is possible to end up only in paths
that catch connection errors. While catching OSErrors there might be
a valid strategy given the points documented, this also catches failed
connection attempts, as ConnectionRefusedError inherits from OSError.

Without this, it is possible that tasks get closed when connection to a gitlab instance cannot be established (Though internet connection is available, e.g. the gitlab instance is behind a non-connected VPN.).

### DIFF
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -256,6 +256,7 @@ class GitlabClient(ServiceClient):
         try:
             result = self._fetch_paged(query)
         except ConnectionError as err:
+            # ConnectionError inherits from OSError and the service should fail without internet.
             raise ConnectionError(err)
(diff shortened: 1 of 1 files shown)


ryneeverett (approved):
Could you add a brief explanatory comment to each of these similar to those in the `except OSError` blocks? Something like "ConnectionError inherits from OSError and the service should fail without internet."

fmauch:
> Could you add a brief explanatory comment to each of these similar to those in the except OSError blocks? Something like "ConnectionError inherits from OSError and the service should fail without internet."

done

ryneeverett (approved):

PR #900, opened 2022-04-15
Fix error in legacy cli deprecation warning.

ryneeverett:
I hadn't properly tested subcommands which give a false positive.

The testing harness doesn't see the root command -- which is the reason
I poorly implemented the warning in the first place -- so I'm removing
the assertion. Under manual testing the correct warnings are logged.


srl295 (approved):
plausible, testing it now

srl295:
Fixes #901

ryneeverett:
Thanks for the review!

srl295:
Thanks for bugwarrior.

srl295:
I don’t know what I’d do (each day) without it. Literally !

PR #908, opened 2022-04-21
Per-service static_fields. Resolve #659.

### LINKED ISSUE #659: Per service static_fields
Hi,

From this talk: https://youtu.be/zl68asL9jZA?t=1377 (22:57) it appears you can set static_fields per service, though I don't see that in the documentation.

But, if I do the same it does not work.

The only place in the code I found usages of static_field is in db.py:syncronize though it appears it only uses that settings from [general] but not from the services definition. Am I correct?

Anyway, this is a feature request to be able to set static_fields per service instead of general.

thanks

ryneeverett:
Add a static_fields option to Common Service Configuration Options. This
enables users to differ behavior of core UDA names (e.g., "description")
amongst services.

The tests fail because we're currently able to test `db.synchronize` using regular dicts as issues whereas in a real bugwarrior pull these are `Issue` instances. This PR is awaiting a much-needed refactor of `db.synchronize` which will break this monstrous function into more testable components.

### DIFF
--- a/bugwarrior/collect.py
+++ b/bugwarrior/collect.py
@@ -94,7 +94,9 @@ def aggregate_issues(conf, main_section, debug):
     while currently_running > 0:
         issue = queue.get(True)
         try:
-            yield TaskConstructor(issue).get_taskwarrior_record()
+            record = TaskConstructor(issue).get_taskwarrior_record()
+            record['target'] = issue.config.target
+            yield record
         except AttributeError:
             if isinstance(issue, tuple):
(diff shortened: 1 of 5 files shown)


ryneeverett:
While restructuring `db.synchronize` would still be a nice step, I think a simpler architectural improvement which would also unblock this for testing purposes would be to make the `Issue` class private to the `services` subpackage which can return simple `dict`s instead. This not only makes testing easier but eliminates unnecessary coupling between `services` and `db`.

PR #931, opened 2022-05-20
Adding Jitamin support

gumulka:
I'm not sure, if this should be integrated upstream, as Jitamin is quite old, unmaintained and has no API, so this Plugin relies on web-scraping to get the needed information. 

But it is the task-management solution my employer uses, and I use taskwarrior, so I wrote this.

### DIFF
--- a/bugwarrior/docs/services/jitamin.rst
+++ b/bugwarrior/docs/services/jitamin.rst
@@ -1,0 +1,55 @@
+Jitamin
+=======
+
+You can import tasks from your Jitamin instance using
+the ``jitamin`` service name.
+
+Example Service
+---------------
+
+Here's an example of a Jitamin target::
+
+    [my_issue_tracker]
+    service = jitamin
+    jitamin.username = ralphbean
+    jitamin.password = OMG_LULZ
+    jitamin.base_uri = https://jitamin.com
+
+The above example is the minimum required to import tasks from
+Jitamin.  You can also feel free to use any of the
+configuration options described in :ref:`common_configuration_options`
+or described in `Service Features`_ below.
+
+Service Features
+----------------
+
+Use Column as tag
++++++++++++++++++
+
+If you want the column of the task to be a tag, you can do this with::
+
+    jitamin.column_as_tag = True
+
+Custom query strings
+++++++++++++++++++++
+
+By default, jitamin will search for all open tasks.
+
+You may override this query string through your `bugwarriorrc` file.
+
+For example::
+
+    jitamin.query = assignee:me project:"My project"
+
+
+Do Not Verify SSL Certificate
++++++++++++++++++++++++++++++
+
+If you want to ignore verifying the SSL certificate, set::
+
+    jitamin.verify_ssl = False
+
+Provided UDA Fields
+-------------------
+
+.. udas:: bugwarrior.services.jitamin.JitaminIssue
--- a/bugwarrior/services/jitamin.py
+++ b/bugwarrior/services/jitamin.py
@@ -47,9 +47,7 @@ class Task:
             return default
 
     def fetch_data(self) -> None:
-        response = self._instance.session.get(self._url)
-
-        soup = BeautifulSoup(response.content, "html.parser")
+        soup = self._instance.get_page_as_soup(self._url)
 
         self._data["title"] = soup.find("h2").text.strip()
 
@@ -125,13 +123,18 @@ class Jitamin:
     ):
         self.session = requests.session()
         self.base_uri = url
+        self._username = username
+        self._password = password
 
         if not verify_ssl:
             self.session.verify = False
             requests.packages.urllib3.disable_warnings()
 
-        response = self.session.get(self.base_uri + "/check")
-        if response.headers.get("Location") is None:
+
+    def get_page_as_soup(self, page: str) -> BeautifulSoup:
+        response = self.session.get(page)
+
+        if response.url.endswith("/login"):
             log.info("Logging into %s", self.base_uri)
             soup = BeautifulSoup(response.content, "html.parser")
             token = soup.find("input", attrs={"name": "csrf_token"})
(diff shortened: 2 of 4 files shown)


bugwarrior/services/jitamin.py
```diff
@@ -0,0 +1,240 @@
...
+from bs4 import BeautifulSoup
+
+from bugwarrior import config
+from bugwarrior.services import IssueService, Issue, ServiceClient
+
+log = logging.getLogger(__name__)
+
+class JitaminConfig(config.ServiceConfig, prefix="jitamin"):
+    service: typing_extensions.Literal["jitamin"]
+    base_uri: config.StrippedTrailingSlashUrl
```
ryneeverett: This type exists for backwards compatibility. I'd prefer to use one of the [upstream types](https://pydantic-docs.helpmanual.io/usage/types/#urls) such as `pydantic.AnyUrl` if possible.

bugwarrior/services/jitamin.py
```diff
@@ -0,0 +1,240 @@
...
+    column_as_tag: bool = False
+
+    query: str = ""
+
+
+class JitaminIssue(Issue):
+    URL = "jitaminurl"
+    POSITION = "jitaminposition"
+    COLUMN = "jitamincolumn"
+    DESCRIPTION = "jitamindescription"
```
ryneeverett, line 225: Description is a core attribute.[^1] Is there a reason we need a separate UDA for this?

[^1]: https://taskwarrior.org/docs/udas.html

bugwarrior/docs/services/jitamin.rst
```diff
@@ -0,0 +1,55 @@
+Jitamin
+=======
+
+You can import tasks from your Jitamin instance using
+the ``jitamin`` service name.
```
ryneeverett, line 5: This would be a good place to note how to install the extra dependencies.

bugwarrior/services/jitamin.py
```diff
@@ -0,0 +1,240 @@
...
+    verify_ssl: bool = True
+    column_as_tag: bool = False
+
+    query: str = ""
+
+
+class JitaminIssue(Issue):
+    URL = "jitaminurl"
+    POSITION = "jitaminposition"
+    COLUMN = "jitamincolumn"
```
ryneeverett, line 224: It seems a little odd to have column as a separate UDA and also optionally use it as a tag. Maybe it makes sense to Jitamin users though.

bugwarrior/services/jitamin.py
```diff
@@ -0,0 +1,240 @@
...
+                continue
+            link_list.append(link)
+        return link_list
+
+    def get_task_list(self):
+        i = 1
+        while True:
+            tasklist = self.get_task_list_page(
+                "?controller=SearchController&action=index&q="
+                + self.config.query + " status:open"
```
ryneeverett: We typically implement the `query` option to be as open ended as possible, which usually means it would be an alternative to the default query string rather than added to it. For example, it appears that one would not be able to query for closed issues the way this is implemented.

ryneeverett (commented):
I don't have a problem with merging support for this service if people are still using it.

Aside from the inline comments, I'd like to see at least a minimal test module before to merging this. It may help to pull out the requests and beautiful soup logic into a separate "client" class. This isn't a dealbreaker but it provides separation of concerns and might make mocking easier.

bugwarrior/services/jitamin.py
gumulka: I first had it as an open query, but then seen, that closed issues are reopend by bugwarrior per default (see https://github.com/ralphbean/bugwarrior/blob/develop/bugwarrior/db.py#L380), only to then be closed by taskwarrior, as they have an ending, which lead to a lot of lines like this:

```
INFO:bugwarrior.db:Updating task c9de8aa8-c329-47f7-89b3-540396fdb2cc, (bw)# - root Passwort not changable vial co; status: 'completed' -> 'pending'; end: datetime.datetime(2022, 2, 25, 12, 34, tzinfo=tzutc()) -> datetime.datetime(2022, 2, 25, 12, 34, tzinfo=<UTC>)
```
every time I do `bugwarrior pull`

So instead of filling the end field, I decided to skip all closed tasks.

bugwarrior/services/jitamin.py
gumulka: No, it was more of a personal preference. I like to see the column in my default task view, which does not include the UDA, so tag seemed to be the first logical value.

bugwarrior/services/jitamin.py
gumulka: Jitamin task consists of a short title and a long description. Kind of like Gitlab issues. I use the title for the UDA description and create my own UDA description for the often long multiline description.

I could opt to make it an annotation, but they also do not feel so good to me. Open to discuss here.

gumulka:
Hey, I "moved" the Jitamin logic to a new class. Will move the new class to a new project later on, but just wanted some feedback if that is what you had in mind.

Will add tests later this week.

ryneeverett:
> Hey, I "moved" the Jitamin logic to a new class. Will move the new class to a new project later on, but just wanted some feedback if that is what you had in mind.

Yes, the way the client class abstracts away the fetching/parsing from the service class is exactly what I had in mind.

I understand why you added the lazy `Task` class but I think it's somewhat confusing given the conceptual overlap with the `Issue` class. We don't have anything like this in any other bugwarrior service so there's no precedent, but I think the terminology might match up better if `Task` was renamed `Record` since that aligns with the fact that it represents remote data prior to the conversion into local taskwarrior data (akin to `Issue.record`). What do you think?

gumulka:
The week after I created the pull request. My employer decided to ditch Jitamin and move to Jira. Since this is now a permanent decision and I won't be using this anymore and will not be able to fix any bugs, I'm closing this merge request.

PR #948, opened 2022-07-07
feat: macOS applescript notifications

### LINKED ISSUE #875: growlnotify is discontinued
re: #59 

https://growl.github.io/growl/ says

> Growl is a retired project, we couldn’t think of another thing to change which would be substantial enough to bring out a new updated release. Growl is stable and should work for as long as intel based programs work. Anyone who wants to run Growl is free to do so in an unsupported fashion.

Ideally switch to use the macos native notify…

nekowinston:
Would close #875.
`pytest` failed for me locally, even before adding the feature, unsure why, exactly.


bugwarrior/notifications.py
```diff
@@ -91,6 +91,36 @@ def send_notification(issue, op, conf):
             priority=1,
         )
         return
+    elif notify_backend == 'applescript':
+        import subprocess
```
ryneeverett: This import should be at the beginning of the module. The other imports in this function are of optional dependencies which otherwise might not be installed, but `subprocess` is part of the standard library so there's no harm in anyone importing it.

bugwarrior/notifications.py
```diff
@@ -91,6 +91,36 @@ def send_notification(issue, op, conf):
...
+                        description
+                    ),
+                ]
+            )
+            return
+
+        message = '%s task: %s' % (op, issue['description'])
+        metadata = _get_metadata(issue)
+        if metadata is not None:
+            message += metadata
```
ryneeverett: It would be nice to move this block making `message` to above the if-statement to avoid duplicating it.

ryneeverett:
We should add a deprecation warning to growlnotify before closing #875. However, that doesn't necessarily need to be done in this pull request.

nekowinston:
Thank you for the review comments, I'll update the PR accordingly.

ryneeverett:
@srl295 I don't have a mac to test this on. Can you test it out?

*(Note that I will eventually merge this even if nobody tests it.)*

bugwarrior/notifications.py
```diff
@@ -91,6 +93,30 @@ def send_notification(issue, op, conf):
...
+            return
+
+        subprocess.call(
+            [
+                'osascript',
+                '-e',
+                'display notification "{}" with title "Bugwarrior"'.format(message),
+            ]
+        )
+        return
```
ryneeverett, line 105: Since this implementation is simpler than growlnotify, the logic could be simpler too. E.g.:

```py
description = ...
notification = description if op == 'bw_finished' else message
subprocess.call(...
    "...".format(
        notification
...
```

ryneeverett (approved):

bugwarrior/notifications.py
nekowinston: Good catch, thanks.

nekowinston:
Here's a screenshot of the notifications for the current implementation, after the reviews.
bugwarrior-applescript

ryneeverett (approved):

bugwarrior/notifications.py
```diff
@@ -91,6 +93,15 @@ def send_notification(issue, op, conf):
             priority=1,
         )
         return
+    elif notify_backend == 'applescript':
+        description = 'Finished querying for new issues.\n{}'.format(issue['description'])
+        notification = description if op == 'bw_finished' else message
+        subprocess.call([
+            'osascript',
+            '-e',
+            'display notification "{}" with title "Bugwarrior"'.format(notification)
```
srl295: Should it escape? What happens if message is:  `Added task: Practice my "air quotes"`

then:

```shell
$ osascript -e 'display notification "Added task: Practice my "air quotes"" with title "Bugwarrior"'
46:50: syntax error: A identifier can’t go after this “"”. (-2740)
```

srl295 (approved):
LGTM with a concern, did not try it yet

bugwarrior/notifications.py
ryneeverett: Nice catch. I believe we want [shlex.quote](https://docs.python.org/3/library/shlex.html#shlex.quote).

nekowinston: Oh this is embarrassing. I'm on it.

nekowinston: `shlex.quote()` caused troubles when being passed into `osascript` on my end, since it always defaulted to single quotes...

```python
# works
subprocess.call(['osascript', '-e', 'display notification "{}" with title "bugwarrior"'.format('Test " Notification'.replace('"', '\\"'))])
# fails
subprocess.call(['osascript', '-e', 'display notification "{}" with title "bugwarrior"'.format(shlex.quote('Test " Notification'))])
```

Apologies if I'm missing something obvious again.

srl295 (approved):
There could be other edge cases but this covers the obvious ones

ryneeverett (approved):

ryneeverett:
Thank you both for your work on this!

PR #954, opened 2022-08-24
Adds support for Nextcloud Deck, a Kanban-Style task management tool …

lenalebt:
…integrated with Nextcloud

It already works more or less, but I'm not very used to writing python code. I'm open for suggestions on how to improve stuff :-). I think functionality-wise it should all be there already.

A bit of context: I'm using nextcloud deck for managing stuff in the household together with my wife, but I'd like to get those tasks into my taskwarrior as well ;-). I currently plan to do this regularly in the upcoming time.

### DIFF
--- a/bugwarrior/docs/services/deck.rst
+++ b/bugwarrior/docs/services/deck.rst
@@ -1,0 +1,73 @@
+Nextcloud Deck
+==============
+
+You can import tasks from your Nextcloud Deck instance using the ``deck`` service name.
+
+This service requires `deck <https://github.com/nextcloud/deck#installationupdate>`_ to be installed in a Nextcloud installation and available to bugwarrior.
+
+Example Service
+---------------
+
+Here's an example of a Nextcloud Deck target::
+
+   [my_nextcloud_deck]
+   service = deck
+   deck.base_uri = https://nextcloud.example.com
+   deck.username = my_user
+   deck.password = my-nextcloud-app-password
+
+The above example is the minimum required to import issues from Deck.  You can also feel free to use any of the configuration options described in :ref:`common_configuration_options` or described in `Service Features`_ below.
+
+The ``base_uri`` should point to the location of your Nextcloud installation.
+``username`` should be the username of the user that has access to the Deck boards you are interested in.
+``password`` should be an app-password (or the real password, although that is not recommended) of the user.
+
+Service Features
+----------------
+
+Import Labels as Tags
++++++++++++++++++++++
+
+The Deck allows you to attach labels to cards. To use those labels as tags, you can use the ``deck.import_labels_as_tags`` option::
+
+    deck.import_labels_as_tags = True
+
+Also, if you would like to control how these labels are created, you can specify a template used for converting the Deck label into a Taskwarrior tag.
+
+For example, to prefix all incoming labels with the string 'deck\_' (perhaps to differentiate them from any existing tags you might have), you could add the following configuration option::
+
+    deck.label_template = deck_{{label}}
+
+In addition to the context variable ``{{label}}``, you also have access
+to all fields on the Taskwarrior task if needed.
+
+.. note::
+
+   See :ref:`field_templates` for more details regarding how templates
+   are processed.
+
+Only if assigned
+++++++++++++++++
+
+You can filter for cards assigned to a specific user::
+
+    deck.only_if_assigned = my_user
+
+``only_if_assigned`` can be set to a username, then only issues assigned to that user will be taken into account. Attention, deck can assign multiple users, and this plugin has no support for multiple assignees - it will always only take the first assignee into account.
+
+Exclude / include board IDs
++++++++++++++++++++++++++++
+
+You can explicitly exclude or include specific boards::
+
+    deck.exclude_board_ids = 5,6
+    deck.include_board_ids = 4
+
+``exclude_board_ids`` can explicitly exclude specific boards from importing;
+``include_board_ids`` can explicitly include specific boards from importing.
+Use them to filter what cards you want to have imported.
+
+Provided UDA Fields
+-------------------
+
+.. udas:: bugwarrior.services.deck.NextcloudDeckIssue
--- a/bugwarrior/docs/services/deck.rst
+++ b/bugwarrior/docs/services/deck.rst
@@ -53,7 +53,8 @@ You can filter for cards assigned to a specific user::
 
     deck.only_if_assigned = my_user
 
-``only_if_assigned`` can be set to a username, then only issues assigned to that user will be taken into account. Attention, deck can assign multiple users, and this plugin has no support for multiple assignees - it will always only take the first assignee into account.
+.. note::
+    ``only_if_assigned`` can be set to a username, then only issues assigned to that user will be taken into account. Attention, deck can assign multiple users, and this plugin has no support for multiple assignees - it will always only take the first assignee into account.
 
 Exclude / include board IDs
 +++++++++++++++++++++++++++
@@ -63,9 +64,11 @@ You can explicitly exclude or include specific boards::
     deck.exclude_board_ids = 5,6
     deck.include_board_ids = 4
 
-``exclude_board_ids`` can explicitly exclude specific boards from importing;
-``include_board_ids`` can explicitly include specific boards from importing.
-Use them to filter what cards you want to have imported.
+.. note::
+    ``include_board_ids`` can explicitly include specific boards from importing;
+    ``exclude_board_ids`` can explicitly exclude specific boards from importing.
+    Use them to filter what cards you want to have imported.
+    If both are defined, ``include_board_ids`` will be used.
 
 Provided UDA Fields
 -------------------
--- a/bugwarrior/services/deck.py
+++ b/bugwarrior/services/deck.py
@@ -53,7 +53,9 @@ class NextcloudDeckClient(ServiceClient):
 
     # see https://deck.readthedocs.io/en/latest/API/#stacks for API docs
     def get_stacks(self, board_id):
-        response = self.session.get(f'{self.base_uri}/{self.api_base_path}/boards/{board_id}/stacks')
+        response = self.session.get(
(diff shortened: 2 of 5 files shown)


bugwarrior/docs/services/deck.rst
```diff
@@ -0,0 +1,73 @@
...
+   are processed.
+
+Only if assigned
+++++++++++++++++
+
+You can filter for cards assigned to a specific user::
+
+    deck.only_if_assigned = my_user
+
+``only_if_assigned`` can be set to a username, then only issues assigned to that user will be taken into account. Attention, deck can assign multiple users, and this plugin has no support for multiple assignees - it will always only take the first assignee into account.
```
ryneeverett: I would move this "Attention" sentence into [a note](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html) using `..note::`.

bugwarrior/services/deck.py
```diff
@@ -0,0 +1,184 @@
...
+log = logging.getLogger(__name__)
+
+
+class NextcloudDeckConfig(config.ServiceConfig, prefix='deck'):
+    service: typing_extensions.Literal['deck']
+    base_uri: pydantic.AnyUrl
+    username: str
+
+    # can be a password or an app-password
+    password: str = ''
```
ryneeverett: Regardless of the type of password, isn't it a required field? If so a default value should not be specified.

bugwarrior/services/deck.py
```diff
@@ -0,0 +1,184 @@
...
+
+from bugwarrior import config
+from bugwarrior.services import IssueService, Issue, ServiceClient
+
+log = logging.getLogger(__name__)
+
+
+class NextcloudDeckConfig(config.ServiceConfig, prefix='deck'):
+    service: typing_extensions.Literal['deck']
+    base_uri: pydantic.AnyUrl
```
ryneeverett: I suspect we should use `config.StrippedTrailingSlashUrl` here so if somebody adds an extra slash at the end in their config it doesn't break url construction.

bugwarrior/services/deck.py
```diff
@@ -0,0 +1,184 @@
...
+            self.AUTHOR: self.record['owner']['uid'],
+            self.BOARD_ID: self.extra['board']['id'],
+            self.BOARD_TITLE: self.extra['board']['title'],
+            self.STACK_ID: self.extra['stack']['id'],
+            self.STACK_TITLE: self.extra['stack']['title'],
+            self.CARD_ID: self.record['id'],
+            self.TITLE: self.record['title'],
+            self.DESCRIPTION: self.record['description'],
+            self.ORDER: self.record['order'],
+            self.ASSIGNEE: ([assignee['participant']['uid'] for assignee in self.record['assignedUsers']] or [None])[0],
```
ryneeverett: I think it would be more clear to express this as `self.record['assignedUsers'][0]['participant']['uid'] if self.record['assignedUsers'] else None`

bugwarrior/services/deck.py
```diff
@@ -0,0 +1,184 @@
...
+            'only_if_assigned': self.config.only_if_assigned,
+        }
+
+    def get_owner(self, issue):
+        return issue[issue.ASSIGNEE]
+
+    def issues(self):
+        for board in self.client.get_boards():
+            if ((not self.config.include_board_ids) or (str(board['id']) in self.config.include_board_ids)) and \
+                    ((not self.config.exclude_board_ids) or (str(board['id']) not in self.config.exclude_board_ids)):
```
ryneeverett: I would like to see this complex logic replaced with a filter function. Many of the services have a method called `filter_repos` you could see for reference. (You'd probably call yours `filter_boards`.)

lenalebt:
Thanks for the review, highly appreciated and very good points raised :smile: !

I would love to test the `filter_repos`, but had some trouble setting the config values needed for that. Is there an easy way to do so? Example code:

    def test_filter_repos_include(self):
        self.service.config['deck.include_board_ids'] = ['5']
        self.assertTrue(self.service.filter_repos({'title': 'testboard', 'id': 5}))
        self.assertFalse(self.service.filter_repos({'title': 'testboard', 'id': 6}))

    def test_filter_repos_exclude(self):
        self.service.config['deck.exclude_board_ids'] = ['5']
        self.assertFalse(self.service.filter_repos({'title': 'testboard', 'id': 5}))
        self.assertTrue(self.service.filter_repos({'title': 'testboard', 'id': 6}))

I think I'm just holding it wrong. It's not that I think those are absolutely crucial tests, but it would be nice to have it working somehow.

ryneeverett (approved):

ryneeverett:
> I would love to test the `filter_repos`, but had some trouble setting the config values needed for that. Is there an easy way to do so?

We don't seem to have the test infrastructure to do this as trivially as would be ideal, but you can check out `test_gitlab.py:TestGitlabService` for an example. If you copied the `setUp` method and altered it to have the minimum configuration for deck, you could then test the configurations with:

```py
def test_filter_repos_include(self):
    self.config.set('myservice', 'deck.include_board_ids', '5')
    self.assertTrue(self.service.filter_repos({'title': 'testboard', 'id': 5}))
    self.assertFalse(self.service.filter_repos({'title': 'testboard', 'id': 6}))

def test_filter_repos_exclude(self):
    self.config.set('myservice', 'deck.exclude_board_ids', '5')
    self.assertFalse(self.service.filter_repos({'title': 'testboard', 'id': 5}))
    self.assertTrue(self.service.filter_repos({'title': 'testboard', 'id': 6}))
```

lenalebt:
I tried doing that, but I had some trouble initially with the setup because all the mocking works differently over there. I worked it out - thanks for your help :-).

Besides that I fixed the validation problems around line lengths. I'm happy with it now, if you're fine with it it can be merged.

bugwarrior/services/deck.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # can be a password or an app-password
+    password: str
+
+    include_board_ids: config.ConfigList = config.ConfigList([])
+    exclude_board_ids: config.ConfigList = config.ConfigList([])
+
+    import_labels_as_tags: bool = False
+    label_template: str = '{{label}}'
+    only_if_assigned: str = None
```
ryneeverett: I believe you should be able to remove the `only_if_assigned` field without consequence. It is defined in `ServiceConfig` from which this class inherits.

bugwarrior/services/deck.py
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+
+    UNIQUE_KEY = (BOARD_ID, STACK_ID, CARD_ID,)
+
+    def to_taskwarrior(self):
+        return {
+            'project': self.extra['board']['title'].lower().replace(' ', '_'),
+            'priority': self.get_priority(),
+            # 'annotations': self.record.get('annotations', []),
```
ryneeverett: Were you planning to complete annotations support? I believe every other service currently has this feature.

ryneeverett (approved):

bugwarrior/services/deck.py
lenalebt: Correct, thanks!

bugwarrior/services/deck.py
lenalebt: I did not initially, but I just added it. It does slow down the process significantly, though, because there is a specific call needed per card / issue. For my cases it's about factor 10 (20 seconds instead of 2). As far as I can see it, it's not possible to skip annotation importing via configuration somehow, correct? How would you handle that - it annoys me quite a lot if it slows down the process considerably, but I also see value in completeness.

ryneeverett: This is typically handled by setting the main configuration option "annotation_comments" to `False`. When this is set `IssueService.build_annotations` just throws the annotations away, so several services have implemented logic to avoid calling the `client.get_comments` method when when this is the case. (You can just pass an empty list to `build_annotations`. You can see the github and gitlab services for examples.)

I don't believe we have any way to disable annotations on an individual service basis or that anyone has requested that.

ryneeverett (approved):

bugwarrior/services/deck.py
```diff
@@ -0,0 +1,207 @@
...
+    def get_stacks(self, board_id):
+        response = self.session.get(
+            f'{self.base_uri}/{self.api_base_path}/boards/{board_id}/stacks'
+        )
+        return response.json()
+
+    # see https://deck.readthedocs.io/en/latest/API/#comments for API docs
+    def get_comments(self, card_id):
+        response = self.session.get(
+            f'{self.base_uri}/ocs/v2.php/apps/deck/api/v1.0/cards/{card_id}/comments?limit=100&offset=0'
```
ryneeverett: Looks like you left out `self.api_base_path`. I would combine them in the `__init__` method. (`self.api_base_path = f'{base_uri/index.php/apps/deck/api/v1.0'`) Note that there's also no need to assign any of the `__init__` arguments to `self` if you take this approach as they're only used in the `__init__` method.

ryneeverett (changes requested):

bugwarrior/services/deck.py
ryneeverett: Oh nevermind, i missed that it is a totally different api path.

ryneeverett (approved):

ryneeverett:
@lenalebt I hope you don't mind I pushed a couple commits to your branch to fix CI and implement conditional annotation fetching. Setting `annotation_comments = False` in the `[general]` section of your bugwarriorrc should satisfy your needs.

lenalebt:
Oh wow, no problem, thank you! I had a little bit too much on my table lately.

So, I'm fine with merging now, I'm using it on a regular basis and don't see problems so far. Anything I can do now in addition to get it merged :-)?

ryneeverett:
> So, I'm fine with merging now, I'm using it on a regular basis and don't see problems so far. Anything I can do now in addition to get it merged :-)?

Nope, I was just waiting for your approval of the changes. :)

PR #962, opened 2022-11-11
Improve doc for obtaining "trello.token"

leandrochiarini:
- Added an alternative method and an image to obtain trello token.


ryneeverett:
Thanks! But the rst refers to `trello_url.png` and the image is named `trello_token.png`.

jhutar:
Maybe also change `toke` to `token`? Hopefully I'm not missing anything.

ryneeverett:
Merged in #992.

PR #963, opened 2023-01-14
Added filter option "only_if_reaction" for gitlab

ForsakenNGS:
I want to use bugwarrior / taskwarrior to keep track of issues for open-source projects I am contributing (reports) or simply watching, which is quite difficult with the current implementation. Those issues are neither created by me, neither assigned to me and I (obviously) don't want to import all issues.

From my research the gitlab search parameter `my_reaction_emoji` is the best option for that. This way you can just thumbsup all issues that are relevant for you and have only those imported/updated.

### DIFF
--- a/bugwarrior/docs/services/gitlab.rst
+++ b/bugwarrior/docs/services/gitlab.rst
@@ -140,6 +140,16 @@ Issues are included by default, if not configured otherwise. To disable querying
 
     gitlab.include_issues = False
 
+Filtering Issues by reaction
+++++++++++++++++++++++++++++
+
+If you want to keep track of issues you contributed to (which you did not create) 
+you can set the gitlab filter `my_reaction_emoji` with the following config::
+
+    gitlab.only_if_reaction = Any
+
+This way you can mark relevant issues by just reacting e.g. with a thumbsup and have them imported automatically.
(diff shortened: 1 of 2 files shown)


ryneeverett:
Great idea!

However, I don't think adding this configuration option is necessary. If I'm not mistaken you can achieve the same outcome with `gitlab.issue_query = issues?state=opened&my_reaction_emoji=Any`.

There's a maintenance cost for every configuration option we add and trying to ensure that all the options play nicely together becomes impossible. IMO we should ultimately deprecate all the filtering options in favor of user-specified query strings, but at the very least we should only add more filtering options due to popular demand or if it isn't possible to achieve the same outcome with the current configuration options.

However, I think the alternative to more options is better documentation of "recipes" such as this one. At the very least I'd like to see this added to the example in the "Custom query strings" section.

ForsakenNGS:
> If I'm not mistaken you can achieve the same outcome with gitlab.issue_query = issues?state=opened&my_reaction_emoji=Any.

I was looking for such an option and didn't see it. If that's true I'll make a new pull request for adding the proper documentation ;-)

ryneeverett:
Closing as stale.

PR #967, opened 2023-01-30
Handle redmine not returning estimated hours.

fedegiova:
In our setup the estimated_hours and spent_hours fields might be empty, this trigger an exception in taskwarrior that it cannot convert 0.0 to a duration.

ERROR:bugwarrior.db:Unable to add task: b"The duration value '0.0' is not supported." Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/bugwarrior/db.py", line 439, in synchronize
    new_task = tw.task_add(**issue)
  File "/usr/lib/python3.10/site-packages/taskw/warrior.py", line 704, in task_add
    stdout, stderr = self._execute('add', *args)
  File "/usr/lib/python3.10/site-packages/taskw/warrior.py", line 498, in _execute

The command line for taskwarrior invocation is:

[b'task'
...
b'rc.uda.redminespenthours.type=duration', b'rc.uda.redminespenthours.label="Redmine Spent Hours"', b'rc.uda.redmineestimatedhours.type=duration', ...
b'redminespenthours:"0.0"'
]

The patch makes sure that the value is formatted correctly for the duration type

### DIFF
--- a/bugwarrior/services/redmine.py
+++ b/bugwarrior/services/redmine.py
@@ -175,14 +175,12 @@ class RedMineIssue(Issue):
             updated_on = self.parse_date(updated_on).replace(microsecond=0)
         if created_on:
             created_on = self.parse_date(created_on).replace(microsecond=0)
-        if not spent_hours:
-            spent_hours = 0
-        spent_hours = str(spent_hours) + ' hours'
-        spent_hours = self.get_converted_hours(spent_hours)
-        if not estimated_hours:
-            estimated_hours = 0
-        estimated_hours = str(estimated_hours) + ' hours'
-        estimated_hours = self.get_converted_hours(estimated_hours)
+        if spent_hours or spent_hours == 0.0:
+            spent_hours = str(spent_hours) + ' hours'
+            spent_hours = self.get_converted_hours(spent_hours)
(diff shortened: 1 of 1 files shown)


ryneeverett (changes requested):
To clarify my understanding of the bug: because `0.0` is falsey, it was getting passed through as a `float` rather than getting properly formatted.

The tests fail because this patch changes the behavior to return a formatted `0` rather than `None` when these values are falsey. This would change the end behavior by inserting the formatted `0` value into every redmine task rather than omitting it when the server doesn't return it. I think a better solution (which would also fix the tests) would be to explicitly handle this condition, i.e.:

```py
if spent_hours or spent_hours == 0.0:
```

This has the same result as your patch in the `0.0` case because `self.get_converted_hours('0.0 hours')` will call `task calc 0.0 hours` which returns `PT0S`.

fedegiova:
You're right, I'll test your fix in  a couple of days and I'll upload a new patch if it works.

Thanks
Federico

fedegiova:
Hi! I've tested your solution and it's working fine, it fix the previous problem and also correctly update the task database.

I've reverted mine and applied your modifications.

Regards
Federico

ryneeverett (approved):

Issue #873, opened 2022-02-22
TOML Configuration


ryneeverett:
Now that configuration parsing is decoupled from the services (#849), it should be fairly easy to add a second configuration parser for a more modern format. I believe TOML is currently the best candidate and thanks to [PEP-518](https://www.python.org/dev/peps/pep-0518/#toml) has a special place in the python ecosystem. *(Edit 2023: Now that much of the taskwarrior code base itself is transitioning to rust, toml seems even more at home in the taskwarrior ecosystem.)*

My goal would be to add support for toml, deprecate the INI format, and convert most of the documentation over while leaving a separate page or section to explain how the legacy format works. I don't foresee a need to remove INI support. In fact it will probably help us keep configuration parsing decoupled from validation.

PR #973, opened 2023-02-22
Toml Configuration

ryneeverett:
<s>*Based on #980 and #981.*</s>

The overall goal was to complete #873, but I ended up carrying it further to the point of deprecating ini format in favor of toml. My main reason for wanting to do so is that I don't think there's currently any point to the service "prefixes", it introduces a lot of extra hacky meta-programming that is otherwise unneeded, and changing configuration file formats seems like a good opportunity to remove them. While this branch should be backwards compatible with INI format, the ini documentation and validation is severely degraded. I don't think it's practical to maintain backwards compatibility with first-class support, so better to just encourage everyone to jump to toml. Fortunately this is really easy with the `bugwarrior ini2toml` subcommand that this branch adds.


ryneeverett:
Closing in favor of #983.

Issue #836, opened 2021-08-30
Youtrack new API


moorchegue:
Youtrack integration is broken by these API changes:

https://blog.jetbrains.com/youtrack/2021/02/discontinuing-the-legacy-rest-api-action-required/

Attempt to log in fails with:

```
ERROR:bugwarrior.services:Worker for [my_youtrack] failed: YouTrack responded with <Response [400]>
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/bugwarrior/services/__init__.py", line 502, in _aggregate_issues
    service = get_service(service_name)(conf, main_section, target)
  File "/usr/lib/python3.9/site-packages/bugwarrior/services/youtrack.py", line 145, in __init__
    self._login(login, password)
  File "/usr/lib/python3.9/site-packages/bugwarrior/services/youtrack.py", line 162, in _login
    raise RuntimeError("YouTrack responded with %s" % resp)
RuntimeError: YouTrack responded with <Response [400]>
```

Response from the server looks like this:
```
{
    "error": "Deprecated REST API is disabled"
    "error_description": "https://blog.jetbrains.com/youtrack/2021/02/discontinuing-the-legacy-rest-api-action-required"
}
```

Looks like some work is required to make it compatible with the new API.

ryneeverett:
cc @babelop @dlobue

ryneeverett:
triage: We should probably remove the service before the next release if nobody steps up to fix it. On the one hand the service is a thin implementation so it doesn't feel like a huge loss. On the other hand it looks like the only changes needed would be to update the base url endpoint from /rest to /api and replace the basic auth endpoint with an oauth handshake. However we'd need a youtrack user to do this work.

peterdb:
Since I'm a user of both bugwarrior and youtrack, I'm willing to work on this issue.
I made some changes, and it seems to be working.
I'll create a fork, and create a pull request for reviewing.

ryneeverett:
Resolved via #982.

PR #982, opened 2023-03-15
[WIP] youtrack api changes

peterdb:


peterdb:
pull request for #836

bugwarrior/services/youtrack.py
```diff
@@ -133,24 +123,16 @@ def __init__(self, *args, **kw):
...
-        if not self.config.anonymous:
-            self._login(self.config.login, password)
-
-    def _login(self, login, password):
-        params = {'login': login, 'password': password}
-        resp = self.session.post(self.rest_url + "/user/login", params)
-        if resp.status_code != 200:
-            raise RuntimeError("YouTrack responded with %s" % resp)
-        self.session.headers['Cookie'] = resp.headers['set-cookie']
+        token = self.config.token
```
ryneeverett: We should use `self.get_password` like was previously done, which enables users to use [oracles](https://bugwarrior-docs.readthedocs.io/en/latest/common_configuration.html#password-management) to inject their password as an alternative to storing it in the configuration file, which would look like:

```py
token = self.get_password('token', self.config.login)
```

ryneeverett (approved):

bugwarrior/services/youtrack.py
peterdb: Ok, I will make the necessary changes

PR #993, opened 2023-04-04
bz: Unify datetime handling

### LINKED ISSUE #878: bugzilla: bz.py assumes datetime fields come from XML-RPC client and have a timetuple() method
Using the bugzilla.mozilla.org server (which only supports REST) I'm seeing errors like the following:
```
ERROR:bugwarrior.services:Worker for [bugzilla_mozilla_org] failed: 'str' object has no attribute 'timetuple'
  File "/home/visbrero/.local/lib/python3.8/site-packages/bugwarrior-1.8.0-py3.8.egg/bugwarrior/services/__init__.py", line 450, in _aggregate_issues
  File "/home/visbrero/.local/lib/python3.8/site-packages/bugwarrior-1.8.0-py3.8.egg/bugwarrior/services/bz.py", line 284, in issues
    extra['assigned_on'] = self._get_assigned_date(issue)
  File "/home/visbrero/.local/lib/python3.8/site-packages/bugwarrior-1.8.0-py3.8.egg/bugwarrior/services/bz.py", line 306, in _get_assigned_date
    time.mktime(assigned_date.timetuple()))
AttributeError: 'str' object has no attribute 'timetuple'
```

I believe this is because the [xmlrpc.client](https://docs.python.org/3/library/xmlrpc.client.html) explicitly defines its own DateTime ([doc](https://docs.python.org/3/library/xmlrpc.client.html#xmlrpc.client.DateTime)) ([source](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/xmlrpc/client.py#L296)) that explicitly provides a [timetuple](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/xmlrpc/client.py#L356-L357), but the REST layer does not perform this same kind of normalization.

Notably, python-bugzilla's CLI [performs a normalization](https://github.com/python-bugzilla/python-bugzilla/blob/a7c324041175a4157823bc2332a046cc2a54d105/bugzilla/_cli.py#L702-L706) that relies on the xmlrpc DateTime [str](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/xmlrpc/client.py#L359-L365) implementation that exposes the received (ISO8601) string in order to make the XML-RPC output look like the REST output.  If a string wasn't received by the xmlrpc layer, xmlrpc uses its own [_strftime](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/xmlrpc/client.py#L285-L294) method which returns an ISO 8601 style date and time but lacking any timezone information.

The bugzilla REST API docs [define datetime](https://bugzilla.readthedocs.io/en/latest/api/core/v1/general.html#common-data-types) as "A date/time. Timezone should be in UTC unless otherwise noted. Example format: `YYYY-MM-DDTHH24:MI:SSZ`" which I believe is an ISO 8601 date and time in UTC.

Since both call-sites in bz.py seem to be attempting to convert the xmlrpc DateTime to a UTC ISO8601, it seems like the fix would be to add a helper method to normalize.  Unless there's an intentional timezone transformation happening (beyond just branding the timezone as UTC), it sorta seems like if the maybe-date object is a string it can be passed through as-is, and if not, `str()` can be called on it and a "Z" tacked on the end.

khardix:
Depending on whether the data comes from XMLRPC or REST API, the timestamps are in different format.

This change puts the datetime conversion into dedicated function, enhancing it with support for multiple source types. See docstring for details.

Fixes: ralphbean/bugwarrior#878

### DIFF
--- a/bugwarrior/services/bz.py
+++ b/bugwarrior/services/bz.py
@@ -3,6 +3,7 @@ import logging
 import urllib.parse
 import time
 import typing
+import xmlrpc.client
 
 import bugzilla
 import pydantic
@@ -294,11 +295,7 @@ class BugzillaService(IssueService):
             )]
             if needinfos:
                 last_mod = needinfos[0]['modification_date']
-                # convert from RPC DateTime string to datetime.datetime object
-                mod_date = datetime.datetime.fromtimestamp(
-                    time.mktime(last_mod.timetuple()))
-
-                extra['needinfo_since'] = pytz.UTC.localize(mod_date).isoformat()
+                extra['needinfo_since'] = _ensure_datetime(last_mod).isoformat()
 
             if issue['status'] == 'ASSIGNED':
                 extra['assigned_on'] = self._get_assigned_date(issue)
@@ -309,8 +306,6 @@ class BugzillaService(IssueService):
             yield issue_obj
 
     def _get_assigned_date(self, issue):
-        assigned_date = None
-
         bug = self.bz.getbug(issue['id'])
         history = bug.get_history_raw()['bugs'][0]['history']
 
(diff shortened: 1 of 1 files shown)


ryneeverett (approved):
Looks good, thanks. Just need to remove that `assigned_date = None` assignment -- which was already a no-op -- to fix the flake8 error and get CI passing.

khardix:
Removed as requested :)

PR #995, opened 2023-04-26
Fix: Add a size limit to gitlab descriptions to prevent 'Argument list too long' OsError

DyspC:
Hello, 

I'm installing bugwarrior on a new machine and am hitting a hard limit on the `task` invocation because some of my assigned gitlab issues have markdown tables / long checklists

I don't really get where the limit is because on my machine `getconf ARG_MAX` tells `2097152` but the error happened with a line length of ~180000

This proposition adds a constant limit (arbitrarly set to 10000) and cuts gitlab descriptions when they exceed this limit, along with a warning line
```
INFO:bugwarrior.services:Done aggregating remote issues.
WARNING:bugwarrior.db:Description exceeded max size of 10000, was 166179
WARNING:bugwarrior.db:Description exceeded max size of 10000, was 11110
```

The same logic should probably be applied to other services (or even all fields) but I don't have enough insights so I kept the impacts to a minimum

### DIFF
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -3,6 +3,7 @@ import requests
 import typing
 
 import pydantic
+import sys
 import typing_extensions
 
 from bugwarrior import config
@@ -41,6 +42,7 @@ class GitlabConfig(config.ServiceConfig):
     default_mr_priority: DefaultPriority = 'unassigned'
     use_https: bool = True
     verify_ssl: bool = True
+    body_length: int = sys.maxsize
     project_owner_prefix: bool = False
     issue_query: str = ''
     merge_request_query: str = ''
@@ -432,7 +434,7 @@ class GitlabIssue(Issue):
             if self.extra['type'] == 'todo' else self.record['title'])
         description = (
             self.record['body'] if self.extra['type'] == 'todo'
-            else self.record['description'])
+            else self.extra['description'])
 
         if milestone and (
                 self.extra['type'] == 'issue' or
@@ -597,7 +599,8 @@ class GitlabService(IssueService):
                 'project': repo['path'],
                 'namespace': repo['namespace']['full_path'],
                 'type': issue_type,
-                'annotations': self.annotations(repo, issue_url, type_plural, issue, issue_obj)
+                'annotations': self.annotations(repo, issue_url, type_plural, issue, issue_obj),
+                'description': self.description(issue),
             }
             issue_obj.update_extra(extra)
             yield issue_obj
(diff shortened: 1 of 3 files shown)


ryneeverett:
Looking back at the first time this issue was raised -- #304 -- I think a similar solution would be more appropriate. We should probably add a `gitlab.body_length` option since `description_length` is already taken. ("description" refers to the title. Yes, the terminology overloading between taskwarrior and services is frustrating.)

It would probably also make sense to set both `body_length` options to a reasonable default -- static so that bugwarrior behavior is deterministic across machines.

> I don't really get where the limit is because on my machine getconf ARG_MAX tells 2097152 but the error happened with a line length of ~180000

On my machine I get a somewhat lower value with `xargs --show-limits` than `ARG_MAX` -- could that explain it? I admittedly don't know much about this.

ryneeverett:
For a little more context on why I don't like this solution -- `db.synchronize` is probably the gnarliest part of our code base and I'd like to avoid adding *anything* to it, let alone service-specific code.

Another alternative that comes to mind: Could `taskw` -- the library we use to shell out to taskwarrior -- limit argument length for us?

ryneeverett:
https://www.in-ulm.de/~mascheck/various/argmax/

DyspC:
Ok I tried something to make it work like in the github service wdyt

bugwarrior/services/gitlab.py
```diff
@@ -667,6 +670,16 @@ def get_all_repos(self):
         repos = list(filter(self.filter_repos, all_repos))
         return repos
 
+    def description(self, issue):
+        description = issue['description']
+
+        if description:
+            description = description.replace('\r\n', '\n')
```
ryneeverett: I think this line should be dropped because it is unrelated to this feature. I don't recall the history or purpose behind it, but you can see in the diff of #782 that the github service was already doing this replacement before introducing the body_length option.

bugwarrior/services/gitlab.py
```diff
@@ -432,7 +434,7 @@ def to_taskwarrior(self):
             if self.extra['type'] == 'todo' else self.record['title'])
         description = (
             self.record['body'] if self.extra['type'] == 'todo'
-            else self.record['description'])
+            else self.extra.get('description', ''))
```
ryneeverett: Is there a reason this can't just be `self.extra['description']`? Seems like the `description` key will always be set.

ryneeverett (changes requested):

bugwarrior/services/gitlab.py
DyspC: I was getting keyerrors from pytest when using the simple accessor, maybe because some of the gitlab tests call to_taskwarrior directly and never go through the code that adds the extra (noticed this because those tests were failing on descriptions being None instead of '', hence the default)

ryneeverett: Hope you don't mind, I went ahead and made this change, adjusted the tests to accommodate it, rebased the branch to fix the unrelated CI failures, and pushed the amended commit to your branch.

ryneeverett (approved):

bugwarrior/services/gitlab.py
DyspC: That's alright

PR #1019, opened 2023-11-10
support for python 3.12

NeilHanlon:
python3.12 removed `readfp()`, recommended to use `read_file()` instead

### DIFF
--- a/.github/workflows/bugwarrior.yml
+++ b/.github/workflows/bugwarrior.yml
@@ -10,7 +10,7 @@ jobs:
     timeout-minutes: 5
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10, 3.11, 3.12]
         jira-version: [1.0.10, 2.0.0]
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


NeilHanlon:
Note: the if condition is to support python 2.X... not sure if that is necessary.

ryneeverett:
Thanks. We've already dropped support for python-2. Please remove the if-condition (which should fix CI since you forgot to `import sys`) and, if you wouldn't mind, add the later python versions to `.github/workflows/bugwarrior.yml`. Note that the `python-version`s are listed two different places in the file. (I'm sure there's a smarter way to avoid that duplication but :shrug:.)

NeilHanlon:
Absolutely, I can do that.

Would you like me to add all the versions between 3.9 (latest in the github action config) and 3.12, or just append 3.12 to the list?

ryneeverett:
If you could add all the versions from 3.9 through 3.12 that would be great.

ryneeverett:
Looking at the CI failure, it seems that yaml interprets `3.10` as a float and converts it to `3.1`. I think it would be fixed by just putting `"3.10"` in quotes.

NeilHanlon:
thanks for that - apologies that this fell off my radar

ryneeverett:
No problem, it happens. I let this fall off my radar too since the repository owner opened a rival pull request and I wanted to give him a chance to get involved if he was interested.

ralphbean:
Sorry for competing! Thanks for resolving @ryneeverett.

PR #1020, opened 2023-11-20
read_file replaces readfp

ralphbean:
https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read_file


ryneeverett:
Essentially a duplicate of #1019.

ryneeverett:
Closing in favor of #1019.

Issue #998, opened 2023-07-14
Upgrade to Pydantic V2


ryneeverett:
See #997.

https://docs.pydantic.dev/latest/migration/

ChasNelson1990:
Any news on this? Is it part of the bugwarrior v2.0 release (#1030)?

I ask because the current Arch Linux packages (https://aur.archlinux.org/packages?K=bugwarrior) are both broken. The main package is broken because it's missing #1019 and the `-git` package because Arch's system python pydantic is now version 2.7.

ryneeverett:
At this point I'm not planning to prioritize it for the 2.0 release because it seems likely that pydantic-1.x will continue to be available for quite some time on all other major distributions. It is unfortunate that Arch doesn't have the capacity to support versioned dependencies.

NexAdn:
How about just using the [compatiblilty namespace](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features) for now? This should allow running bugwarrior with pydantic v2 while avoiding the need to upgrade to v2 features immediately.

ryneeverett:
I didn't know about that. Nice!

If it turns out to be as easy as it looks there's no reason this shouldn't happen for the next bugwarrior release. I've added this to the milestone because we should at least try the compatibility namespace approach even though it is unlikely this issue will be fully resolved.

ryneeverett:
Thanks to @NexAdn we no longer have pydantic-1 as a dependency which relieves the distribution concerns. Remaining work is to actually switch over to the pydantic-2 API, but that is not urgent.

ryneeverett:
I just noticed that V2 has [an issue](https://github.com/pydantic/pydantic/issues/7186) (also [referenced in the migration guide](https://docs.pydantic.dev/latest/migration/#url-and-dsn-types-in-pydanticnetworks-no-longer-inherit-from-str)) where a trailing slash is automatically added to url's. This will require a new approach for all fields using `StrippedTrailingSlashUrl` or types inheriting from it.

ryneeverett:
Resolved by #1155.

PR #1028, opened 2024-01-03
Drop python-3.6 support.

ryneeverett:
It EOL'ed a couple years ago. https://devguide.python.org/versions/


ryneeverett:
I originally put this up because I was considering a refactor that involved more extensive use of dataclasses (introduced in 3.7) which didn't end up panning out. I'm merging this now though in service of #998 because pydantic-2 dropped python-3.6 support.

PR #1035, opened 2024-02-12
Adding draft status for GH and Gerrit

thekad:
A draft pull request is identified by a different `draft` boolean flag, not by the `state` field, but it might be an important field to display.

### DIFF
--- a/bugwarrior/services/gerrit.py
+++ b/bugwarrior/services/gerrit.py
@@ -24,6 +24,8 @@ class GerritIssue(Issue):
     FOREIGN_ID = 'gerritid'
     BRANCH = 'gerritbranch'
     TOPIC = 'gerrittopic'
+    STATUS = 'gerritstatus'
+    WIP = 'gerritwip'
 
     UDAS = {
         SUMMARY: {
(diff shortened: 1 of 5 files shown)


ryneeverett:
I'm not sure about this. My initial thought is that it is probably reasonable to add the draft status as an additional UDA rather than combining it into the `state` field.

We generally pass data from external API's into UDA's with minimal transformation with the premise that the taskwarrior database is largely for consumption by further automation. For example, somebody could have their [priority](https://taskwarrior.org/docs/priority/) tuned by `uda.githubstate.open`, which would be somewhat broken by this change. The point is not so much that this is a breaking change as that it breaks the assumption that bugwarrior will pass through API fields with minimal transformation, which allows bugwarrior users to develop scripts based directly on the third party API's rather than based on bugwarrior's intermediary data format.

I don't know if it is relevant to your use case but I would note that the `github.query` option can be used to filter out draft pull requests with `draft:false`, which I'd imagine might be a common thing to want.

thekad:
@ryneeverett I have the status displayed in the subject line of my tasks e.g. `[{{ githubstatus | upper }}]` and if something is shown as `[DRAFT]` I want to know is coming (but don't look at it just yet) or adjust its priority when it changes to `[OPEN]`. I had the same thought about adding a new UDA but wasn't sure which direction was best, I will update the PR to use this approach instead.

ryneeverett:
You can of course use taskwarrior in whatever way is useful to you but it does strike me as odd to want to see unactionable tasks. I would think unactionable tasks should have such a low priority that one would never actually see them. Perhaps you don't have that many actionable tasks though.

ryneeverett (changes requested):
I believe the only example we currently have is `gitlabwip`, but the consensus so far has been that we should handle booleans as `numeric` UDA's with a binary value (0 or 1) since there is no such thing as a boolean UDA.

ryneeverett (approved):

PR #1036, opened 2024-02-13
gmail: add pagination support

mdujava:


bugwarrior/services/gmail.py
```diff
@@ -178,12 +179,28 @@ def get_labels(self):
...
+        threads = []
 
-        result = thread_service.list(
-            userId=self.config.login_name, q=self.config.query).execute()
-        return [
-            thread_service.get(userId='me', id=thread['id']).execute()
-            for thread in result.get('threads', [])]
+        pageToken = None
+
+        while len(threads) <= self.config.email_limit:
```
ryneeverett: Shouldn't this be `<` rather than `<=`? If it's equal we've already hit the limit.

bugwarrior/services/gmail.py
```diff
@@ -178,12 +179,28 @@ def get_labels(self):
...
+            maxResults = min(100, max(self.config.email_limit,
+                                      self.config.email_limit - len(threads)))
```
ryneeverett: Shouldn't this be `min(100, self.config.email_limit - len(threads))`?

The second `max` seems wrong (I think you meant `min`) but also redundant (`email_limit` will always be >= `email_limit - len(threads)`).

bugwarrior/services/gmail.py
```diff
@@ -24,6 +24,7 @@ class GmailConfig(config.ServiceConfig):
         '~/.gmail_client_secret.json')
     query: str = 'label:Starred'
     login_name: str = 'me'
+    email_limit: int = 100
```
ryneeverett: This field should be documented in bugwarrior/docs/services/gmail.rst.

bugwarrior/services/gmail.py
```diff
@@ -178,12 +179,28 @@ def get_labels(self):
...
+                threads.append(thread_service.get(userId='me', id=thread['id']).execute())
+
+            if len(result.get('threads', [])) == 0:
+                break
+
+            pageToken = result.get('nextPageToken', None)
+            if not pageToken:
+                break
+
+        return threads[:self.config.email_limit]
```
ryneeverett: Isn't this slice redundant since the logic above already guarantees that threads will be fetched up to the exact `email_limit`? So couldn't this just be `return threads`? Seems like the purpose of the limit is mostly to avoid excessive http requests anyway.

ryneeverett (commented):
What happens without pagination? Does google send everything with no upper limit by default?

bugwarrior/services/gmail.py
mdujava: You are right, I used the pagination for a long time. Before pushing the update, it passed the quick glance.

bugwarrior/services/gmail.py
mdujava: This might have been needed correction when `<=` in while, and the problem with the `min(100, max( ... )`

mdujava:
Without pagination only first 100 email threads matching the filter are returned. Any threads/tasks not in the results are ~deleted~, regardless of the read/unread status.

Correction: Marked completed.

bugwarrior/docs/services/gmail.rst
```diff
@@ -41,6 +42,9 @@ You do not need to specify the ``login_name``, but it can be useful to avoid
 accidentally fetching data from the wrong account. (This also allows multiple
 targets with the same login to share the same authentication token.)
 
+By default, the last 100 threads matching the query will be returned. You may
+adjust the number of returned threads with ``gmail.thread_limit``.
```
ryneeverett: For consistency with other services, the Example Service section only includes essential fields and optional fields go in their own section with an example of changing from the default behavior.

```suggestion
Thread Limit
-----------------

Gmail search results are capped at 100 max results by default, but may be
adjusted to meet your needs by specifying `gmail.email_limit`:

.. config::
    :fragment: gmail
    
    gmail.thread_limit = 1000
```

bugwarrior/docs/services/gmail.rst
```diff
@@ -33,6 +33,7 @@ Here's an example of a gmail target:
     service = gmail
     gmail.query = label:action OR label:readme
     gmail.login_name = you@example.com
+    gmail.thread_limit = 100
```
ryneeverett:
```suggestion
```

bugwarrior/services/gmail.py
```diff
@@ -178,12 +179,27 @@ def get_labels(self):
...
+            maxResults = min(100, self.config.thread_limit - len(threads))
+
+            result = thread_service.list(userId=self.config.login_name, q=self.config.query,
+                                         maxResults=maxResults, pageToken=pageToken).execute()
+
+            for thread in result.get('threads', []):
+                threads.append(thread_service.get(userId='me', id=thread['id']).execute())
+
+            if len(result.get('threads', [])) == 0:
+                break
```
ryneeverett: Is this `break` condition necessary? Surely in any circumstance in which no threads are returned there would also be no `nextPageToken` and the next `break` condition would apply.

mdujava: checking it even after for loop does not make sense

bugwarrior/docs/services/gmail.rst
```diff
@@ -41,6 +41,17 @@ You do not need to specify the ``login_name``, but it can be useful to avoid
 accidentally fetching data from the wrong account. (This also allows multiple
 targets with the same login to share the same authentication token.)
 
+Thread Limit
+-----------------
+
+Gmail search results are capped at 100 max results by default, but may be
+adjusted to meet your needs by specifying `gmail.email_limit`:
```
ryneeverett: Sorry I missed this before. Need double backticks and we try to keep the docs configuration-language neutral by omitting prefixes.

```suggestion
adjusted to meet your needs by specifying ``email_limit``:
```

ryneeverett (changes requested):

ryneeverett (approved):
Thanks for sticking with it through all the nitpicks.

PR #1039, opened 2024-03-18
Add ability to ignore certain users comments

thekad:
Some build/CI systems can become too noisy, you still want to pull real people's comments and ignore bot accounts when adding annotations.

### DIFF
--- a/bugwarrior/services/gerrit.py
+++ b/bugwarrior/services/gerrit.py
@@ -148,7 +148,6 @@ class GerritService(Service, Client):
                     break
             else:
                 username = item['author']['_account_id']
-            # Gerrit messages are really messy
             if username in self.config.ignore_user_comments:
                 log.debug(" ignoring comment from %s", username)
                 continue
--- a/bugwarrior/services/gerrit.py
+++ b/bugwarrior/services/gerrit.py
@@ -6,6 +6,9 @@ import requests
 from bugwarrior import config
 from bugwarrior.services import Service, Issue, Client
 
+import logging
+log = logging.getLogger(__name__)
+
 
 class GerritConfig(config.ServiceConfig):
     service: typing.Literal['gerrit']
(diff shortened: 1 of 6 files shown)


bugwarrior/services/gerrit.py
```diff
@@ -151,6 +155,9 @@ def annotations(self, change):
             else:
                 username = item['author']['_account_id']
             # Gerrit messages are really messy
```
ryneeverett: This comment was about the cleaning of the message below. You should move it below your addition or, perhaps better, delete it.

ryneeverett (changes requested):
Implementation looks fine but these options need to be documented and tested.

thekad:
> Implementation looks fine but these options need to be documented and tested.

I'll add the test bits for github, there is no annotation testing done at all for gerrit but I can take a shot at it.

Question: Do you think it makes sense to leave these options on a per-service basis?

ryneeverett:
> I'll add the test bits for github, there is no annotation testing done at all for gerrit but I can take a shot at it.

if you can't test gerrit that's ok. Github is the critical one due to it's popularity and number of features.

> Question: Do you think it makes sense to leave these options on a per-service basis?

Yes, I think we should leave them on a per-service basis. Unfortunately, it isn't a safe assumption that usernames are owned by the same entity (or in the case of bots, running the same software) across services and I don't think we should introduce that assumption here. For example, you might want to ignore a spammer who is username-squatting on one platform without ignoring the legitimate user on another platform.

ryneeverett (changes requested):
Just about there, thanks! Still need:

- [x] Gerrit documentation
- [x] Please address or respond to [this review comment](https://github.com/GothenburgBitFactory/bugwarrior/pull/1039/files#r1531088975).

thekad:
I amended the gerrit docs with a similar "sync content" section as github, and removed the comment

ryneeverett (approved):

Issue #573, opened 2018-05-15
gitea: support gitea (and maybe gogs)


genofire:
See swagger comments:
- https://github.com/go-gitea/gitea/blob/master/routers/api/v1/repo/issue.go
- https://github.com/go-gitea/gitea/blob/master/routers/api/v1/repo/pull.go

ralphbean:
Going to close this (cleaning up the queue). However, do see #410 for anyone who wants to contribute this as a patch.

PR #720, opened 2020-04-01
(WIP) Adding service support for Gitea

wamsachel:
Mostly just a copy of github.py

Still need to add tests, auth support for tokens, re-write github api strings into gitea

PR #1048, opened 2024-05-21
Fixes: #573. Add Basic Gitea support

msglm:
This PR lets you use Gitea when using some basic configuration, built off #720.

To test this, you should use this as your working config, it's what's worked for me:
```
[user_gitea]
service = gitea
gitea.login = YOUR-GITEA-USERNAME #no clue if this is a necessary given that username exists
gitea.username = YOUR-GITEA-USERNAME
gitea.host = YOUR-GITEA-INSTANCE
gitea.password = YOUR-GITEA-PASSWORD
gitea.token = A-VALID-GITEA-ACCESS-TOKEN
#Perhaps optional
gitea.description_template = "{{giteanumber}}: {{giteatitle}}"
gitea.only_if_assigned = true
```

Fair warning, this is a hack-job that sucks and needs a lot of clean up, quality of life, documentation, and feature improvements. I'll help whenever convenient, but this patch gets the job done for me as-is.

### DIFF
--- a/bugwarrior/docs/other-services/tutorial.rst
+++ b/bugwarrior/docs/other-services/tutorial.rst
@@ -197,7 +197,14 @@ Now for the main service class which bugwarrior will invoke to fetch issues.
       def __init__(self, *args, **kwargs):
           super().__init__(*args, **kwargs)
 
-          self.client = GitBugClient(path=self.config.path, port=self.config.port)
+          self.client = GitBugClient(
+              path=self.config.path,
+              port=self.config.port,
+              annotation_comments=self.main_config.annotation_comments)
+
+      @staticmethod
+      def get_keyring_service(config):
+          return f'gitbug://{config.path}'
 
       def issues(self):
           for issue in self.client.get_issues():
@@ -214,6 +221,8 @@ Now for the main service class which bugwarrior will invoke to fetch issues.
               yield self.get_issue_for_record(issue)
 
 Here we see two required class attributes (pointing to the classes we previously defined) and two required methods.
+
+The ``get_keyring_service`` method returns a string identifier for secrets in the keyring. Ideally, this string uniquely identifies a given instance of the service when it is possible to have multiple instances of the service configured.
 
 The ``issues`` method is a generator which yields individual issue dictionaries.
 
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -318,9 +318,9 @@ class Service(abc.ABC):
         raise NotImplementedError()
 
     @staticmethod
-    # @abc.abstractmethod
-    def get_keyring_service(service_config):
-        """ Given the keyring service name for this service. """
+    @abc.abstractmethod
+    def get_keyring_service(service_config) -> str:
+        """ Return the keyring name for this service. """
         raise NotImplementedError
 
 
--- a/bugwarrior/services/bts.py
+++ b/bugwarrior/services/bts.py
@@ -145,6 +145,10 @@ class BTSService(Service, Client):
     ISSUE_CLASS = BTSIssue
     CONFIG_SCHEMA = BTSConfig
 
+    @staticmethod
+    def get_keyring_service(config):
+        return 'bts://'
+
     def _record_for_bug(self, bug):
         return {'number': bug.bug_num,
                 'url': 'https://bugs.debian.org/' + str(bug.bug_num),
--- a/bugwarrior/services/deck.py
+++ b/bugwarrior/services/deck.py
@@ -165,6 +165,10 @@ class NextcloudDeckService(Service):
             password=self.config.password
         )
 
+    @staticmethod
+    def get_keyring_service(config):
+        return f'deck://{config.username}@{config.base_uri}'
+
     def get_owner(self, issue):
         return issue[issue.ASSIGNEE]
 
--- a/bugwarrior/services/gitbug.py
+++ b/bugwarrior/services/gitbug.py
@@ -160,6 +160,10 @@ class GitBugService(Service):
             port=self.config.port,
             annotation_comments=self.main_config.annotation_comments)
 
+    @staticmethod
+    def get_keyring_service(config):
+        return f'gitbug://{config.path}'
+
     def issues(self):
         for issue in self.client.get_issues():
             comments = issue.pop('comments')
--- a/bugwarrior/services/gmail.py
+++ b/bugwarrior/services/gmail.py
@@ -135,6 +135,10 @@ class GmailService(Service):
             'gmail_credentials_%s.pickle' % (credentials_name,))
         self.gmail_api = self.build_api()
 
+    @staticmethod
+    def get_keyring_service(config):
+        return f'gmail://{config.login_name}'
+
     def build_api(self):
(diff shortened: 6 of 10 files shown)


bugwarrior/services/gitea.py
```diff
@@ -0,0 +1,564 @@
...
+
+log = logging.getLogger(__name__)  # pylint: disable-msg=C0103
+
+class GiteaConfig(config.ServiceConfig):
+    service: typing_extensions.Literal['gitea']
+
+    host = "gitea.com"
+    login: str
+    token: str
+    login: str
```
ryneeverett:
```suggestion
```
duplicate

bugwarrior/services/gitea.py
```diff
@@ -0,0 +1,564 @@
...
+    username: str
+    password: str
+    exclude_repos = []
+    include_repos = []
```
ryneeverett:
```suggestion
    exclude_repos = config.ConfigList([])
    include_repos = config.ConfigList([])
```

Needed for backwards compatibility with INI format.

bugwarrior/services/gitea.py
```diff
@@ -0,0 +1,564 @@
...
+
+    @classmethod
+    def validate_config(cls, service_config, target):
+        if 'login' not in service_config:
+            log.critical('[%s] has no \'gitea.login\'' % target)
+            sys.exit(1)
+
+        if 'token' not in service_config and 'password' not in service_config:
+            log.critical('[%s] has no \'gitea.token\' or \'gitea.password\'' % target)
+            sys.exit(1)
```
ryneeverett:
```suggestion
```

Validation now happens automagically using `GiteaConfig`.

ryneeverett (commented):
Thanks for updating this to be compatible with the latest development branch!

As you noted, there are a number of things needed yet for us to be able to merge this. Here's what I'm noticing in a quick scan:

- [x] documentation
- [ ] tests
- [x] remove `six`

Looking at some of the remnant TODO's from #720, I would note that additional features are *not* necessary before this is merged.

msglm:
I've never wrote python tests before, so I have not done it yet. 

Documentation is not very good, but I think gets the point across. The gitea feature here is rather indev, so it shouldn't be shocking.

I would like to remove the username-password login combo since making tokens is easy in Gitea, and the username-password login combo tends to break when you use OTP (I enforce it on my instance). Do you think that would be a good idea?

bugwarrior/services/gitea.py
```diff
@@ -370,7 +368,7 @@ def __init__(self, *args, **kw):
             'import_labels_as_tags', default=False, to_type=bool
         )
         self.label_template = self.config.get(
-            'label_template', default='{{label}}', to_type=six.text_type
+            'label_template', default='{{label}}', to_type=bool
```
ryneeverett: `bool` seems wrong. I suspect this whole `self.label_template` definition should probably be eliminated and the whole `get_tags` function could be reduced by using `self.get_tags_from_labels`. See example: https://github.com/GothenburgBitFactory/bugwarrior/blob/71efc97435a442646782dcb0979df43f69ce955d/bugwarrior/services/github.py#L305-L307

ryneeverett (commented):
Another TODO:

- [ ] Add Gitea to the list of services at the top of README.rst

> I would like to remove the username-password login combo since making tokens is easy in Gitea, and the username-password login combo tends to break when you use OTP (I enforce it on my instance). Do you think that would be a good idea?

I have no objection to removing password authentication. Many of the third-party services we integrate with have been deprecating and removing password authentication and I wouldn't be surprised if the rest follow suit.

ryneeverett:
> I've never wrote python tests before, so I have not done it yet.

Not sure if you'd seen this already or if you'll find it helpful but there is a brief section in the guide on adding tests: https://bugwarrior.readthedocs.io/en/latest/contributing/new-service.html#tests

msglm:
Noting this down for myself:
the `involved_issues` feature is broken and will need to be removed.

bugwarrior/services/gitea.py
```diff
@@ -269,6 +273,9 @@ def to_taskwarrior(self):
         if body:
             body = body.replace('\r\n', '\n')
 
+        if len(body) < 1:
+            body = "No annotation was provided."
```
ryneeverett: I'm not sure about this. For consistency with the other services I think an empty body should be allowed.

bugwarrior/services/gitea.py
```diff
@@ -335,44 +336,42 @@ def __init__(self, *args, **kw):
...
+        
+        self.include_assigned_issues = self.config.include_assigned_issues
+        
+        self.include_created_issues = self.config.include_created_issues
+
+        self.include_review_requested_issues = self.config.include_review_requested_issues
+
+        self.import_labels_as_tags = self.config.import_labels_as_tags
+        
+        self.label_template = self.config.label_template
```
ryneeverett: I think all these instance variables should now be removed in favor of accessing `self.config` directly.

bugwarrior/services/gitea.py
```diff
@@ -29,16 +29,28 @@
...
-    login: str
     token: str
     username: str
-    password: str
-    exclude_repos = config.ConfigList([])
-    include_repos = config.ConfigList([])
+    include_assigned_issues: bool = False
+    include_created_issues: bool = False
+    include_mentioned_issues: bool = False
+    include_review_requested_issues: bool = False
```
ryneeverett: Rather than adding a separate configuration option for these specific purposes, I would rather see a more powerful `query` configuration option which can pass arbitrary filters to the gitea api. (This would apply to the `issue_limit` option as well.) See [github.query](https://bugwarrior.readthedocs.io/en/latest/services/github.html#queries) for example.

bugwarrior/services/gitea.py
```diff
@@ -524,10 +522,38 @@ def issues(self):
...
-            )
+
+            '''
+            A variable used to represent the attachable HTTP query that can be attached to the /repos/issues/search API end.
+
+            if httpQuery is set to "review_requested=True?mentioned=True" for example, then the /repos/issues/search API end will be told to search for all issues where a review is requested AND where the user is mentioned.
+            '''
+            httpQuery = "limit=" + str(self.config.issue_limit) + "&"
+
+            if self.config.get('include_assigned_issues', True, bool):
```
ryneeverett: While I questioned whether these options are necessary above, I would note that the way you're accessing the configuration here and below doesn't make a lot of sense to me. You are defining the default values as `False` in the `GiteaConfig` class.

```py
include_assigned_issues: bool = False
```

But here you change the effective default to be `True`. It would make more sense to set the desired default in `GiteaConfig` and access is simply with `self.config.include_assigned_issues`.

Fischmiep:
@msglm Any updates on this? I'd love to see this merged, and I'm willing to contribute in my (albeit very limited) free time if you need someone.

msglm:
I've just been using a patched version of this program with this PR
merged into it on my personal system. If you want to take over polishing
this, be my guest.

msglm:
Accidentally closed this while dealing with git history issues. My personal branch still has the code all intact for this feature.

PR #1052, opened 2024-06-22
New service: Logseq

scross01:
New service to pull tasks from [Logseq](https://logseq.com/)

This integration works with the local HTTP APIs server provided by the Logseq application to query Tasks from the active Logseq graph


bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
+Logseq
+======
+
+You can import `tasks <https://docs.logseq.com/#/page/tasks>`_ from `Logseq <https://logseq.com/>`_ using the ``logseq`` service name.
+
+
+Additional Requirements
+-----------------------
+
+Logseq application must active with the HTTP API server running for bugwarrior to connect.
```
ryneeverett:
```suggestion
Logseq application must interact with the HTTP API server running for bugwarrior to connect.
```

Is this what you meant? Then again, this sentence might be superfluous since the next one says the same thing more clearly.

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def get_issues(self):
+        task_filter = self.filter.replace(" ", "").replace(",", '" "')
+        return self._datascript_query(
+            f"""
+            [:find (pull ?b [*])
+                :where [?b :block/marker ?marker]
+                [(contains? #{{\"{task_filter}\"}} ?marker)]
+            ]
+        """
```
ryneeverett, line 77: I don't know clojure (or is this datascript?) but the examples on [the docs you linked](https://docs.logseq.com/#/page/task) seem to be explicitly filtering for `task` items with, for example, `(task ?b #{"NOW" "LATER" "TODO" "DOING"})`. Could you explain why that isn't desirable here?

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+
+    DOING, TODO, NOW, LATER, IN-PROGRESS, WAIT, WAITING
+
+You can override this filter by setting the ``logseq.task_state`` option to a 
+comma separated list of required task states.
+
+.. config::
+    :fragment: logseq
+
+    logseq.task_state = DOING, NOW, IN-PROGRESS
```
ryneeverett, line 75: We generally prefer adding a single `query` configuration option rather than building the query out of many different configuration options. Is there a reason that approach wouldn't work here?

For example:

```py
class LogseqConfig(config.ServiceConfig):
    ...
    query = ":where [?b :block/marker ?marker] [(contains? #{{\"DOING TODO NOW LATER IN-PROGRESS WAIT WAITING\"}} ?marker)]"
```

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Priority mapping
+++++++++++++++++
+
+Logseq task priorities ``A``, ``B``, and ``C`` mapped to the taskwarrior priorities
```
ryneeverett:
```suggestion
Logseq task priorities ``A``, ``B``, and ``C`` are mapped to the taskwarrior priorities
```

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Task state and data/time mappings
++++++++++++++++++++++++++++++++++
+
+``DOING``, ``TODO``, ``NOW``, ``IN-PROGRESS`` are mapped to the default ``pending`` state.
+The Logseq task ``SCHEDULED:`` and ``DEADLINE:`` fields are mapped to the ``scheduled`` and 
+``due`` date fields.
+
+``LATER``, ``WAITING``, ``WAIT`` are mapped to the ``waiting`` state.
+The ``SHEDULED:`` date or ``DEADLINE`` date is used to set the ``wait`` date on the task.
```
ryneeverett:
```suggestion
The ``SCHEDULED:`` date or ``DEADLINE`` date is used to set the ``wait`` date on the task.
```

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Character replacement
++++++++++++++++++++++
+
+taskwarrior encodes the ``[`` and ``]`` characters commonly used in Logseq as ``&open;`` and ``&close;``. To
+avoid display issues ``[[`` and ``]]`` are replaced by ``【`` and ``】`` for page links, and single
+``[`` and ``]`` are replaced by ``〈`` and ``〉``. 
+
+You can override behaviour and use customer characters by setting the ``logseq.char_*`` options in your
+``bugwarriorrc`` config.
```
ryneeverett:
```suggestion
You can override this behavior and use custom characters by setting the ``logseq.char_*`` options.
```

Since one can also use bugwarrior.toml it's best to avoid naming the file.

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+Waiting tasks can be listed using ``task waiting``
+
+``DONE`` is mapped to the ``completed`` state.
+
+``CANCELED`` and ``CANCELLED`` are mapped to the ``deleted`` state.
+
+Character replacement
++++++++++++++++++++++
+
+taskwarrior encodes the ``[`` and ``]`` characters commonly used in Logseq as ``&open;`` and ``&close;``. To
```
ryneeverett:
```suggestion
Taskwarrior encodes the ``[`` and ``]`` characters commonly used in Logseq as ``&open;`` and ``&close;``. To
```

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+You can override behaviour and use customer characters by setting the ``logseq.char_*`` options in your
+``bugwarriorrc`` config.
+
+.. config::
+    :fragment: logseq
+
+    logseq.char_open_link = 〖
+    logseq.char_close_link = 〗
+    logseq.char_open_bracket = (
+    logseq.char_close_bracket = )
```
ryneeverett: I'm not sure I understand the problem this feature is addressing. Why would regular square brackets cause display issues? Does the Logseq application render them differently?

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Task filters
+++++++++++++
+
+You can configure the service to import tasks in different states.
+By default the service will import all tasks in an active tasks states
+
+    DOING, TODO, NOW, LATER, IN-PROGRESS, WAIT, WAITING
+
+You can override this filter by setting the ``logseq.task_state`` option to a
```
ryneeverett:
```suggestion
You can override this filter by setting the ``task_state`` option to a 
```

Best to leave out the service prefix in prose since those are not used in TOML. (In code blocks, the service prefixes are automatically removed when one clicks on the toml tab.)

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+``CANCELED`` and ``CANCELLED`` are mapped to the ``deleted`` state.
+
+Character replacement
++++++++++++++++++++++
+
+taskwarrior encodes the ``[`` and ``]`` characters commonly used in Logseq as ``&open;`` and ``&close;``. To
+avoid display issues ``[[`` and ``]]`` are replaced by ``【`` and ``】`` for page links, and single
+``[`` and ``]`` are replaced by ``〈`` and ``〉``. 
+
+You can override behaviour and use customer characters by setting the ``logseq.char_*`` options in your
```
ryneeverett:
```suggestion
You can override behaviour and use customer characters by setting the ``char_*`` options in your
```

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Logseq URI links
+++++++++++++++++
+
+A ``logseq://`` URI is generated for each task to enable easy navigation directly to the specific task in
+the Logseq application. 
+
+By default bugwarrior incorporates the links into task description. To disable this behaviour either 
+modify the ``inline_links`` option to affect all services, or to modify for the logseg sevice only you can 
+set the ``logseq.inline_links`` option to False in your ``bugwarriorrc``.
```
ryneeverett:
```suggestion
modify the ``inline_links`` option in the main section to affect all services, or to modify for the logseg sevice only you can 
set it in your Logseq section.
```

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+Logseq API connection issues
+++++++++++++++++++++++++++++
+
+If you get the following error when running bugwarrior:
+
+    CRITICAL:bugwarrior.services.logseq:Unable to connect to Logseq HTTP APIs server. HTTPConnectionPool(host='localhost', port=12315): Max retries exceeded with url: /api (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x105764f20>: Failed to establish a new connection: [Errno 61] Connection refused'))
+
+- Check that the LogSeq application is running
+- Check that the HTTP APIs server is started
+- Check that authorization token is set the APIs server settings and matches the
```
ryneeverett:
```suggestion
- Check that authorization token is set in the API server settings and matches the 
```

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+++++++++++++++++++++++++++++
+
+If you get the following error when running bugwarrior:
+
+    CRITICAL:bugwarrior.services.logseq:Unable to connect to Logseq HTTP APIs server. HTTPConnectionPool(host='localhost', port=12315): Max retries exceeded with url: /api (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x105764f20>: Failed to establish a new connection: [Errno 61] Connection refused'))
+
+- Check that the LogSeq application is running
+- Check that the HTTP APIs server is started
+- Check that authorization token is set the APIs server settings and matches the 
+  ``logseq.token`` in your ``bugwarriorrc``
```
ryneeverett:
```suggestion
  ``token``.
```

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class LogseqConfig(config.ServiceConfig):
+    service: typing_extensions.Literal["logseq"]
+    host: str = "localhost"
+    port: int = 12315
+    token: str = ""
+    task_state: str = "DOING, TODO, NOW, LATER, IN-PROGRESS, WAIT, WAITING"
```
ryneeverett: *If* we're going to keep this setting, `config.ConfigList` would probably make more sense than `str`.

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+    char_open_link: str = "【"
+    char_close_link: str = "】"
+    char_open_bracket: str = "〈"
+    char_close_bracket: str = "〉"
+    inline_links = True  # set to False to override bugwarrior config
```
ryneeverett:
```suggestion
    inline_links: bool = True
```

May as well annotate the type for consistency. The comment seems superfluous.

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+from bugwarrior.services import IssueService, Issue, ServiceClient
+
+log = logging.getLogger(__name__)
+
+
+class LogseqConfig(config.ServiceConfig):
+    service: typing_extensions.Literal["logseq"]
+    host: str = "localhost"
+    port: int = 12315
+    token: str = ""
```
ryneeverett:
```suggestion
    token: str
```

Leave out the default value in order to make this required. I'd do the same with the `port` too unless 12315 is Logseq's default value.

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,150 @@
...
+Example Service
+---------------
+
+Here's an example of a Logseq target:
+
+.. config::
+
+    [my_logseq_graph]
+    service = logseq
+    logseq.host = localhost
```
ryneeverett:
```suggestion
```

We only put required configuration fields here and this seems like a default most people would be using. Please add a section below to mention how one can change the `host` value.

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+        self.port = port
+        self.token = token
+        self.filter = filter
+
+        self.headers = {
+            "Authorization": "Bearer " + self.token,
+            "content-type": "application/json; charset=utf-8",
+        }
+
+        self.project = None
```
ryneeverett:
```suggestion
```

Unless I'm missing something this attribute is unused.

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+        try:
+            response = requests.post(
+                f"http://{self.host}:{self.port}/api",
+                headers=self.headers,
+                json={"method": "logseq.DB.datascriptQuery", "args": [query]},
+            )
+            return self.json_response(response)
+        except requests.exceptions.ConnectionError as ce:
+            log.fatal("Unable to connect to Logseq HTTP APIs server. %s", ce)
+            exit()
```
ryneeverett: Today I learned there is [a builtin exit function](https://docs.python.org/3.8/library/constants.html#exit). While I'm more used to seeing `sys.exit` or simply raising an exception, I guess this is fine. However, I do think you should specify an exit code because otherwise when it propagates to [SystemExit](https://docs.python.org/3.8/library/exceptions.html#SystemExit) " if it is None, the exit status is zero".
```suggestion
            exit(1)
```

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+        try:
+            response = requests.post(
+                f"http://{self.host}:{self.port}/api",
+                headers=self.headers,
+                json={"method": "logseq.getCurrentGraph", "args": []},
+            )
+            return self.json_response(response)
+        except requests.exceptions.ConnectionError as ce:
+            log.fatal("Unable to connect to Logseq HTTP APIs server. %s", ce)
+            exit()
```
ryneeverett:
```suggestion
            exit(1)
```

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+    def _unescape_content(self, content):
+        return (
+            content.replace('"', "'")  # prevent &dquote; in task details
+            .replace("[[", self.config.char_open_link)  # alternate brackets for linked items
+            .replace("]]", self.config.char_close_link)
+            .replace("[", self.config.char_open_bracket)  # prevent &open; and &close;
+            .replace("]", self.config.char_close_bracket)
+        )
+
+    # get a optimized and
```
ryneeverett: ?

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+
+    # get a list of tags from the task content
+    def get_tags_from_content(self):
+        # this includes #tagname, but ignores tags that are in the #[[tag name]] format
+        tags = re.findall(
+            r"(#[^" + self.config.char_open_link + r"^\s]+)", self.get_formated_title()
+        )
+        return tags
+
+    # get a list of annotations form the content
```
ryneeverett:
```suggestion
    # get a list of annotations from the content
```

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+        # format is <YYYY-MO-DD DAY HH:MM .+1d>
+        # e.g. <2024-06-20 Thu 10:55 .+1d>
+        date_split = (
+            scheduled.replace("DEADLINE: <", "")
+            .replace("SCHEDULED: <", "")
+            .replace(">", "")
+            .split(" ")
+        )
+        date = ""
+        date_format = ""
```
ryneeverett:
```suggestion
```

These variables don't need to be initialized.

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+            "due": deadline_date,
+            "scheduled": scheduled_date,
+            "wait": wait_date if self._is_waiting() else None,
+            "status": self.STATE_MAP[self.get_logseq_state()],
+            self.ID: self.record["id"],
+            self.UUID: self.record["uuid"],
+            self.STATE: self.record["marker"],
+            self.TITLE: self.get_formated_title(),
+            self.SCHEDULED: scheduled_date,
+            self.DEADLINE: deadline_date,
```
ryneeverett: Why do we need these UDA's when the same data is already in `scheduled` and `due`?

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+        # Issue assignment hasn't been implemented yet.
+        raise NotImplementedError(
+            "This service has not implemented support for 'only_if_assigned'."
+        )
+
+    def issues(self):
+        graph_name = self.client.get_graph_name()
+        for issue in self.client.get_issues():
+            extra = {}
+            extra["graph"] = graph_name
```
ryneeverett:
```suggestion
            extra = {"graph": graph_name}
```

ryneeverett (commented):
Thanks for submitting this! I've played around with Logseq from time to time but hadn't even considered integrating it with my taskwarrior workflow.

bugwarrior/services/logseq.py
scross01: I had some trouble with this, using the `task` query style doesn't work with the `logseq.DB.datascriptQuery` API.  The contains/marker approach is a common alternative used is lots of Logseq query examples.

I'm no expert on clojure and datascript, but I suspect this is because `task` is a clojure function where as the API requires just the expanded datascript query.

bugwarrior/docs/services/logseq.rst
scross01: Simplicity, and to avoid query issues. Arbitrary datascript queries can return result sets that do not match the expected list of tasks, plus issues like above where the `logseq.DB.datascriptQuery` function doesn't match to queries that work in the Logseq app itself.

bugwarrior/services/logseq.py
scross01: This is a work around for what is possibly an underlying bug in taskwarrior. The `[` and `]` characters in the task description, and other fields, are replaced with `&open;` and `&close;`.

bugwarrior/services/logseq.py
scross01: good point, they can be dropped.

bugwarrior/docs/services/logseq.rst
scross01: What I meant was "Logseq application must **be** active". Your right, next line says much the same thing so this can be dropped.

bugwarrior/docs/services/logseq.rst
ryneeverett: > Simplicity...

That is arguable. If `task_state` were the only thing one might want to configure it is certainly simpler than a `query` option, but our experience with other services is that folks are going to want to configure every aspect of a query which leads to a much more complicated outcome than a single `query` parameter.

> Arbitrary datascript queries can return result sets that do not match the expected list of tasks,

Not all of our services map to things that are strictly "tasks" or "issues". For example, the Gmail service fetches emails. It seems to me that folks are likely to want to track artifacts in Logseq that are not "Logseq tasks".

> plus issues like above where the `logseq.DB.datascriptQuery` function doesn't match to queries that work in the Logseq app itself.

Is the `logseq.DB.datascriptQuery` API documented? Seems like it would be fine to say that the `query` option takes a `datascriptQuery` and link to the docs.

*BTW, I'm not hard set against your current approach, I'm just trying to draw out your rationale.*

bugwarrior/services/logseq.py
ryneeverett: I'm not sure this is the review comment you meant to respond to. This comment was about the python comment above which seems like an incomplete thought.

bugwarrior/docs/services/logseq.rst
scross01: The API is *not* very well documented https://plugins-doc.logseq.com/logseq/DB/datascriptQuery
I got it working through trial and error.

I'd be happy to add a full query override as an advanced option, but maybe see if anyone actually asks for it first. I would suspect the simpler `task_state` selection will address 99%+ of users needs - if anyone other than me actually uses this integration :-)

bugwarrior/services/logseq.py
scross01: LOL. Yes, an incomplete thought. I think I meant something like "get an optimized title"

bugwarrior/docs/services/logseq.rst
ryneeverett: > the simpler `task_state` selection will address 99%+ of users needs

Ok, that's basically what I was hoping to hear which leads into the next question i had: Is there anything special in Logseq about the default `task_state` values you've set or are they arbitrary strings used by convention? That is, why do you suspect users will want to change this value?

scross01: The complete set of available task states in LogSeq is 
```
DOING, TODO, NOW, LATER, IN-PROGRESS, WAIT, WAITING, CANCELLED, CANCELED, DONE
```
I picked the set of what would be considered "open" tasks as the default.  My for my personal use I filter to just getting DOING, TODO, and WAITING

ryneeverett: Can you explain your use case for wanting to exclude certain "open" states?

scross01: LogSeq has two task workflow options TODO -> DOING -> DONE and LATER -> NOW -> DONE. I actually use a mix of both approaches where I tag things "I" need to do with TODO and DOING, and things I've delegated to others with NOW/LATER.  In taskwarrior I just want to pull the things I need to work on.

ryneeverett: Thanks, that's the last piece of the puzzle. To summarize my understanding of this thread:

- The meaning of Logseq's states is ill-defined, so we need to provide a configuration mechanism for filtering out states that users may use for unactionable tasks.
- Mastery of the poorly documented datascriptQuery language is not necessary for all Logseq users, so merely exposing the entire query as a configuration option is not very user friendly.

Time will tell if you made the right judgment but I'm satisfied that you thought through this decision, so I'm going to mark this conversation as resolved.

bugwarrior/docs/services/logseq.rst
ryneeverett: I think [this comment](https://github.com/GothenburgBitFactory/bugwarrior/pull/1052#discussion_r1650043792) belongs here:

> This is a work around for what is possibly an underlying bug in taskwarrior. The `[` and `]` characters in the task description, and other fields, are replaced with `&open;` and `&close;`.

Hmm, at a cursory glance it seems like Taskwarrior is correctly encoding them but failing to decode in `src/Task.cpp`:

```cpp
////////////////////////////////////////////////////////////////////////////////
// Encode values prior to serialization.
//   [  -> &open;
//   ]  -> &close;
const std::string Task::encode (const std::string& value) const
{
  auto modified = str_replace (value,    "[", "&open;");
  return          str_replace (modified, "]", "&close;");
}

////////////////////////////////////////////////////////////////////////////////
// Decode values after parse.
//   [  <- &open;
//   ]  <- &close;
const std::string Task::decode (const std::string& value) const
{
  if (value.find ('&') == std::string::npos)
    return value;

  auto modified = str_replace (value,    "&open;",  "[");
  return          str_replace (modified, "&close;", "]");
}
```

@scross01 What version of taskwarrior are you on?

@djmitche Any thoughts on this potential Taskwarrior bug?

scross01: I'm using taskwarrior 3.0.2 on MacOS, installed using Homebrew

ryneeverett: I cannot reproduce the decoding bug on the command line by adding a task with brackets in the description on either tw-2 or tw-3. I'm not sure why bugwarrior would be any different since it is using the command line interface under the hood. @scross01 Can you verify manually adding and reading a task on the command line with brackets is or is not buggy on your system? Also, are you seeing the bug with the default report (e.g., running `task 42`) or are you using something fancier?

djmitche: This has been a long-standing bug, which I thought I reported once upon a time but can't find in https://github.com/GothenburgBitFactory/taskwarrior/issues right now. I filed https://github.com/GothenburgBitFactory/taskwarrior/issues/3524 just now.

scross01: I ran
```shell
task add "test [[Task]]"
```
and running `task list` returns the result with
```
22 1min                                         test [[Task]]                                           0
```

I removed the escape code from the integration, with a task in Logseq called 
```
TODO Test Logseq [[Task]]
```
then ran `bugwarrior pull` (note was previously synced with the escaping in place)
```
INFO:bugwarrior.db:Updating task a633286d-c62c-4ec5-88c0-669ff29a2e3a, (bw)Is#126 - TODO Test Logseq [[Task]]; logseqid: '126' -> 126; logseqtitle: 'TODO Test Logseq 〖Task〗' -> 'TODO Test Logseq [[Task]]'; description: '(bw)Is#126 - TODO Test Logseq 〖Task〗' -> '(bw)Is#126 - TODO Test Logseq [[Task]]'
``` 
and then `task list` includes
```
23 1min   MyLogseqGraph                                    (bw)Is#126 - TODO Test Logseq &open;&open;Task&close;&close;        1
```

djmitche: Can you see what `task` invocation bugwarrior-pull is running in that case?

scross01: Looks like the problem could be within the `taskw` library. I ran a simple test
```python
from taskw import TaskWarrior
w = TaskWarrior()
w.task_add("Task added using [[taskw]]")
```
```python
{'id': 24,
 'description': 'Task added using &open;&open;taskw&close;&close;',
 'entry': '20240628T102423Z',
 'modified': '20240628T102423Z',
 'status': 'pending',
 'uuid': '9f1544f9-e66e-41ca-b65e-79c328bac19c',
 'urgency': 0}
```

scross01: specifically this appears to be the cause https://github.com/ralphbean/taskw/blob/3256c92bed4bb3d53269c843e8201e9ac65780cc/taskw/utils.py#L27

`taskw` is escaping the `[`, `]` and `"` characters before invoking `task` command line.
```python
[b'task', b'rc.verbose=new-uuid', b'rc.json.array=TRUE', b'rc.confirmation=no', b'rc.dependency.confirmation=no', b'rc.recurrence.confirmation=no', b'9f1544f9-e66e-41ca-b65e-79c328bac19c', b'modify', b'description:"Task added using &open;&open;taskw&close;&close; + (does) {this} &dquot;quote&dquot; work?"', b'entry:"20240628T102423Z"', b'modified:"20240628T110925Z"', b'status:"pending"']
```

djmitche: Thanks! I've updated the taskwarrior bug.

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    STATE_MAP = {
+        "IN-PROGRESS": "pending",
+        "DOING": "pending",
+        "TODO": "pending",
+        "NOW": "pending",
+        "LATER": "waiting",
+        "WAIT": "waiting",
+        "WAITING": "waiting",
```
ryneeverett: Per https://github.com/GothenburgBitFactory/taskwarrior/pull/2563, I'm now under the impression that the status should never be manually set to `waiting`. Given that, I'm surprised it worked. @djmitche Can you confirm that I'm understanding this correctly?

scross01: It does look like the "waiting" status is set based on the `wait` date field being set, not the actual `status`.  I'm also reconsidering whether LATER should be considered a waiting state. I think the logic should be: if the task type is WAIT or WAITING then set the `wait` date to the earliest of the SCHEDULED date,  DEADLINE date, or the SOMEDAY date

scross01: Updated the logic to use wait date rather that waiting status.

bugwarrior/services/logseq.py
scross01: Changed to use `config.ConfigList`

bugwarrior/docs/services/logseq.rst
scross01: Moved to new section as suggested

bugwarrior/services/logseq.py
scross01: Not used.  Removed

bugwarrior/docs/services/logseq.rst
ryneeverett: Created https://github.com/ralphbean/taskw/issues/172 to track the actual underlying bug in taskw. I'm thinking if we documented at the beginning of this section that this is a workaround for https://github.com/ralphbean/taskw/issues/172 to give a little more context to these options that would resolve this thread.

scross01: Updated with reference to the `taskw` issue in the code and the service doc.

ryneeverett: Thanks. Maybe I'm reading too much into it but I'm curious if there was intent behind the phrase "in part" where you say "This capability is *in part* to workaround...". I presume you'd be fine with deprecating and eventually removing this feature once the taskw bug is fixed?

scross01: Specifically for the double braces `[[` and `]]` I would consider leaving the capability in place, to allow alternate styling of the page references.

djmitche: It would be great to update the comment to say that `taskw` does this encoding, not Taskwarrior.

scross01: Already updated in commit https://github.com/GothenburgBitFactory/bugwarrior/pull/1052/commits/7a94d57f663b89b8445360e234772c0c1f75deb3

djmitche: Oh, sorry, the GitHub UI confused me

ryneeverett: > Specifically for the double braces `[[` and `]]` I would consider leaving the capability in place, to allow alternate styling of the page references.

Does Logseq itself have a feature to change this styling? To me this would seem like a vanity feature which doesn't provide enough value to justify its maintenance and documentation if it were not for the taskw bug it serves to work around.

*Note: Resolving this difference of opinion is not blocking this PR at this point and can be postponed until the upstream bug is fixed.*

bugwarrior/docs/services/logseq.rst
```diff
@@ -0,0 +1,183 @@
...
+Task state and data/time mappings
++++++++++++++++++++++++++++++++++
+
+``DOING``, ``TODO``, ``NOW``, ``LATER``, and ``IN-PROGRESS`` are mapped to the default ``pending`` state.
+The Logseq task ``SCHEDULED:`` and ``DEADLINE:`` fields are mapped to the ``scheduled`` and 
+``due`` date fields.
+
+``WAITING`` and ``WAIT`` are dynamically mapped to either ``pending`` or ``waiting`` states based on 
+the ``wait`` date. The ``SCHEDULED:`` date or ``DEADLINE`` date is used to set the ``wait`` date on the
+task. If no scheduled or deadlines date is available then the wait date is set to ``someday``
```
ryneeverett:
```suggestion
task. If no scheduled or deadline date is available then the wait date is set to ``someday`` 
```

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,309 @@
...
+        return self._unescape_content(first_line)
+
+    # get a list of tags from the task content
+    def get_tags_from_content(self):
+        # this includes #tagname, but ignores tags that are in the #[[tag name]] format
+        tags = re.findall(
+            r"(#[^" + self.config.char_open_link + r"^\s]+)",
+            self.get_formated_title()
+        )
+        # and this add teh #[[multi word]] formated tags
```
ryneeverett:
```suggestion
        # and this add the #[[multi word]] formatted tags
```

bugwarrior/services/logseq.py
```diff
@@ -0,0 +1,309 @@
...
+    # e.g from #[[Multi Word]] to #MultiWord
+    def _compress_tag_format(self, tag):
+        return (
+            tag.replace(self.config.char_open_link, "")
+            .replace(" ", "")
+            .replace(self.config.char_close_link, "")
+        )
+
+    # get an optimized and formatted title
+    def get_formated_title(self):
```
ryneeverett:
```suggestion
    def get_formatted_title(self):
```

ryneeverett (approved):

bugwarrior/docs/services/logseq.rst
scross01: fixed

bugwarrior/services/logseq.py
scross01: fixed

bugwarrior/services/logseq.py
scross01: fixed

ryneeverett (approved):

scross01:
Anything else needed to merge this PR?

ryneeverett:
Just for somebody to hit the button. :)

PR #1053, opened 2024-06-22
exclude "tests.*" from find_packages

Jimmy2027:
Otherwise tests.config is taken as a package source.


ryneeverett (approved):
I'm surprised that subdirectories need to be explicitly excluded but it does seem consistent with [the docs](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#packages). Thanks!

Jimmy2027:
yeah I had to look it up as well... thanks for accepting the fix!

PR #1068, opened 2024-09-17
use pydantic v2 backwards compat mode

NexAdn:
This should allow using the existing pydantic v1 features while migration to v2 is not done.

Issue: https://github.com/GothenburgBitFactory/bugwarrior/issues/998


setup.py
```diff
@@ -60,7 +60,7 @@
           "dogpile.cache>=0.5.3",
           "jinja2>=2.7.2",
           "lockfile>=0.9.1",
-          "pydantic[email]<2",
+          "pydantic[email]",
```
ryneeverett: With this change, pydantic-1 will no longer work because it doesn't have the `v1` module. That's the cause of the CI failure -- pydantic-1 is still the default on some distributions.

```suggestion
          "pydantic[email]>=2",
```

NexAdn: Yes. Although, before forcing pydantic >= 2, maybe it might make sense to do some fallback magic like catching the ImportError and using the old import again, if `pydantic.v1` cannot be found? This would probably imply having to use `from pydantic.v1 import ...` style imports as we can no longer use `pydantic.v1.[ClassName]` style names.

What's the maintainers' opinion on this?

ryneeverett: If you want to support fallback to pydantic-1 I would accept it but personally I don't see a lot of value in continuing pydantic-1 support because I don't think there are going to be any platforms which don't have pydantic-2 available. I could be wrong though.

NexAdn: Okay then, I've applied the v2 requirement now, since making the code version agnostic would be way more work. If pydantic v2 is available on most systems anyway, backwards compatibility is IMO not worth the extra work.

ryneeverett:
I believe CI is failing because pydantic-2 dropped support for python-3.6. I just removed 3.6 from the test matrix -- we don't need to support it anyways -- so if you rebase on `develop` we should be good to go.

NexAdn:
> I believe CI is failing because pydantic-2 dropped support for python-3.6. I just removed 3.6 from the test matrix -- we don't need to support it anyways -- so if you rebase on `develop` we should be good to go.

Rebase done, CI seems to be happy now

ryneeverett (approved):

PR #1082, opened 2025-01-07
Remove redundant state marker from logseq task description

ctem:
`logseqstate` is already a provided UDA field and thus can be inserted into the description via templating at the user's discretion.


ryneeverett (approved):

ryneeverett:

scross01:
I totally missed the original ping on this.  +1 for the change. Thanks @ctem for this and the other fixes.

ctem:
Thanks to both of you!

PR #1083, opened 2025-01-07
Improve logseq tag handling

ctem:
This PR includes two fixes:

- Remove the `#` tag prefix to avoid "Malformed entry" errors in Taskwarrior when applying coefficients to tags.
- Use non-greedy regex to ensure page link-style tags are extracted separately.

The relevant documentation is also updated.

### DIFF
--- a/bugwarrior/docs/services/logseq.rst
+++ b/bugwarrior/docs/services/logseq.rst
@@ -149,10 +149,10 @@ From the command line you can open a specific task using taskwarior task id, e.g
 Tags
 ++++
 
-LogSeq tasks with ``#tag`` style tag entries in the description are added to the Taskwarrior tags.
-Multi and single word tags using the Logseq ``#[[Tag]]`` or ``#[[Multi Word]]`` format are
-condenced to a ``#Tag`` and ``#MultiWord`` style before adding the Taskwarrior tags. The format of 
-the tag content in task desciption is unchanged.
+Logseq tasks with ``#tag`` style tag entries in the description are added to the Taskwarrior tags.
+Single- and multi-word tags using the Logseq ``#[[Tag]]`` or ``#[[Multi Word]]`` format are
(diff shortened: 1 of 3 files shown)


ryneeverett (approved):
I don't entirely follow the non-greedy regex change because I'm not familiar enough with logseq to know what "page link-style tags" are. If you added such a tag to the tests somewhere that would be a nice touch.

ryneeverett:

ctem:
> "page link-style tags"

Logseq supports several syntaxes, including:

- `#PageTitle`: conventional, no-whitespace tag (called "single-word tag" in bugwarrior docs)
- `[[Page Title]]`: page link—_not treated as a tag in bugwarrior_
- `#[[Page Title]]`: "page link-style" tag for whitespace support (called "multi-word tag" in bugwarrior docs)

> the non-greedy regex change

I'll try to illustrate the issue with an example procedure:

1. Consume the logseq issue title:
   `TODO Do something #[[Test tag]] #[[TestTag]] #TestTag`
1. Format the title, replacing brackets with respective special characters:
   `TODO Do something #【Test tag】 #【TestTag】 #TestTag`
1. Extract tags from the formatted title, starting with conventional tags, then greedily matching page link-style tags (which will match from the first `#` to the last `】`, resulting in a single match that includes both tags):
   `['#TestTag', '#【Test tag】 #【TestTag】']`
1. Remove whitespace and the special characters from all tags:
   `['#TestTag', '#Testtag#TestTag']`
1. Strip the leading `#` from all tags for the final result:
   `['TestTag', 'Testtag#TestTag']`

By switching to non-greedy matching, the two page link-style tags are matched separately, resulting in the correct final list, `['TestTag', 'Testtag', 'TestTag']`, which taskwarrior will ultimately deduplicate to `['TestTag', 'Testtag']`.

Hope that makes sense!

ryneeverett (approved):
Thanks!

scross01:
Nice fix. Thanks

PR #1098, opened 2025-06-03
Don't add annotations for Logseq-specific lines

twaugh:
The `:LOGBOOK:` lines, as well as `id::` strings, are only useful within Logseq. Importing them into Taskwarrior just adds clutter.


ryneeverett (approved):
It would be a nice touch if we could test this by adding these lines to `test_record['content']` in test_logseq.py.

ryneeverett:

scross01:
Looks good to me.

twaugh:
Updated test case

ryneeverett:
Thanks!

Issue #1078, opened 2024-11-13
taskw


ryneeverett:
Bugwarrior currently depends on [taskw](https://github.com/ralphbean/taskw/) to interface with the taskwarrior database. While there is an implementation which interacts with the taskwarrior-2 database files directly, bugwarrior has for some time used the interface which interacts through the `task` shell command. This turned out to be a fortunate choice, because bugwarrior continues to work with taskwarrior-3 even though it has a totally different database.

However, taskw currently lacks much in the way of maintenance. Ralph offered to transfer taskw to GothenburgBitFactory earlier this year, but there were some communication issues which prevented that from happening. I don't think there's any question that this is an option and that all parties are open to it, but I am unsure who has the interest and skill to maintain taskw if that transfer were made. I would be willing to try to help if that were the path forward, but my recollection is that taskw maintenance was particularly difficult because it was hard to fix something for one dependent application without introducing bugs in another dependent application.

I'd be remiss if I didn't mention that I've become aware of [taskw-ng](https://github.com/bergercookie/taskw-ng), a fork (or rewrite?) which appears to be maintained. It also might be worth considering asking bergercookie to be a taskw maintainer if it were to be transferred to GothenbergBitFactory. I haven't looked at it closely at all.

In addition to the current lack of a maintainer, we have always had a few buggy edge cases in which certain data can't be serialized such that it is interpreted correctly by the shell. I suspect we're dealing with another one in #1075, which got me thinking about this this week.

Another option at this particular moment would be to create a taskw shim package in bugwarrior which mimics the small portion of the taskw API we currently use and integrates with taskchampion (the taskwarrior-3 database). We could then use our shim package and fall back to the real taskw for taskwarrior-2 users. (The motivation to continue supporting taskwarrior-2 may expire before this issue is resolved, in which case the shim API is irrelevant...) I created a separate issue about integrating with taskchampion: #1077

KevinMFong:
Has switching to [tasklib](https://github.com/GothenburgBitFactory/tasklib) been considered? I recognize it's a big project, but `tasklib` is already in GothenburgBitFactory, and this could be a good opportunity to consolidate the community on a single library for interfacing with the taskwarrior database.

ryneeverett:
[...]

PR #1101, opened 2025-07-15
New service: Todoist

scross01:
New service to pull tasks from [Todoist](https://www.todoist.com)

This integration uses the [Todoist API Python Client](https://pypi.org/project/todoist_api_python/) which requires Python >=3.9. 

Tasks
- [x] Remove Python 3.8 (EOL) from bugwarrior test harness
- [x] add option to map due→scheduled vs due→due
- [x] use `import_labels_as_tags` option when setting tags from labels
- [x] add support to get comments for each issue and add as annotations
- [x] add support for `only_if_assigned` and `also_unassigned` options


ryneeverett:
Please go ahead and drop python-3.8 since it is EOL.

bugwarrior/docs/services/todoist.rst
```diff
@@ -0,0 +1,98 @@
...
+Authorization Token
++++++++++++++++++++
+
+The authorization ``token`` is used to authenticate with Todoist. See the Todoist
+documentition on how to `find your API token <https://www.todoist.com/help/articles/find-your-api-token-Jpzx9IIlB>`
+
+.. config::
+    :fragment: todoist
+
+    todoist.token = <API_TOKEN>
```
ryneeverett: I don't think we need to cover the token a second time. Required options do not typically have a subsection in Service Features so maybe just add the additional find-your-api-token link to the above paragraph if you think it's helpful.

bugwarrior/docs/services/todoist.rst
```diff
@@ -0,0 +1,98 @@
...
+
+.. config::
+    :fragment: todoist
+
+    todoist.filter = (today | tomorrow | overdue | next 5 days)
+
+Priority mapping
+++++++++++++++++
+
+Todosit task priorities ``p1``, ``p2``, and ``p3`` are mapped to the taskwarrior priorities
```
ryneeverett:
```suggestion
Todoist task priorities ``p1``, ``p2``, and ``p3`` are mapped to the taskwarrior priorities
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+        },
+        URL: {
+            "type": "string",
+            "label": "Todoist URL",
+        },
+    }
+
+    UNIQUE_KEY = (ID, ID)
+
+    # replace characters that cause escaping issues in teh description like [] and "
```
ryneeverett: This eliminates a typo but it also eliminates the implication that this only impacts the description when it is also applied to the content.

```suggestion
    # replace characters that cause escaping issues like [] and "
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+    # this is a workaround for https://github.com/ralphbean/taskw/issues/172
+    def _unescape_content(self, content):
+        return (
+            content.replace('"', "'")  # prevent &dquote; in task details
+            .replace("[", self.config.char_open_bracket)  # prevent &open; and &close;
+            .replace("]", self.config.char_close_bracket)
+        )
```
ryneeverett, line 156: Do you think we should consider moving this method to the base `Issue` class and applying it to the relevant fields in every service? Alternately, I wonder if it would break anything if we just had the base class apply this to every field on every service.

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+        all_users = []
+        users_iter = self._api.get_collaborators(project_id)
+        for users in users_iter:
+            for user in users:
+                all_users.append(user)
+        return all_users
+
+    def get_issues(self):
+        tasks_iter = self._api.filter_tasks(query=self.filter)
+        return tasks_iter
```
ryneeverett: It would be nice to abstract away the API a bit more here.
```suggestion
        for tasks in tasks_iter:
            for task in tasks:
                yield task
```
Then you could eliminate the nested loops in the `issues` method below and just do `for issue in self.client.get_issues()`

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+            "type": "string",
+            "label": "Todoist Assigner",
+        },
+        URL: {
+            "type": "string",
+            "label": "Todoist URL",
+        },
+    }
+
+    UNIQUE_KEY = (ID, ID)
```
ryneeverett:
```suggestion
    UNIQUE_KEY = (ID,)
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def to_taskwarrior(self):
+        default_time = time(0, 0, 0)
+        # use due date "scheduled".
+        # adjust timezone to use local time for "floating" dates
+        if self.record.due and type(self.record.due.date) is datetime:
+            if self.record.due.timezone:
+                scheduled = self.record.due.date
+            else:
+                scheduled = self.record.due.date.replace(tzinfo=None)
```
ryneeverett: So if `due.timezone` is not set, the `tzinfo` has the wrong timezone? Am I understanding that correctly? Seems odd.

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+        else:
+            scheduled = (
+                datetime.combine(self.record.due.date, default_time, tzinfo=None)
+                if self.record.due
+                else None
+            )
```
ryneeverett: I believe this is logically identical but easier to follow.
```suggestion
        if self.record.due:
            if type(self.record.due.date) is datetime:
                if self.record.due.timezone:
                    scheduled = self.record.due.date
                else:
                    scheduled = self.record.due.date.replace(tzinfo=None)
            else:
                scheduled = datetime.combine(self.record.due.date, default_time, tzinfo=None)
        else:
            scheduled = None
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+            content.replace('"', "'")  # prevent &dquote; in task details
+            .replace("[", self.config.char_open_bracket)  # prevent &open; and &close;
+            .replace("]", self.config.char_close_bracket)
+        )
+
+    def to_taskwarrior(self):
+        default_time = time(0, 0, 0)
+        # use due date "scheduled".
+        # adjust timezone to use local time for "floating" dates
+        if self.record.due and type(self.record.due.date) is datetime:
```
ryneeverett: This is what I'd typically expect to see here. The functional difference is that it will be True if the date is a subclass of datetime but I'd think that would be ok here.
```suggestion
        if self.record.due and isinstance(self.record.due.date, datetime):
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+    def _unescape_content(self, content):
+        return (
+            content.replace('"', "'")  # prevent &dquote; in task details
+            .replace("[", self.config.char_open_bracket)  # prevent &open; and &close;
+            .replace("]", self.config.char_close_bracket)
+        )
+
+    def to_taskwarrior(self):
+        default_time = time(0, 0, 0)
+        # use due date "scheduled".
```
ryneeverett:
```suggestion
        # use due date as "scheduled".
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+                scheduled = self.record.due.date.replace(tzinfo=None)
+        else:
+            scheduled = (
+                datetime.combine(self.record.due.date, default_time, tzinfo=None)
+                if self.record.due
+                else None
+            )
+
+        # use deadline as "due".
+        # deadline if set is only a date with no time or timezone. adjust to locla time
```
ryneeverett: This eliminates a typo but also it seems to be an error -- it isn't adjusted to local time it's just set to midnight with no timezone, right?
```suggestion
        # deadline if set is only a date with no time or timezone.
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+        due = (
+            datetime.combine(self.record.deadline.date, default_time, tzinfo=None)
+            if self.record.deadline
+            else None
+        )
+
+        task = {
+            "project": self.extra["project"],
+            "priority": self.PRIORITY_MAP[self.record.priority],
```
ryneeverett: I'm realizing we ought to make this change in the logseq service too because this will respect the `default_priority` option.
```suggestion
            "priority": self.get_priority(),
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+            # "annotations": None,
```
ryneeverett: If there's nothing comparable to comments for this service then we don't need to worry about it.
```suggestion
```

scross01:
Removed 3.8 support from the test harness. Not sure why the 3.9, 2.0.0 test result above shows as Cancelled.

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+            "tags": self.record.labels if self.record.labels else [],
```
ryneeverett:
```suggestion
            "tags": self.get_tags_from_labels(self.record.labels) if self.record.labels else [],
```

bugwarrior/docs/services/todoist.rst
```diff
@@ -0,0 +1,98 @@
...
+
+.. config::
+    :fragment: todoist
+    
+    todoist.inline_links = False
+
+Lables
+++++++
+
+Todoist `Labels <https://www.todoist.com/help/articles/introduction-to-labels-dSo2eE>` are added as Taskwarrior tags.
```
ryneeverett: Using `Issue.get_tags_from_labels`, as suggested below, aligns this functionality with that of other services. Probably also should do this with logseq. (Which makes label import an optional non-default behavior.)

```suggestion
Import Labels as Tags
+++++++++++++++++

Todoist allows you to attach `labels <https://www.todoist.com/help/articles/introduction-to-labels-dSo2eE>` to issues; to
use those labels as tags, you can use the ``import_labels_as_tags``
option:

.. config::
    :fragment: todoist

    todoist.import_labels_as_tags = True

Also, if you would like to control how these labels are created, you can
specify a template used for converting the Todoist label into a Taskwarrior
tag.

For example, to prefix all incoming labels with the string 'todoist_' (perhaps
to differentiate them from any existing tags you might have), you could
add the following configuration option:

.. config::
    :fragment: todoist

    todoist.label_template = todoist_{{label}}

In addition to the context variable ``{{label}}``, you also have access
to all fields on the Taskwarrior task if needed.

.. note::

   See :ref:`field_templates` for more details regarding how templates
   are processed.
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+log = logging.getLogger(__name__)
+
+
+class TodoistConfig(config.ServiceConfig):
+    service: typing_extensions.Literal["todoist"]
+    token: str
+    filter: str = None
+    char_open_bracket: str = "〈"
+    char_close_bracket: str = "〉"
+    inline_links: bool = True
```
ryneeverett:
```suggestion
    inline_links: bool = True
    import_labels_as_tags: bool = False
```

bugwarrior/services/todoist.py
ryneeverett: Looking briefly through some Todoist docs, it seems like the way you map due->scheduled and deadline->due is the implied suggestion. However, it seems totally valid for users to use Todoist's due date as an actual due date and they might be unpleasantly surprised at this behavior if for example their tasks don't show up in a report until they're overdue. So I'm thinking this should be an optional, non-default behavior.

bugwarrior/docs/services/todoist.rst
```diff
@@ -0,0 +1,98 @@
...
++++++++++++++++++
+
+By default bugwarrior incorporates the links into task description. To disable this behaviour either 
+modify the ``inline_links`` option in the main section to affect all services, or to modify for the todoist sevice only you can 
+set it in your todoist section.
+
+.. config::
+    :fragment: todoist
+    
+    todoist.inline_links = False
```
ryneeverett: Apologies if I asked this about Logseq too, but is there a particular reason you want this as a service option rather than using the main section configuration? You want links for some services but not others?

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+            token=self.config.token,
+            filter=self.config.filter,
+        )
+
+    def get_owner(self, issue):
+        # Issue assignment hasn't been implemented yet.
+        raise NotImplementedError(
+            "This service has not implemented support for 'only_if_assigned'."
+        )
+
```
ryneeverett: This is no longer a required method.

```suggestion
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,220 @@
...
+        }
+        for issue_iter in self.client.get_issues():
+            for issue in issue_iter:
+                extra = {
+                    "project": project_index[issue.project_id],
+                    "section": (
+                        section_index[issue.section_id]
+                        if issue.section_id in section_index.keys()
+                        else None
+                    ),
```
ryneeverett: This will have the same behavior. But is it actually possible for an issue to have a section that doesn't correspond to an active section?
```suggestion
                    "section": section_index.get(issue.section_id),
```

tests/test_todoist.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+
+    def setUp(self):
+        super().setUp()
+
+        self.service = self.get_mock_service(TodoistService)
+        self.service.client = mock.MagicMock(spec=TodoistClient)
+        self.service.client.get_issues = mock.MagicMock(
+            return_value=[self.test_record, self.test_extra]
+        )
```
ryneeverett: I'm pretty sure this is unneeded because it is overwritten in `test_issues`. I think the comparable lines could also be deleted in the logseq test.
```suggestion
```

ryneeverett (changes requested):

bugwarrior/services/todoist.py
scross01: The proper fix should be in taskw https://github.com/ralphbean/taskw/issues/172 to eliminate the need for this workaround.

bugwarrior/services/todoist.py
scross01: much cleaner, making the same change for project, assigner and assignee to use get()

bugwarrior/services/todoist.py
scross01: Todoist treats dates as timezone independent unless a timezone is specifically set, but the python date returned defaults to UTC so the timezone needs to be is removed.  Arguably this is probably something that should be handled better in the Todoist Python SDK to return the date without the timezone set.

bugwarrior/services/todoist.py
scross01: The Todosit API is returning either a `date` or a `datetime`.  The check here is specifically to see if it's a `datetime`

bugwarrior/services/todoist.py
scross01: Todoist does have comments, but they're not returned in the get_issues(), I've not looked at how to add them yet

bugwarrior/services/todoist.py
scross01: I think that makes sense. Need to think through what the choices would be:
- option 1: always due→due, ignore deadline
- option 2: due→due unless deadline is set, then due→scheduled and deadline→due
- option 3: always due→scheduled and deadline→due

bugwarrior/docs/services/todoist.rst
scross01: It can be removed here, I was just copying what I did for Logseq. I do see Logseq a special case because it uses a special URI logseq:// rather than a URL, and most terminals don't treat it as clickable.

bugwarrior/services/todoist.py
scross01: > it isn't adjusted to local time it's just set to midnight with no timezone, right?

Correct

bugwarrior/services/todoist.py
ryneeverett: Yeah, I suspect we'll end up either vendoring or replacing that dependency rather than actually fixing it upstream (#1078).

bugwarrior/services/todoist.py
ryneeverett: Ok, but a `date` is not a `datetime`:

```py
[ins] In [1]: import datetime

[ins] In [2]: today = datetime.date.today()

[ins] In [3]: isinstance(today, datetime.datetime)
Out[3]: False

[ins] In [4]: isinstance(today, datetime.date)
Out[4]: True
```

bugwarrior/services/todoist.py
ryneeverett: Ah, you might just add a TODO in this comment for now then.

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,216 @@
...
+            .replace("[", self.config.char_open_bracket)  # prevent &open; and &close;
+            .replace("]", self.config.char_close_bracket)
+        )
+
+    def get_priority(self):
+        return self.PRIORITY_MAP.get(
+            self.record.priority,
+            self.config.default_priority
+        )
+
```
ryneeverett: This is practically identical to the inherited `get_priority` method inherited from `Issue` -- sorry that was unclear.
```suggestion
```

scross01: The `self.record.get('priority')` from the inherited method fails because `record` is a `Task` and priority needs to be accessed as `self.record.priority`.

ryneeverett: Oh wow, I had missed that. I think it ought to be converted to a dict for consistency with other services. (And for conformity with the `Issue` type hint.)

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,216 @@
...
+        for users in users_iter:
+            for user in users:
+                all_users.append(user)
+        return all_users
+
+    def get_issues(self):
+        tasks_iter = self._api.filter_tasks(query=self.filter)
+        for tasks in tasks_iter:
+            for task in tasks:
+                yield task
```
ryneeverett:
```suggestion
                yield dataclasses.as_dict(task)
```

scross01: This seems rather fragile. Forcing `get_issues()` to return a dict rather than the native Todoist SDK class means all references now need to be accessed with `.get()` rather than as Todoist `Task` class attributes:
```
self.record.due -> self.record.get("due")
self.record.deadline -> self.record.get("deadline")
etc...
```
As `to_taskwarrior()` is the method for converting the Todoist Task representation to the Issue dict, and `get_priority()` is used within that method it seems more appropriate to override the default `get_priority()` to handle the Todosit specific implementation rather that force reuse of the inherited method.

bugwarrior/services/todoist.py
scross01: Giving this some more thought, I think option 2 should be the default (use due as due unless deadline is set).

Proposed approach: new option setting `todoist.due_date_mapping`
- `default` - due to due unless deadline, then due to scheduled and deadline to due
- `always_due` - always map due to due and ignore deadline
- `always_scheduled` - always map due to scheduled and deadline to due.

bugwarrior/services/todoist.py
ryneeverett: Note that `get_labels_from_tags` (which I'm also suggesting you use) also accesses the `record` field and assumes it is a dict.

However, the main motivation is not to use the inherited helper methods, but to keep a consistent API between `Service` and `Issue`. It may work now, but "bugwarrior core" might make changes in the future that assume that services follow the typing spec and I think that's a reasonable right to reserve.

> This seems rather fragile.

I'm not sure what the concern is -- that the Todoist SDK will change such that the class doesn't convert to a dict the same way?

> all references now need to be accessed with .get()

I would think you'd use regular dictionary lookup syntax (`self.record['field']`) -- `get()` is mostly for when you need to set a default value or the field might not exist but if you were ok with using `Task` class attributes without a fallback then you must be able to assume these fields exist.

scross01: Okay, i see what you mean, I just hit the issue with `get_labels_from_tags` before I saw your comment.

bugwarrior/services/todoist.py
ryneeverett: The problem I see with that approach is that people might use `due` as a "scheduled" date but not set `deadline` for every task. I think that while users will have different mapping preferences, the mapping should be consistent for all tasks.

I'd forgotten that we'd discussed this before in #385 and determined that the best course of action is generally to add such fields as due and deadline as UDA's. I don't think any other services do so but I wouldn't be opposed to documenting a suggested use of due_template/scheduled_template as a Service Feature if you wanted.

bugwarrior/services/todoist.py
scross01: switched to use `isinstance()`

bugwarrior/docs/services/todoist.rst
scross01: added

bugwarrior/services/todoist.py
scross01: The default behavior I added I think aligns best to the Todoist philosophy for dates vs deadlines [The difference between a date and a deadline](https://www.todoist.com/help/articles/introduction-to-deadlines-uMqbSLM6U#h_01JDS5KZG9AMRPBWBEK366TXGE), and the other options should cover user specific alternatives.  I'll add the UDA's as well.

The due_template/schedule_template is an interesting idea, but probably overkill

ryneeverett: > The due_template/schedule_template is an interesting idea, but probably overkill

All you have to do to get this functionality is add UDA's for due and schedule thanks to [field templates](https://bugwarrior.readthedocs.io/en/latest/common_configuration.html#field-templates).

scross01: UDA's added. Does that resolve this thread - or are any other changes needed?

bugwarrior/services/todoist.py
scross01: Comments are now added as annotations

bugwarrior/services/todoist.py
scross01: The code is now using `get_tags_from_labels()`.
One side affect is I have some labels with emoji's in the name, but the get_tags_from_labels replaces non alphanumeric characters with `_`.  Not a big deal though.

bugwarrior/services/todoist.py
scross01: `import_labels_as_tags` config option is now supported

bugwarrior/services/todoist.py
scross01: `record` is now a dict

bugwarrior/services/todoist.py
scross01: `record` is now a dict

bugwarrior/services/todoist.py
ryneeverett: Frankly, I think the `due_date_mapping` option should be removed and the mapping in logseq should likewise be removed.

I'm not really convinced that the default behavior matches the Todoist philosophy. I'd seen that article and to me it seems intentionally ambiguous to paper over the fact that their field semantics don't make any sense. I find it hard to believe that the default behavior represents the way people use Todoist -- using `deadline` as a due date for some tasks and `due` as a due date for other tasks. I would expect the meaning of these fields to be consistent within an organization.

The other point is that bugwarrior services don't generally import dates into date fields and we have (more or less successfully) been intentional about this since #385.

scross01: Okay, so:
- [x] Remove the `due_date_mapping` config option
- [x] only map Todoist due date to Taskwarrior due date. Ignore Todoist deadline dates. Nothing is mapped to taskwarrior scheduled date
- [x] Todoist Due and Deadline dates added as UDAs
- [x] Document how to use `due_template` and `scheduled_template` to alter the mapping

The following templates work the for behavior I was suggesting:
```
todoist.due_template = {{ todoistdeadline if todoistdeadline else todoistdue if todoistdue else "" }}
todoist.scheduled_template = {{ todoistdue if todoistdeadline else "" }}
```

bugwarrior/services/todoist.py
ryneeverett: > One side affect is I have some labels with emoji's in the name, but the get_tags_from_labels replaces non alphanumeric characters with `_`. Not a big deal though.

I suspect with modern terminal emulators and modern taskwarrior this behavior could be safely removed from `get_tags_from_labels`.

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,279 @@
...
+        self._api = TodoistAPI(token)
+        self.filter = filter
+
+    @classmethod
+    def task_to_dict(cls, task: Task):
+        record = asdict(task)
+        # add data items for additional properties
+        record["is_completed"] = task.is_completed
+        record["url"] = task.url
+        record["labels"] = task.labels
```
ryneeverett: This shouldn't be needed, because `labels` is a proper attribute of the dataclass: https://github.com/Doist/todoist-api-python/blob/e268cfc88c313d6219ae17fa6bc160103bb3c533/todoist_api_python/models.py#L111
```suggestion
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,279 @@
...
+                    self.config.filter
+                    + f" & (!assigned | shared & assigned to: {self.config.only_if_assigned})"
+                )
+            else:
+                filter = (
+                    self.config.filter
+                    + f" & (!shared | shared & assigned to: {self.config.only_if_assigned})"
+                )
+        else:
+            filter = self.config.filter
```
ryneeverett: I would find this easier to follow if it were a bit more concise. For example:
```suggestion
        filter = self.config.filter
        if self.config.only_if_assigned:
            assigned = "assigned" if self.config.also_unassigned else "shared"
            filter += f" & (!{assigned} | shared & assigned to: {self.config.only_if_assigned})"
```

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,279 @@
...
+        if self.config.only_if_assigned:
+            if self.config.also_unassigned:
+                filter = (
+                    self.config.filter
+                    + f" & (!assigned | shared & assigned to: {self.config.only_if_assigned})"
+                )
+            else:
+                filter = (
+                    self.config.filter
+                    + f" & (!shared | shared & assigned to: {self.config.only_if_assigned})"
```
ryneeverett: Are you sure `!shared` is right? I'm looking at [the Todoist filter docs](https://www.todoist.com/help/articles/introduction-to-filters-V98wIH) but I don't follow what "unshared" projects would be or why we need to include them here.

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,279 @@
...
+                ),
+            }
+            # optionally add comments as annotations
+            if self.main_config.annotation_comments:
+                comments = self.client.get_comments(issue["id"])
+                annotations = self.build_annotations(
+                    [(user_index.get(comment.poster_id), comment.content) for comment in comments],
+                    issue["url"]
+                )
+                extra["annotations"] = annotations
```
ryneeverett: Please follow the strong convention of assembling annotations in a separate method like (I believe) all the other services do. The field in `extra` would then be something like `"annotations": self.annotations(user_index, issue)`.

ryneeverett (changes requested):

bugwarrior/services/todoist.py
scross01: Todoist Personal projects (i.e. not shared) don't set the assigned id, so the logic here is the fetch all tasks in Personal projects (`!shared`) and all assigned tasks in shared projects.  When the `also_unassigned` is enabled then the `!assigned` is matching all unassigned tasks in both personal and shard projects, so the `!shared` isn't needed because all `!shared` is a subset of `!assigned`.

I've also confirmed that creating an assigned task in a shared project and then moving it to a Personal project removes the assignment, so personal project tasks are always unassigned. Assignment can only be done to tasks in shared projects.

bugwarrior/services/todoist.py
scross01: How about this.  The final filter is a bit more explicit to include both !shared and !assigned, but a bit clearer to follow.
```python
        # apply additional filters
        filter = self.config.filter
        if self.config.only_if_assigned:
            # fetch personal tasks (!shared)
            personal = "!shared"
            # fetch assigned tasks in shared projects (shared & assigned)
            shared_assigned = f"| shared & assigned to: {self.config.only_if_assigned}"
            # fetch unassigned tasks in shared projects (!assigned)
            unassigned = "| !assigned" if self.config.also_unassigned else ""
            filter += f" & ({personal} {shared_assigned} {unassigned})"
```

bugwarrior/services/todoist.py
scross01: removed

bugwarrior/services/todoist.py
scross01: added annotations method

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,281 @@
...
+
+from todoist_api_python.models import Task
+
+log = logging.getLogger(__name__)
+
+
+class TodoistConfig(config.ServiceConfig):
+    service: typing_extensions.Literal["todoist"]
+    token: str
+    filter: str = None
```
ryneeverett: I'm surprised pydantic isn't throwing errors here since None isn't a str. I thought you had to use `typing.Optional[str]` if you wanted to allow None. Hmm.

bugwarrior/services/todoist.py
scross01: Created https://github.com/GothenburgBitFactory/bugwarrior/issues/1104 to track proposal to move this to the based Issue class

bugwarrior/services/todoist.py
ryneeverett: Much better than the original or my suggestion, thanks! One question:

```py
            # fetch unassigned tasks in shared projects (!assigned)
            unassigned = "| !assigned" if self.config.also_unassigned else ""
```
I take it the `assigned` filter implicitly applies the `shared` filter, but would there be any harm in making it explicit to make the code more clear? (`"| shared & !assigned"`)

bugwarrior/services/todoist.py
scross01: Found a related issue - if filter is left as `None` and the `only_if_assigned` filter is appended then it fails because it's trying to appending the additional filter string to None.  I think the safest option is to specifically set the default filter to `"(view all)"` which gets all uncompleted tasks.

ryneeverett (approved):

bugwarrior/services/todoist.py
```diff
@@ -0,0 +1,286 @@
...
+            self.URL: self.record["url"],
+        }
+        return task
+
+    def get_default_description(self):
+        description = self.build_default_description(
+            title=self._unescape_content(self.record["content"]),
+            url=self.record["url"],
+            number=self.record["id"],
+            cls="issue",
```
ryneeverett:
```suggestion
            cls="task",
```

This will remove the "Is" from the description which probably doesn't make sense for this service.

scross01: Never realized there was different task types. I'll look at adding proper `subtask` support as well as Todoist has subtasks.

Should Logseq also change the task type from issue to `task` or `todo`?

bugwarrior/services/todoist.py
```diff
@@ -137,6 +138,10 @@ class TodoistIssue(Issue):
             "type": "string",
             "label": "Todoist URL",
         },
+        PARENT_ID: {
+            "type": "string",
+            "label": "Todoist Parent ID",
+        }
```
ryneeverett, line 144: Is this actually useful as a UDA?

bugwarrior/services/todoist.py
```diff
@@ -282,5 +288,6 @@ def issues(self):
                     else None
                 ),
                 "annotations": self.annotations(user_index, issue),
+                "parent_id": issue["parent_id"]
```
ryneeverett: We shouldn't need to add this to extra since it's already in the issue and can be accessed as `self.record['parent_id']` in the `Issue` class..

bugwarrior/services/todoist.py
ryneeverett: > Should Logseq also change the task type from issue to `task` or `todo`?

Yeah, I suspect both services are better aligned with `task`.

bugwarrior/services/todoist.py
scross01: Looking at the Jira service as an example that has subtasks it's setting the parent as a UDA. I can't find an example where `self.record['parent_id']` is getting used, but I tried setting it and it's breaking the subtask description generation:

Subtask description when `self.record['parent_id']` is not set.

```
(bw)Subtask ##6cV2xVwcxhxvC2mX - Test Sub Task 1 .. https://app.todoist.com/app/task/test-sub-task-1-6cV2xVwcxhxvC2mX 
```

Subtask description when `self.record['parent_id']` is set to the parent Id.

```
parent_id:"6cV2xV4W5VcxgHQX"
```

bugwarrior/services/todoist.py
scross01: Yes I think so. it can be used by a script to create and then open the parent task url using the ID.  For example in taskwarrior-tui I have a shortcut setup that runs a script to open the task in the web browser. Similar approach could add another shortcut to open then parent task for subtasks.

bugwarrior/services/todoist.py
ryneeverett: I'm not following what you mean by "when `self.record['parent_id']` is [not] set". The code here suggests that parent_id is *always* a key in `self.record` because it accesses `issue["parent_id"]` and then passes `issue` as the foreign record (`self.record` in `Issue`).

scross01: My bad, I miss-read your suggestion thinking you meant to that `parent_id` was already attribute of the Taskwarrior Issue class, the weird description behavior was occurring when trying to set it in `to_taskwarrior()`

You are correct, the extra isn't needed, as it's in the Todoist record. Pushed the changes.

ryneeverett (approved):

ryneeverett:
rebased

scross01:
Thanks @ryneeverett for all the review feedback and guidance.

ryneeverett:
And thanks @scross01 for accepting a lot of feedback and perhaps a little nitpicking.

PR #1106, opened 2025-07-24
Add support for syncing to Linear.app

### LINKED ISSUE #1088: Linear support
This is a great project! I wanted to highlight that Linear is quite popular nowadays. It is modern, quick, and seamless compared to JIRA and other issue-tracking apps. It would be great if we had linear support. 

Thanks.

Linear - https://linear.app/

djmitche:
This is a pretty simple GraphQL API, so no additional libraries are required.

### DIFF
--- a/bugwarrior/docs/services/linear.rst
+++ b/bugwarrior/docs/services/linear.rst
@@ -1,0 +1,44 @@
+Linear
+======
+
+You can import tasks from Linear.app using the ``linear`` service name.
+
+
+Example Service
+---------------
+
+Here's an example of a Linear target:
+
+.. config::
+
+    [my_issues]
+    service = linear
+    linear.api_token = <your API token>
+
+The above example is the minimum required to import issues from Linear.  You
+can also feel free to use any of the configuration options described in
+:ref:`common_configuration_options`.
+
+.. describe:: api_token
+
+    Linear offers API keys at Settings -> Security & Access -> Personal API
+    Keys. You will need to provide ``api_token`` to allow bugwarrior to read
+    issues. The token needs only Read access.
+
+.. describe:: import_labels_as_tags
+
+    A boolean that indicates whether the Linear labels should be imported as
+    tags in taskwarrior. (Defaults to false.)
+
+.. describe:: label_template
+
+   Template used to convert Linear labels to taskwarrior tags.
+   See :ref:`field_templates` for more details regarding how templates
+   are processed.
+   The default value is ``{{label|replace(' ', '_')}}``.
+
+Provided UDA Fields
+-------------------
+
+.. udas:: bugwarrior.services.linear.LinearIssue
+
--- a/bugwarrior/docs/services/linear.rst
+++ b/bugwarrior/docs/services/linear.rst
@@ -19,6 +19,9 @@ The above example is the minimum required to import issues from Linear.  You
 can also feel free to use any of the configuration options described in
 :ref:`common_configuration_options`.
 
+Options
+----------
+
 .. describe:: api_token
 
     Linear offers API keys at Settings -> Security & Access -> Personal API
--- a/bugwarrior/services/linear.py
+++ b/bugwarrior/services/linear.py
@@ -54,7 +54,7 @@ class LinearIssue(Issue):
 
     def parse_date(self, input):
         '''Parse a date, stripping microseconds'''
-        parsed = super(LinearIssue, self).parse_date(input)
+        parsed = super().parse_date(input)
         if parsed:
             parsed = parsed.replace(microsecond=0)
         return parsed
--- a/bugwarrior/services/linear.py
+++ b/bugwarrior/services/linear.py
@@ -21,6 +21,7 @@ class LinearConfig(config.ServiceConfig):
     )
     import_labels_as_tags: bool = False
     label_template: str = "{{label|replace(' ', '_')}}"
+    also_unassigned: config.UnsupportedOption[bool] = False
 
 
 class LinearIssue(Issue):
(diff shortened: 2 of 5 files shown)


djmitche:
Whoops, I missed a few places where "Story" is still mentioned!

djmitche:
> Whoops, I missed a few places where "Story" is still mentioned!

fixed now :)

bugwarrior/docs/services/linear.rst
```diff
@@ -0,0 +1,44 @@
...
+.. config::
+
+    [my_issues]
+    service = linear
+    linear.api_token = <your API token>
+
+The above example is the minimum required to import issues from Linear.  You
+can also feel free to use any of the configuration options described in
+:ref:`common_configuration_options`.
+
```
ryneeverett, line 21:
```suggestion

Options
----------

```

bugwarrior/services/linear.py
```diff
@@ -0,0 +1,181 @@
...
+        CREATED_AT: {"type": "date", "label": "Issue Created"},
+        UPDATED_AT: {"type": "date", "label": "Issue Updated"},
+        CLOSED_AT: {"type": "date", "label": "Issue Closed"},
+    }
+
+    UNIQUE_KEY = (URL,)
+
+    def parse_date(self, input):
+        '''Parse a date, stripping microseconds'''
+        parsed = super(LinearIssue, self).parse_date(input)
```
ryneeverett:
```suggestion
        parsed = super().parse_date(input)
```

bugwarrior/services/linear.py
```diff
@@ -0,0 +1,181 @@
...
+        created = self.parse_date(self.record.get("createdAt"))
+        modified = self.parse_date(self.record.get("updatedAt"))
+        closed = self.parse_date(self.record.get("completedAt"))
+
+        # Get a value, defaulting empty results to the given default.
+        def get(v, k, default=None):
+            r = v.get(k, default)
+            if not r:
+                return default
+            return r
```
ryneeverett: This is a strange looking pattern. Your comment explains what it does but it would be nice to also note why it's needed. I deduce that `null` ends up as an empty string for "empty" results when it would otherwise be a structured dictionary or list?

I don't recall how bugwarrior handles `None` versus an empty string here but it is interesting to note in the tests that null ends up as an empty string for some fields (e.g. projects) and as `None` for others (e.g. creator).

bugwarrior/services/linear.py
```diff
@@ -0,0 +1,181 @@
...
+class LinearService(Service, Client):
+    ISSUE_CLASS = LinearIssue
+    CONFIG_SCHEMA = LinearConfig
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.session = requests.Session()
+        self.session.headers.update(
+            {"Authorization": self.config.api_token, "Content-Type": "application/json"}
```
ryneeverett: This enables keyring support.
```suggestion
            {"Authorization": self.get_password("api_token"), "Content-Type": "application/json"}
```

bugwarrior/services/linear.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+
+        filter = "{}"
+        if self.config.only_if_assigned and not self.config.also_unassigned:
```
ryneeverett: if `also_unassigned` is unsupported then we can remove this check and set `also_unassigned: config.UnsupportedOption[bool] = False` in the config class.

ryneeverett (changes requested):

bugwarrior/services/linear.py
djmitche: Huh, my Python is rusty!

bugwarrior/services/linear.py
djmitche: I intend to support it - it's one of the `common_configuration_options`. Should it also be included in LinearConfig? I didn't see it there for PivotalTracker.

bugwarrior/services/linear.py
djmitche: IIRC it was for `project` in particular which is either `null` or `{"name": ..}`, so a simple `self.record.get("project", {})` can still result in an error.

bugwarrior/services/linear.py
```diff
@@ -0,0 +1,181 @@
...
+        # Get a value, defaulting empty results to the given default.
+        def get(v, k, default=None):
+            r = v.get(k, default)
+            if not r:
+                return default
+            return r
+
+        return {
+            "project": re.sub(
+                r"[^a-zA-Z0-9]", "_", get(get(self.record, "project", {}), "name", "")
```
djmitche, line 81: This is an empty string since it's an argument to `re.sub`. That came from PivotalTracker, but it looks like project is always set there. I'll experiment a little more.

bugwarrior/services/linear.py
ryneeverett: I know you based this off PivotalTracker but frankly that service is somewhat anomalous. Most services are based on the Github service.

Every service should either support this option with the correct behavior or throw a validation error if it is configured. It seems like the current behavior is that `only_if_assigned` is ignored if `also_unassigned` is set, which is wrong.

bugwarrior/services/linear.py
```diff
@@ -178,4 +185,7 @@ def get_issues(self):
             messages = [error.get("message", "Unknown error") for error in res['errors']]
             raise ValueError(messages.join("; "))
 
-        return res.get("data", {}).get("issues", {}).get("nodes", [])
+        r = res.get("data", {}).get("issues", {}).get("nodes", [])
+        import pprint
+        pprint.pprint(r)
```
ryneeverett:
```suggestion
```

bugwarrior/services/linear.py
djmitche: Got it, thanks!

ryneeverett (approved):

djmitche:
(rebased onto latest develop, which no longer has `setup.py`)

djmitche:
Thanks as always for your thorough reviews!

PR #1114, opened 2025-07-30
Modernize formatting and testing

edmundmiller:
- **refactor: modernize codebase with ruff and pytest**
- **fix: complete pytest modernization and uv setup**
- **feat: complete pytest conversion for config test files**


ryneeverett:
Frankly, I am not interested in reviewing a "grab bag of changes" PR, regardless of whether it is generated with AI or written by a chaotic human hand.

Briefly scanning, I notice changes which seem of questionable value:

- Is flake8's half-second runtime a performance problem?
- Should we actually get further locked in to pytest?

There are probably a few good things in here but I doubt it's worth the effort to extract them.

PR #1115, opened 2025-07-30
Allow filtering tasks by status and status type

djmitche:
Now that I've actually finished an issue at $JOB, I realized it doesn't filter out finished issues!

Linear's API uses "state" while the docs use "status". I have used "status" in the user-facing configuration.


bugwarrior/services/linear.py
```diff
@@ -19,6 +19,10 @@ class LinearConfig(config.ServiceConfig):
     host: config.StrippedTrailingSlashUrl = config.StrippedTrailingSlashUrl(
         "https://api.linear.app/graphql", scheme="https", host="api.linear.app"
     )
+    statuses: config.ConfigList = config.ConfigList([])
+    status_types: config.ConfigList = config.ConfigList(
+        ["backlog", "unstarted", "started"]
+    )
     import_labels_as_tags: bool = False
     label_template: str = "{{label|replace(' ', '_')}}"
     also_unassigned: config.UnsupportedOption[bool] = False
```
ryneeverett: It would be better to enforce option compatibility constraints rather than rely on documentation. I think this would work:

```suggestion
    statuses: config.ConfigList = config.ConfigList([])
    status_types: typing.Optional[config.ConfigList] = None
    import_labels_as_tags: bool = False
    label_template: str = "{{label|replace(' ', '_')}}"
    also_unassigned: config.UnsupportedOption[bool] = False
    
    @pydantic.v1.root_validator
    def statuses_or_status_types(cls, values):
        if values["statuses"] and values["status_types"]:
            raise ValueError("statuses and status_types are incompatible")
        if not values["statuses"] and not values["status_types"]: # set default value
            values["status_types"] = ["backlog", "unstarted", "started"]
        return values
```

bugwarrior/docs/services/linear.rst
```diff
@@ -40,6 +40,24 @@ Options
...
    The default value is ``{{label|replace(' ', '_')}}``.
 
+.. describe:: status_types
+
+   Comma-separated list of Linear status types that should be included. The
+   default statuses for a team and any custom statuses fall into one of a few
+   pre-defined types: `backlog`, `unstarted`, `started`, `completed`, and
+   `canceled`, and are case-sensitive. See
+   https://linear.app/docs/configuring-workflows#overview. The default value is
+   ``backlog,unstarted,started``.
```
ryneeverett: Using the config directive is especially preferred for structured values since they're formatted pretty differently in toml.

```suggestion
   https://linear.app/docs/configuring-workflows#overview. The default value is:
   
   .. config::
       :fragment: linear

       linear.status_types = backlog,unstarted,started
```

bugwarrior/docs/services/linear.rst
```diff
@@ -40,6 +40,24 @@ Options
...
+
+.. describe:: statuses
+
+   If filtering by status types is not sufficient, this option allows filtering
+   by a comma-separated list of status names. For example, if your team has a
+   "Not A Priority" status with type "Canceled", which you would like to
+   include, use ``linear.statuses = Backlog,Todo,In Progress,Not A Priority``.
```
ryneeverett:
```suggestion
   include, use:
   
   .. config::
       :fragment: linear
       
       linear.statuses = Backlog,Todo,In Progress,Not A Priority
       
```

bugwarrior/docs/services/linear.rst
```diff
@@ -40,6 +40,24 @@ Options
...
+   These values are case-sensitive. If this configuration is provided,
+   ``status_types`` is ignored. The default value is empty.
```
ryneeverett: Assuming the mutual exclusivity is validated, the wording should be adjusted.

```suggestion
   These values are case-sensitive. This setting overrides the default value of
   ``status_types``.
```

ryneeverett (changes requested):

djmitche:
Thanks -- I wasn't aware of either of those bits of functionality!

ryneeverett (approved):

PR #1116, opened 2025-07-30
Use an email address for the assignee and creator

djmitche:
These UDAs had the assignee / creator shown as user name ("Dustin Mitchell") but email is probably more appropriate.


ryneeverett (approved):
I think the most typical thing a service would use is a "handle" ("djmitche") but I agree that email is more closely aligned than name if that's not a thing in linear.

djmitche:
AFAICT that's Linear's unique ID for a person. Using name was just an oversight of mine while exploring the GraphQL.

PR #1117, opened 2025-08-01
fix: 🚑️ config_filename was missing

massa:
In my setup, `taskrc` was located as `${XDG_CONFIG_HOME}/task/taskrc` and the `get_converted_hours` function was panic'ing because it couldn't open `${HOME}/.taskrc` ...

### DIFF
--- a/bugwarrior/services/redmine.py
+++ b/bugwarrior/services/redmine.py
@@ -220,7 +220,7 @@ class RedMineIssue(Issue):
         return self.config.url + "/issues/" + str(self.record["id"])
 
     def get_converted_hours(self, estimated_hours):
-        tw = TaskWarriorShellout()
+        tw = TaskWarriorShellout(config_filename=self.main_config.taskrc)
         calc = tw._execute('calc', estimated_hours)
         return (
             calc[0].rstrip()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,7 +95,7 @@ todoist = "bugwarrior.services.todoist:TodoistService"
 bugwarrior = "bugwarrior.config.ini2toml_plugin:activate"
 
 [tool.setuptools]
-packages = ["bugwarrior"]
+packages = ["bugwarrior", "bugwarrior.config", "bugwarrior.services"]
 
 [build-system]
 requires = ["setuptools"]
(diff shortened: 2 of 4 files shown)


bugwarrior/services/redmine.py
```diff
@@ -220,7 +220,8 @@ def get_issue_url(self):
         return self.config.url + "/issues/" + str(self.record["id"])
 
     def get_converted_hours(self, estimated_hours):
-        tw = TaskWarriorShellout()
+        cfg = config.load_config('general', False, False)
+        tw = TaskWarriorShellout(config_filename=cfg['general'].taskrc)
```
ryneeverett: The configuration should already be loaded and available in `self.main_config`:
```suggestion
        tw = TaskWarriorShellout(config_filename=self.main_config.taskrc)
```

pyproject.toml
```diff
@@ -95,7 +95,7 @@ todoist = "bugwarrior.services.todoist:TodoistService"
 bugwarrior = "bugwarrior.config.ini2toml_plugin:activate"
 
 [tool.setuptools]
-packages = ["bugwarrior"]
+packages = ["bugwarrior", "bugwarrior.config", "bugwarrior.services"]
```
ryneeverett, line 98: Is this needed?

ryneeverett (changes requested):

ryneeverett:
Thanks for bringing this to our attention. When I started looking into this I realized that we ought to be checking $XDG_CONFIG_HOME/task/taskrc so I submitted #1118. With that change you should no longer have to use the [general] -> taskrc option to use $XDG_CONFIG_HOME/task/taskrc.

pyproject.toml
massa: the package installed with uvx does not come with those modules, unless I include them... I know there is some difference in setuptools versions, but I hadn't had the time to solve it.

bugwarrior/services/redmine.py
massa: Thank you! In the end, I just adopted both this solution **and** your factory commit...

pyproject.toml
ryneeverett: Thanks again! After reading the setuptools docs, I'm surprised other tools didn't have a problem with this either. I'd rather use package discovery than list each subpackage. This is implemented in #1119 which i confirmed fixes uvx.

ryneeverett:
Merged in 963fa8720668f0a773670ac88ddb482054f19b80.

PR #1123, opened 2025-09-06
Queries all issues from Jira not just the default 50

fuero:
Apparantly the python jira module changed its API and the parameter must evaluate to `false` now.

### DIFF
--- a/bugwarrior/services/jira.py
+++ b/bugwarrior/services/jira.py
@@ -434,7 +434,7 @@ class JiraService(Service):
         )
 
     def issues(self):
-        cases = self.jira.search_issues(self.query, maxResults=None)
+        cases = self.jira.search_issues(self.query, maxResults=False)
 
         for case in cases:
             issue = self.get_issue_for_record(


ryneeverett (approved):
This seems reasonable since the python jira library suggests "If maxResults evaluates to False, it will try to get all issues in batches."

However, this isn't a recent change and the python library itself seems to always evaluate maxResults's falsiness (and None is falsey) so I suspect if this caused a recent breakage it is coming from the jira service itself since the python library seems to pass this value to the jira API.

As far as the [API docs](https://developer.atlassian.com/server/jira/platform/rest/v10005/api-group-search/#api-group-search) go, the result of either value would seem to be ambiguous. If this gives anyone more trouble perhaps it would be better to switch to `5000` which would seem to be the jira service's actual maximum? :man_shrugging:

ryneeverett:
@srl295 Any thoughts on this? Seems like it could plausibly be related to your issue.

srl295:
> @srl295 Any thoughts on this? Seems like it could plausibly be related to your issue.

yeah maybe related to #1124 you mean?

Possibly although in my case it failed completely with that call, so maybe something changed 2 weeks ago and then more recently.

If you look at the suggested client-code patch from the `jira` python library it includes a pagination loop:

```py
    …
    nextPageToken=next_page_token,
```

I don't know if that's handled by that library automatically or not. but it's possible that bugwarrior will have to add a pagination loop.

srl295 (approved):
nevertheless it seems like a reasonable change

srl295:
OK ,i can confirm my local copy (other PR) is only getting 50 tickets…

PR #1139, opened 2025-11-15
Service API: add a version number and enforce it

ryneeverett:
One of the lessons I've learned from observing taskchampion development is that "lazy spec versioning" isn't always such a great idea. Analogous is https://github.com/GothenburgBitFactory/taskchampion/issues/477, in which the concern is that older versions of taskchampion don't know how to handle newer versions of the database. Likewise, older versions of bugwarrior might not handle newer implementations of the service spec properly. We also adopt a comparable major/minor version scheme.

### DIFF
--- a/bugwarrior/docs/other-services/api.rst
+++ b/bugwarrior/docs/other-services/api.rst
@@ -1,5 +1,5 @@
-Python API
-==========
+Python API v1.0
+===============
 
 The interfaces documented here are considered stable. All other interfaces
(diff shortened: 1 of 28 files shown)


djmitche (approved):
Good call! This will be particularly important for third-party extensions, I assume..

ryneeverett:
Yes -- I assume the services maintained in this repository will always be on the latest version. Versioning is entirely about playing well with third party services.

PR #1149, opened 2025-11-23
fix: limit python to <=3.11, require setuptools

jochumdev:
Both are needed to make an install with uvx as seamless as possible.

Python 3.12 and later give:

```text
/home/jochumdev/.local/share/uv/tools/bugwarrior/lib/python3.12/site-packages/kitchen/text/converters.py:804: SyntaxWarning: invalid escape sequence '\*'
  :arg \*args, \*\*kwargs: extra arguments to this function are passed on to
/home/jochumdev/.local/share/uv/tools/bugwarrior/lib/python3.12/site-packages/kitchen/text/converters.py:836: SyntaxWarning: invalid escape sequence '\*'
  :arg \*args, \*\*kwargs: extra arguments to this function are passed on to
```

This command install bugwarrior without this PR on any platform uv supports:

```sh
uv tool install --python=3.11 git+https://github.com/GothenburgBitFactory/bugwarrior --with setuptools
```

After a merge it should be possible to do:

```sh
uvx git+https://github.com/GothenburgBitFactory/bugwarrior
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,8 @@ name = "bugwarrior"
 version = "2.0.0"
 description = "a command line utility for updating your local taskwarrior database from your forge issue trackers"
 readme = "bugwarrior/README.rst"
-requires-python = ">=3.10,<4"
+# pydantic.v1 does not support python-3.14 See <https://github.com/GothenburgBitFactory/bugwarrior/issues/998>.
+requires-python = ">=3.10,<3.14"
 license = "GPL-3.0-or-later"
 license-files = ["LICENSE.txt"]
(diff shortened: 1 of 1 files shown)


ryneeverett (changes requested):
I see that adding setuptools is necessary for uvx and that's a reasonable change. But reducing the supported python version for everyone -- regardless of installation tool -- just to silence some warnings from a nested dependency isn't a good tradeoff. FYI, the `-q` flag also works to that end: `uvx -q --python=3.13 --with setuptools git+https://github.com/GothenburgBitFactory/bugwarrior`.

Issue #1151, opened 2025-11-24
Improve tooling and add formatter


Lotram:
Hello,

While having a closer look at the code before trying to start a MR for https://github.com/GothenburgBitFactory/bugwarrior/issues/998, I noticed there is no formatter in this repo. I then considered adding one (and maybe a type checker ?), I then saw you @ryneeverett rejected this MR https://github.com/GothenburgBitFactory/bugwarrior/pull/1114 . 

I can get why you refused it, since there is no description, no comment, and it looks like it was mostly done by an AI without human review, still I think using ruff could be beneficial here, at least to format the code. black + isort is also an option, though I think ruff makes more sense in 2025. 

I'm open to do it myself,  but I wanted to check with you first, as a maintainer, whether you're open to such a change, and if so, how to do it to ease the review. 

I'm suggesting the following changes, they can of course be done separately:
* switch to ruff as both a formatter and a linter
* use a type-checker (either mypy, pyright, or astral's ty)
* add pre-commits for these tools, as there is already a pre-commit config.

ryneeverett:
Thanks for opening this! Your assessment of my rejection of that other pull request is correct.

I looked into ruff and I think it makes a lot of sense. I hope you don't mind that I'm implementing it myself (#1152). I wanted to play around with it a bit before giving you the green light and by the time I was done there wasn't much left to do.

As for type checking, I'm for it. I hadn't seen astral ty but it seems like probably the best choice. I'm currently using pywright in my editor to avoid introducing more issues. But really, if we can get any tool "passing" then we can start running it in tests and avoid future regressions which would be great. My only real concern here is that I hope we don't have to add too much complex boilerplate that makes the code less readable. I haven't played much with typing since the early mypy days though, so I'm assuming the situation has gotten better in this regard.

I'm less convinced of the value of pre-commit but I'm not entirely against it. Rust was really my first positive exposure to formatters and looking at our rust projects [taskchampion](https://github.com/GothenburgBitFactory/taskchampion) and [taskchampion-sync-server](https://github.com/GothenburgBitFactory/taskchampion-sync-server), they do make use of rustfmt but not precommit and it rarely seems to be a problem. Maybe I'm missing something here though...

I'd be thrilled if somebody took on #998 and I can see why you'd want to improve the tooling before taking it on.

Lotram:
[...]

PR #1152, opened 2025-11-25
Ruff

ryneeverett:
Use ruff, per the suggestion in #1151.


pyproject.toml
```diff
@@ -97,6 +97,13 @@ todoist = "bugwarrior.services.todoist:TodoistService"
 [project.entry-points."ini2toml.processing"]
 bugwarrior = "bugwarrior.config.ini2toml_plugin:activate"
 
+[tool.ruff.format]
+quote-style = "single"
+
+[tool.ruff.lint]
+# Default value plus pycodestyle warnings.
+select = ["E4", "E7", "E9", "F", "W"]
```
Lotram: Consider adding "I" (for isort), so imports are automatically sorted

pyproject.toml
```diff
@@ -97,6 +97,13 @@ todoist = "bugwarrior.services.todoist:TodoistService"
 [project.entry-points."ini2toml.processing"]
 bugwarrior = "bugwarrior.config.ini2toml_plugin:activate"
 
+[tool.ruff.format]
+quote-style = "single"
```
Lotram: ruff default value, and the usual python standard (statement open to debate) is to use double quotes, though there is nothing wrong with using single quotes I guess

ryneeverett: We're mostly using single quotes at the present. This tweak reduces the diff from >5k lines to >3k lines. But I also see the advantage of going with "the usual" if that's what people are more used to these days...

ryneeverett: By switching to "preserve" the diff is reduced to >2k lines. I'm not sure there's much of an advantage to enforcing project-wide quote consistency but I'm open to revisiting the question.

bugwarrior/notifications.py
```diff
@@ -4,11 +4,9 @@
 
 import requests
 
-
 cache_dir = os.path.expanduser(os.getenv('XDG_CACHE_HOME', "~/.cache") + "/bugwarrior")
 logo_path = cache_dir + "/logo.png"
-logo_url = "https://upload.wikimedia.org/wikipedia/" + \
-    "en/5/59/Taskwarrior_logo.png"
+logo_url = "https://upload.wikimedia.org/wikipedia/" + "en/5/59/Taskwarrior_logo.png"
```
ryneeverett: Kind of silly to leave it like this.

ryneeverett:
While skimming through the diff I made a number of adjustments to the configuration which further reduced the diff by aligning better with the existing style (and my own sensibilities). There's a few things I still don't care for -- newlines before closing parenthesis/brackets, indentation/newline for every nested parenthesis/bracket, spaces around operators -- but overall it seems like an improvement to the style by catching all the inconsistencies.

I ended up being very glad I chose to "preserve" the existing single/double quotes because it would have been hard to skim with all that distraction and the skimming led me not only to the configuration adjustments but to being able to see the places where cleanup from the formatting transformation (and preemptive adjustments before it) made sense. If we were to unify the quote style in the future there would be little need to review the diff.

@Lotram I'm sure you'd prefer a vanilla configuration, but I think this is a reasonable compromise for now. I'm happy to revisit changing any of these settings in the future, but I'd want feedback from other contributors. Unless you have further comments I will merge as is so i can unblock your other proposals in #1151.

Lotram:
> @Lotram I'm sure you'd prefer a vanilla configuration, but I think this is a reasonable compromise for now. I'm happy to revisit changing any of these settings in the future, but I'd want feedback from other contributors. Unless you have further comments I will merge as is so i can unblock your other proposals in #1151.

Indeed, even tough I would have preferred a cleaner config, that's still a huge benefit, and I understand your point. 

To be honest, I know a lot of people try to limit the number of diffs as much as possible, I don't really see the point when that's a commit we won't review anyway (because it's just formatting, and I trust the formatters to do their job without breaking the code), except maybe to avoid adding noise when using git blame / git history. 

But I don't contribute to open source that much, so I may not have all the right use-cases in mind, and you're the maintainer here, so I'm ok with this.

ryneeverett:
Thanks for agreeing to the compromise.

> To be honest, I know a lot of people try to limit the number of diffs as much as possible, I don't really see the point when that's a commit we won't review anyway (because it's just formatting, and I trust the formatters to do their job without breaking the code), except maybe to avoid adding noise when using git blame / git history.

I actually did review this diff, though not thoroughly. My point is that if we were to unify the quotations in the future, *that* would be a diff I would not review. While ruff didn't break the code, it did make some formatting worse like `"some string" + "another string"` on the same line or leaving continuations of dictionary items unindented which I addressed in the third commit.

PR #1155, opened 2025-12-03
wip migrate config to pydantic v2

Lotram:
First draft for #998 

This is a wip commit. I've mostly migrated the schema.py file. For now, I've copied everything in schema_v2, and split it into multiple files. I've added a bunch of tests (helped by cursor), and copied the ones from test_schema.py

I tried to avoir doing to many changes, while migrating to more standard pydantic. I think we can further improve the config implementation, with less dynamic class creations (so easier typing), but I'll do that in another PR, if validated.


Lotram:
@ryneeverett I'm obviously not done, but I created the PR so you can have a look if you want. I created a bunch of small files on purpose, but next thing I'll do is merge my changes back to the main schema.py file, so changes can be properly reviewed.

bugwarrior/config/schema_v2/config.py
```diff
@@ -0,0 +1,139 @@
...
+    '_ServiceConfig',
+    __base__=BaseConfig,
+    **{
+        f'{key}_template': (typing.Optional[str], None)
+        for key in taskw.task.Task.FIELDS
+    },
+)
+
+
+# NOTE: I removed the deprecated fields
```
ryneeverett: FYI, these ought to be added back since we don't want breaking changes in the next release (2.1), which I'm thinking will be shortly after pydantic-2 support lands.

tests/config/test_schema_v2/test_models.py
```diff
@@ -0,0 +1,52 @@
...
+        self.assertFalse(notifications.notifications)
+        self.assertIsNone(notifications.backend)
+
+    def test_valid_backend(self):
+        notifications = schema_v2.Notifications(backend='gobject')
+        self.assertEqual(notifications.backend, 'gobject')
+
+    def test_invalid_backend(self):
+        with self.assertRaises(pydantic.ValidationError):
+            schema_v2.Notifications(backend='invalid')
```
ryneeverett: I've spot checked a few places in the tests and some of them seem somewhat valuable while others don't. I'm cherry-picking TestNotifications as an example of tests that don't seem valuable to me. It seems to me that many of the tests are encoding implementation details that we don't care about in the long run and my concern is that we'll become desensitized to just deleting tests when they fail. Right now, every assertion in the test suite is there because somebody thought it was important.

ryneeverett (commented):
Overall plan sounds good, thanks for sharing your progress!

tests/config/test_schema_v2/test_models.py
Lotram: Makes sense. I added lots of tests mostly to ensure the migrated models were working as expected. We can drop most of them, if not all, before merging this.

bugwarrior/config/schema_v2/config.py
Lotram: Ok, I'll revert this

ryneeverett:
If you're going to remove the python version constraint, please also add 3.14 to the CI matrix in `.github/workflows/bugwarrior.yml`. Note that we don't know if pydantic is the only problem with 3.14 but you'll find out soon enough if there's another major problem.

Looking at the failed CI api-change-check, it looks like for the most part the changes to the API docs are fine, if not improvements. I'll be reviewing the diff more carefully later but I don't expect it to be green. (BTW, as somebody who is deep into python and knows about best practices, perhaps you know if there's a better way to avoid accidentally breaking our "public" API? Something like [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) for python?)

Lotram:
> perhaps you know if there's a better way to avoid accidentally breaking our "public" API? 

Nope, nothing that I'm aware of. I don't work on heavily used CLIs though, but I don't know any well established tool to ensure there is no breaking change.

bugwarrior/config/schema.py
```diff
@@ -151,17 +179,14 @@ class Config(PydanticConfig):
...
+    # optional
+    taskrc: TaskrcPath = Field(default_factory=get_default_taskrc)
 
-    @pydantic.v1.root_validator
-    def compute_data(cls, values):
-        values['data'] = BugwarriorData(get_data_path(values['taskrc']))
-        return values
+    @computed_field
+    @property
+    def data(self) -> BugwarriorData:
```
ryneeverett, line 150: Noticed that the data attribute was removed from the docs in the api-change-check. I think this will put it back.

```suggestion
    def data(self) -> BugwarriorData:
        """Local data storage."""
```

bugwarrior/config/ini2toml_plugin.py
```diff
@@ -22,42 +27,46 @@ def to_type(section: IntermediateRepr, key: str, converter: typing.Callable):
...
+        ]
+        if len(non_null_types) == 1:
+            return non_null_types[0]
+    return None
+
+
+# NOTE: not sure about this one, but model_json_schema is definitely different than previous function
+def convert_section(section: IntermediateRepr, schema: type[pydantic.BaseModel]):
+    for prop, attrs in schema.model_json_schema()['properties'].items():
+        field_type = get_field_type(attrs)
```
ryneeverett: What is the reasoning behind this change? `schema.schema()['properties']` still seems to have what we want and doesn't require a whole method to extract the type.

bugwarrior/config/load.py
```diff
@@ -63,7 +63,7 @@ def parse_file(configpath: str) -> dict:
         config = {}
         for section in rawconfig.sections():
             if section in ['hooks', 'notifications']:
-                config[section] = rawconfig[section].items()
+                config[section] = dict(rawconfig[section].items())
```
ryneeverett:
```suggestion
                config[section] = dict(rawconfig[section])
```

bugwarrior/config/ini2toml_plugin.py
```diff
@@ -22,42 +27,46 @@ def to_type(section: IntermediateRepr, key: str, converter: typing.Callable):
...
-    """
-
-    bool_value: bool
+# Use Pydantic to convert various strings to booleans.
+_bool_adapter = pydantic.TypeAdapter(bool)
 
 
 def to_bool(section: IntermediateRepr, key: str):
-    to_type(section, key, lambda val: BooleanModel(bool_value=val).bool_value)
+    to_type(section, key, _bool_adapter.validate_python)
```
ryneeverett:
```suggestion
    to_type(section, key, pydantic.TypeAdapter(bool).validate_python)
```

bugwarrior/config/ini2toml_plugin.py
```diff
@@ -22,42 +27,46 @@ def to_type(section: IntermediateRepr, key: str, converter: typing.Callable):
...
-    """
-    Use Pydantic to convert various strings to booleans.
-
-    "True", "False", "yes", "no", etc.
-    Adapted from https://docs.pydantic.dev/usage/types/#booleans
-    """
-
-    bool_value: bool
+# Use Pydantic to convert various strings to booleans.
+_bool_adapter = pydantic.TypeAdapter(bool)
```
ryneeverett:
```suggestion
```

bugwarrior/config/schema.py
```diff
@@ -16,133 +28,149 @@
...
-    @classmethod
-    def validate(cls, value, field, config):
-        return super().validate(value.rstrip('/'), field, config)
+def validate_url(url: str):
+    return str(AnyUrl(url)).rstrip("/")
 
 
-class UrlSchemeError(pydantic.v1.UrlSchemeError):
-    msg_template = "URL should not include scheme ('{scheme}')"
+# NOTE: breaking change in pydantic v2: AnyUrl is not longer a subclass of str
```
ryneeverett: Therefore we're making our own type derived from str here. Got it.

```suggestion
```

bugwarrior/config/schema.py
```diff
@@ -16,133 +28,149 @@
...
-        user = parts['user']
-        if cls.user_required and user is None:
-            raise pydantic.v1.errors.UrlUserInfoError()
+    # NOTE: is this needed ?
+    try:
+        AnyUrl(url_with_scheme)
+    except Exception as exc:
+        raise PydanticCustomError(
+            "url_parsing", "Invalid URL format: {error}", {"error": str(exc)}
+        ) from exc
```
ryneeverett: I think your point here is correct -- apart from the scheme and the `://`, there don't seem to be any constraints on url validity. I believe that's a change from pydantic-v1 but probably a good one.

```suggestion
```

bugwarrior/config/schema.py
```diff
@@ -16,133 +28,149 @@
...
 
-        return parts
+    # NOTE: I did not migrate the user_required part, because it does not appear anywhere in the code
```
ryneeverett: :+1: I have no idea what that was about, I'm pretty sure it was just copied from the AnyUrl method I was masking.
```suggestion
```

bugwarrior/config/schema.py
```diff
@@ -16,133 +28,149 @@
...
+
+def parse_config_list(value: str | list[str]) -> list[str]:
+    """Cast ini string to a list of strings."""
+    if isinstance(value, str):
+        return [
+            item.strip()
+            for item in re.split(r",(?![^{]*})", value.strip())
+            if item != ""
+        ]
+    return list(value)
```
ryneeverett: It's already a list. Surely type checkers understand that?

```suggestion
    return value
```

bugwarrior/config/schema.py
```diff
@@ -16,133 +28,149 @@
...
+
+
+ConfigList = Annotated[list[str], BeforeValidator(parse_config_list)]
+
+
+def _expand_env_vars(value: str | Path) -> str:
+    """Expand $VAR and ${VAR} style environment variables."""
+    return Template(str(value)).safe_substitute(os.environ)
```
ryneeverett: Is there a reason this is preferred over `os.path.expandvars`?

bugwarrior/config/schema.py
```diff
@@ -16,133 +28,149 @@
...
+        except ValueError:
+            return path.resolve()
 
 
-# HACK https://stackoverflow.com/a/34116756
-class ExpandedPath(type(pathlib.Path())):  # type: ignore
-    @classmethod
-    def __get_validators__(cls):
-        yield cls.validate
+LoggingPath = Annotated[ExpandedPath, AfterValidator(_relativize_path)]
```
ryneeverett: Do you understand why we're making the path relative here? This seems like a lot of code to maintain a behavior which may not have made any sense in the first place.

bugwarrior/config/schema.py
```diff
@@ -151,17 +179,14 @@ class Config(PydanticConfig):
     #: Interactive status.
     interactive: bool
 
-    # added during validation (computed field support will land in pydantic-2)
-    #: Local data storage.
-    data: typing.Optional[BugwarriorData] = None
+    # optional
+    taskrc: TaskrcPath = Field(default_factory=get_default_taskrc)
```
ryneeverett: Please move this back below the `data` property -- the `# optional comment` applies to that whole next block of attributes.

bugwarrior/config/schema.py
```diff
@@ -172,104 +197,80 @@ def compute_data(cls, values):
...
     task_crud_sticky: bool = True
     only_on_new_tasks: bool = False
 
 
-class SchemaBase(pydantic.v1.BaseSettings):
-    class Config(PydanticConfig):
-        # Allow extra top-level sections so all targets don't have to be selected.
-        extra = 'ignore'
+class SchemaBase(BaseConfig):
+    model_config = ConfigDict(extra="ignore")
```
ryneeverett, line 189: Add back comment so we remember why we're doing this.

```suggestion
    # Allow extra top-level sections so all targets don't have to be selected.
    model_config = ConfigDict(extra="ignore")
```

bugwarrior/config/schema.py
```diff
@@ -172,104 +197,80 @@ def compute_data(cls, values):
...
-        super().__init__(self.flatten(error))
+    def __init__(self, error: pydantic.ValidationError):
+        super().__init__(self._format_error(e) for e in error.errors())
+
+    @staticmethod
+    def _format_error(error: ErrorDetails) -> str:
+        msg = error["msg"]
+        if error["type"] == "extra_forbidden":
+            msg = "unrecognized option"
+        return f"{ValidationErrorEnhancedMessages.display_error_loc(error['loc'])}  <- {msg}\n"
```
ryneeverett:
```suggestion
        return f"{self.display_error_loc(error['loc'])}  <- {msg}\n"
```

bugwarrior/config/schema.py
```diff
@@ -172,104 +197,80 @@ def compute_data(cls, values):
...
-        loc_len = len(e['loc'])
-        if loc_len == 1 or e['loc'][1] == '__root__':
-            return f"[{e['loc'][0]}]"
+    def display_error_loc(loc: tuple) -> str:
+        loc_len = len(loc)
+        if loc_len == 1 or (loc_len > 1 and loc[1] == "__root__"):
+            return f"[{loc[0]}]"
         elif loc_len == 2:
-            # TODO We should be able to display the value itself:
-            # https://github.com/samuelcolvin/pydantic/issues/784
```
ryneeverett, line 224: Per this link, it looks like we should be able to display the value with pydantic v2. So instead of an error message like

```
import_labels_as_tags  <- Input should be a valid boolean, unable to interpret input
```

it could show the whole line:

```
import_labels_as_tags = 100  <- Input should be a valid boolean, unable to interpret input
```

bugwarrior/config/schema.py
```diff
@@ -291,27 +292,35 @@ def validate_config(config: dict, main_section: str, config_path: str) -> dict:
...
-        'bugwarriorrc',
+    bugwarrior_config_model = pydantic.create_model(
+        "bugwarriorrc",
         __base__=SchemaBase,
-        __validators__={'compute_target': get_target_validator(targets)},
         general=(MainSectionConfig, ...),
-        flavor={
-            flavor: (MainSectionConfig, ...)
-            for flavor in config.get('flavor', {}).values()
+        # NOTE: I add to change this, but not sure about my change though
```
ryneeverett: Why did you have to change this? Seems like this change will cause a KeyError on `config['flavor']` in `bugwarrior/config/load.py` but evidently we don't have any automated tests for flavors.

bugwarrior/config/schema.py
```diff
@@ -321,41 +330,41 @@ def validate_config(config: dict, main_section: str, config_path: str) -> dict:
...
     },
 )
 
 
-class ServiceConfig(_ServiceConfig):  # type: ignore  # (dynamic base class)
-    """Pydantic_ base class for service configurations.
-
-    .. _Pydantic: https://docs.pydantic.dev/latest/
-    """
+# NOTE: I removed the deprecated fields
```
ryneeverett: These need to be added back. Like other deprecation warnings, they may have been in the code base for a long time but until a couple weeks ago there hadn't been a release for 5 years.

bugwarrior/config/schema.py
```diff
@@ -321,41 +330,41 @@ def validate_config(config: dict, main_section: str, config_path: str) -> dict:
...
+class ServiceConfig(_ServiceConfig):
+    """Pydantic base class for service configurations."""
 
-    Config = PydanticConfig
+    model_config = ConfigDict(extra="forbid")
```
ryneeverett: Doesn't `extra = 'forbid'` get inherited from `BaseConfig` via `_ServiceConfig`?

bugwarrior/config/schema.py
```diff
@@ -321,41 +330,41 @@ def validate_config(config: dict, main_section: str, config_path: str) -> dict:
...
 
-    # Added during validation (computed field support will land in pydantic-2)
+    # Added during validation (computed field)
     templates: dict = {}
     target: typing.Optional[str] = None
 
-    # Optional fields shared by all services.
-    only_if_assigned: str = ''
+    # typing.Optional fields shared by all services.
```
ryneeverett:
```suggestion
    # Optional fields shared by all services.
```

bugwarrior/config/schema.py
```diff
@@ -383,31 +392,15 @@ def compute_templates(cls, values):
...
-            if values['project_name'] != '':
-                log.warning('project_name is deprecated in favor of project_template')
+        # NOTE: 1. Not sure I understand why we try both "templates", and f"key_template".
+        # Couldn't we simply use "templates" as dict ? It would make the pydantic easiser to define
+        # NOTE: 2. Since taskwarrior fields aren't updated often, shouldn't we define this list statically
+        # NOTE: 3. this field is broken for ArrayFields such as annotations, right?
+        if isinstance(values, dict):
+            templates = values.get("templates", {})
+            for key in taskw.task.Task.FIELDS.keys():
+                template = values.pop(f"{key}_template", None)
```
ryneeverett: Any reason this needed to change?

```suggestion
                template = values.get(f'{key}_template')
```

bugwarrior/config/schema.py
```diff
@@ -383,31 +392,15 @@ def compute_templates(cls, values):
...
-    def deprecate_project_name(cls, values):
-        if hasattr(cls, '_DEPRECATE_PROJECT_NAME'):
-            if values['project_name'] != '':
-                log.warning('project_name is deprecated in favor of project_template')
+        # NOTE: 1. Not sure I understand why we try both "templates", and f"key_template".
+        # Couldn't we simply use "templates" as dict ? It would make the pydantic easiser to define
+        # NOTE: 2. Since taskwarrior fields aren't updated often, shouldn't we define this list statically
+        # NOTE: 3. this field is broken for ArrayFields such as annotations, right?
+        if isinstance(values, dict):
+            templates = values.get("templates", {})
```
ryneeverett:
```suggestion
            templates = {}
```

bugwarrior/config/schema.py
```diff
@@ -383,31 +392,15 @@ def compute_templates(cls, values):
...
-
-    @pydantic.v1.root_validator
-    def deprecate_project_name(cls, values):
-        if hasattr(cls, '_DEPRECATE_PROJECT_NAME'):
-            if values['project_name'] != '':
-                log.warning('project_name is deprecated in favor of project_template')
+        # NOTE: 1. Not sure I understand why we try both "templates", and f"key_template".
+        # Couldn't we simply use "templates" as dict ? It would make the pydantic easiser to define
+        # NOTE: 2. Since taskwarrior fields aren't updated often, shouldn't we define this list statically
+        # NOTE: 3. this field is broken for ArrayFields such as annotations, right?
```
ryneeverett: 1. There is no `templates` dict and we weren't trying it before. Maybe if we were writing bugwarrior from scratch today we'd do a `templates` dict rather than `{key}_template` fields, but dicts are a nasty hack in INI files.
2. No, the main use for this is not the core taskwarrior fields but user defined attributes (UDA's). While I primarily see people using this feature for the UDA's bugwarrior adds, one could also add their own personal UDA and compute it with templates.
3. I believe that is correct. I believe you'd get an error when it passed a non-string into Jinja.

bugwarrior/services/azuredevops.py
```diff
@@ -29,9 +22,8 @@ class AzureDevopsConfig(config.ServiceConfig):
     project: EscapedStr
     organization: EscapedStr
 
-    host: config.NoSchemeUrl = config.NoSchemeUrl(
-        'dev.azure.com', scheme='https', host='azure.com'
-    )
+    # NOTE: I don't understand why we had a scheme passed in NoSchemUrl here
```
ryneeverett: Me either!

```suggestion
```

bugwarrior/services/bitbucket.py
```diff
@@ -24,22 +22,14 @@ class BitbucketConfig(config.ServiceConfig):
...
-    @pydantic.v1.root_validator
-    def deprecate_password_authentication(cls, values):
-        if values['login'] != 'Undefined' or values['password'] != 'Undefined':
-            log.warning(
-                'Bitbucket has disabled password authentication and, as such, '
-                'the "login" and "password" options are deprecated and should '
-                'be removed from your configuration file.'
-            )
-        return values
+    # NOTE: I removed the deprecation validation, as it's quite old
```
ryneeverett: Please add back all deprecations, per comment elsewhere.

bugwarrior/services/bz.py
```diff
@@ -15,27 +14,11 @@
...
-                'error.'
-            )
-        return super().validate(value.rstrip('/'), field, config)
-
-
 class BugzillaConfig(config.ServiceConfig):
     service: typing.Literal['bugzilla']
     username: str
-    base_uri: OptionalSchemeUrl
+    # NOTE: removed OptionalSchemeURl based on deprecation period
```
ryneeverett: Please add back all deprecations, per comment elsewhere.

bugwarrior/services/github.py
```diff
@@ -27,45 +27,39 @@ class GithubConfig(config.ServiceConfig):
...
     import_labels_as_tags: bool = False
     label_template: str = '{{label}}'
     filter_pull_requests: bool = False
     exclude_pull_requests: bool = False
     include_user_issues: bool = True
     involved_issues: bool = False
-    host: config.NoSchemeUrl = config.NoSchemeUrl(
-        'github.com', scheme='https', host='github.com'
-    )
+    # NOTE: why did we use a scheme with NoSchemeUrl ?
```
ryneeverett: Probably by mistake. I wish pydantic would validate the default values.

```suggestion
```

bugwarrior/services/github.py
```diff
@@ -27,45 +27,39 @@ class GithubConfig(config.ServiceConfig):
...
-    ignore_user_comments: config.ConfigList = config.ConfigList([])
-
-    @pydantic.v1.root_validator
-    def deprecate_password(cls, values):
-        if values['password'] != 'Deprecated':
-            log.warning(
-                'Basic auth is no longer supported. Please remove '
-                '"password" in favor of "token".'
-            )
-        return values
```
ryneeverett, line 53: Please add back all deprecations, per comment elsewhere.

bugwarrior/services/github.py
```diff
@@ -27,45 +27,39 @@ class GithubConfig(config.ServiceConfig):
...
 
-    @pydantic.v1.root_validator
+    @model_validator(mode='before')
+    @classmethod
     def issue_urls_consistent_with_host(cls, values):
         issue_url_paths = []
-        for url in values['issue_urls']:
+        issue_urls = values.get('issue_urls', [])
+        if isinstance(issue_urls, str):
+            issue_urls = [url.strip() for url in issue_urls.split(',') if url.strip()]
```
ryneeverett: Why are we supposing issue_urls is optional and/or a string?

bugwarrior/services/github.py
```diff
@@ -27,45 +27,39 @@ class GithubConfig(config.ServiceConfig):
...
+        for url in issue_urls:
             parsed_url = urllib.parse.urlparse(url)
-            if parsed_url.netloc != values['host']:
+            if parsed_url.netloc != values.get('host', 'github.com'):
                 raise ValueError(
-                    f'issue_urls: {url} inconsistent with host {values["host"]}'
+                    f'issue_urls: {url} inconsistent with host {values.get("host")}'
```
ryneeverett:
```suggestion
                    f'issue_urls: {url} inconsistent with host {values.get("host", "github.com")}'
```

bugwarrior/services/jira.py
```diff
@@ -72,9 +68,14 @@ def extract_value(self, fields):
...
 
+JiraExtraFields = typing.Annotated[
+    list[JiraExtraField], BeforeValidator(parse_jira_extra_fields)
+]
+
+
 class JiraConfig(config.ServiceConfig):
     service: typing.Literal['jira']
-    base_uri: pydantic.v1.AnyUrl
+    base_uri: config.StrippedTrailingSlashUrl
```
ryneeverett, line 78: Why can't this stay an `AnyUrl`?

ryneeverett (changes requested):
Thanks for figuring out the multiprocessing issue! It is somewhat unfortunate that we're no longer exercising the multiprocessing in our tests but the additional code we'd need to inject mocks into child processes probably isn't worth it to test the very small distinction in the multiprocessing logic branch.

Looking at the api-change-check, I don't think we want `model_config` added to the docs. I think you should be able to remove it by adding it to the relevant `exclude-members` in `bugwarrior/docs/other-services/api.rst`.

ryneeverett:
I enabled pull request previews in readthedocs and it would have been helpful here but they're not working. Per [the readthedocs docs](https://docs.readthedocs.com/platform/stable/guides/pull-requests.html#limitations) my only thought is that the readthedocs webhook is not configured to send pull request events. @djmitche Could you check and enable that if it's not?

bugwarrior/config/ini2toml_plugin.py
Lotram: The `schema()` method is deprecated, and now actually calls `model_json_schema`:
```
    @classmethod
    @typing_extensions.deprecated('The `schema` method is deprecated; use `model_json_schema` instead.', category=None)
    def schema(  # noqa: D102
        cls, by_alias: bool = True, ref_template: str = DEFAULT_REF_TEMPLATE
    ) -> Dict[str, Any]:  # noqa UP006
        warnings.warn(
            'The `schema` method is deprecated; use `model_json_schema` instead.',
            category=PydanticDeprecatedSince20,
            stacklevel=2,
        )
        return cls.model_json_schema(by_alias=by_alias, ref_template=ref_template)
```

bugwarrior/config/schema.py
Lotram: `pathlib` is usually considered the more modern, pythonic way to manipulate PathLike objects, and often clearer than os.path. 
In this specific case, it's true it does not bring much clarity, so I can revert it if you want

bugwarrior/config/schema.py
Lotram: Nope :sweat_smile:  I tried to migrate the code without changing the behaviour, as my PR is already quite complex, but I completely agree, I don't really see the point. Maybe so the config can be copied as-is to another machine ?

bugwarrior/config/schema.py
Lotram: Note that I need to change those method to `classmethod` instead of `staticmethod`. 

While re-reading this code, I think the whole `ValidationErrorEnhancedMessages` class could be a simple function, don't you think ?

bugwarrior/config/schema.py
Lotram: You're right about the first comment, I don't know why I thought it could be an input.

bugwarrior/services/jira.py
Lotram: Because it's not a str anymore, and we use it as a string in some places (e.g in `get_url`), so this was broken.

bugwarrior/services/github.py
Lotram: Because I was using a model_validator in 'before' mode, so `values` were basically the input dict. I changed it to use a `field_validator` in 'after' mode

bugwarrior/config/schema.py
Lotram: Ok so I now know why I thought that: we're using a 'templates' key in TestTemplates.get_issue. I fixed this to use the expected keys instead

bugwarrior/config/schema.py
Lotram: I refactored this, I think it's much simpler now

bugwarrior/config/schema.py
Lotram: I reverted this, I actually just needed to also pass the field type

Lotram:
@ryneeverett I took all your remarks into account.

bugwarrior/config/schema.py
ryneeverett: I'd prefer reverting. I generally agree that pathlib is more expressive but in this case it is not.

bugwarrior/config/schema.py
ryneeverett: >  I tried to migrate the code without changing the behaviour

This is the same line of thinking I had a few years ago when I migrated from using ConfigParser directly to pydantic and is the reason we have all the different kinds of paths and urls. In my opinion this particular behavior has crossed the line of unreasonableness to maintain. I'd vote to remove the relativization.

I found that I'm actually guilty of merging this behavior in #614 without noticing or questioning it. The original lines were:

```py
def fix_logging_path(config, main_section):
    """
    Expand environment variables and user home (~) in the log.file and return
    as relative path.
    """
    log_file = config.get(main_section, 'log.file')
    if log_file:
        log_file = os.path.expanduser(os.path.expandvars(log_file))
        if os.path.isabs(log_file):
            log_file = os.path.relpath(log_file)
    return log_file
```

I think it is more likely that relativization *seemed* to be necessary at the time than that it was added to support some obscure use case.

bugwarrior/config/schema.py
Lotram: > No, the main use for this is not the core taskwarrior fields but user defined attributes (UDA's). While I primarily see people using this feature for the UDA's bugwarrior adds, one could also add their own personal UDA and compute it with templates.

I'm probably missing something, but we're iterating over `taskw.task.Task.FIELDS` to build the template keys. This `FIELDS` attribute does not have UDAs, so they can't be templated, right ?

ryneeverett: Oh that surprises me but you're right!

> Since taskwarrior fields aren't updated often, shouldn't we define this list statically

I assume your thought here is that we could fix the broken ArrayFields by excluding them from our static list. I think a better solution would be twofold:

1. Support UDA's by iterating over both the core fields and the actual fields of the record. (I don't think the record is guaranteed to have all the core fields so that alone wouldn't be enough.)
2. Catch ArrayFields in the pydantic validator for templates.

Regardless, I'd like this to be separate PRs or issues.

bugwarrior/config/schema.py
Lotram: Done

bugwarrior/config/schema.py
Lotram: Removed

ryneeverett:
I suspect you overlooked this comment:

> Looking at the api-change-check, I don't think we want `model_config` added to the docs. I think you should be able to remove it by adding it to the relevant `exclude-members` in `bugwarrior/docs/other-services/api.rst`.

bugwarrior/config/ini2toml_plugin.py
ryneeverett: So if we wanted to we could pass those exact same arguments (`model_json_schema(by_alias=True, ref_template=pydantic.json_schema.DEFAULT_REF_TEMPLATE)`) and eliminate `get_field_type`. I don't have a strong opinion about it though. Since the ini2toml test passes I think we can be fairly confident the behavior is the same.

ryneeverett: If you do decide to leave this as is, please remove the NOTE comment.

Lotram: > So if we wanted to we could pass those exact same arguments (model_json_schema(by_alias=True, ref_template=pydantic.json_schema.DEFAULT_REF_TEMPLATE)) and eliminate get_field_type.

Not sure I understand this comment. I tried to look at the parameters, but I could not get rid of `get_field_type` anyway. The pydantic function just changed with v2, and now creates proper json schemas, so we have to parse `AnyOf`.

I removed the NOTE

bugwarrior/services/bz.py
```diff
@@ -2,40 +2,40 @@
...
-        return super().validate(value.rstrip('/'), field, config)
+
+OptionalSchemeUrl = Annotated[StrippedTrailingSlashUrl, BeforeValidator(validate_url)]
 
 
 class BugzillaConfig(config.ServiceConfig):
     service: typing.Literal['bugzilla']
     username: str
-    base_uri: OptionalSchemeUrl
+    base_uri: config.StrippedTrailingSlashUrl
```
ryneeverett:
```suggestion
    base_uri: OptionalSchemeUrl
```

bugwarrior/config/schema.py
```diff
@@ -172,98 +163,72 @@ def compute_data(cls, values):
...
-    notifications: Notifications = Notifications()
-
+class SchemaBase(BaseConfig):
+    # Allow extra top-level sections so all targets don't have to be selected.
+    model_config = ConfigDict(extra="ignore")
 
-class ValidationErrorEnhancedMessages(list):
-    """Methods loosely adapted from pydantic.error_wrappers."""
+    hooks: Hooks = Field(default_factory=Hooks)
+    notifications: Notifications = Field(default_factory=Notifications)
```
ryneeverett: Per [the docs](https://docs.pydantic.dev/latest/concepts/models/#nested-models) it seems like this would work. I'm assuming there's a good reason it doesn't?

```suggestion
    hooks: Hooks
    notifications: Notifications
```

bugwarrior/config/schema.py
```diff
@@ -172,98 +163,72 @@ def compute_data(cls, values):
...
         elif loc_len == 2:
-            # TODO We should be able to display the value itself:
-            # https://github.com/samuelcolvin/pydantic/issues/784
-            return f"[{e['loc'][0]}]\n{e['loc'][1]}"
-        raise ValueError(
-            'Configuration should not be nested more than two layers deep.'
-        )
+            formatted_error_loc = f"[{loc[0]}]\n{loc[1]}"
+            if _error["type"] not in {"missing", "extra_forbidden"}:
+                formatted_error_loc = f"{formatted_error_loc} = '{_error['input']}'"
```
ryneeverett: IMO it should look as similar as possible to the configuration file, which will never have single quotes. This looks particularly bad with arrays and tables which end up having nested single quotes.

```suggestion
                formatted_error_loc = f"{formatted_error_loc} = {_error['input']}"
```

bugwarrior/config/schema.py
```diff
@@ -172,98 +163,72 @@ def compute_data(cls, values):
...
+            formatted_error_loc = f"[{loc[0]}]"
         elif loc_len == 2:
-            # TODO We should be able to display the value itself:
-            # https://github.com/samuelcolvin/pydantic/issues/784
-            return f"[{e['loc'][0]}]\n{e['loc'][1]}"
-        raise ValueError(
-            'Configuration should not be nested more than two layers deep.'
-        )
+            formatted_error_loc = f"[{loc[0]}]\n{loc[1]}"
+            if _error["type"] not in {"missing", "extra_forbidden"}:
```
ryneeverett: Per the principle of looking as similar as possible to the configuration file:

```suggestion
            if _error["type"] != "missing":
```

bugwarrior/config/schema.py
```diff
@@ -321,41 +289,38 @@ def validate_config(config: dict, main_section: str, config_path: str) -> dict:
...
     },
 )
 
 
-class ServiceConfig(_ServiceConfig):  # type: ignore  # (dynamic base class)
-    """Pydantic_ base class for service configurations.
+class ServiceConfig(_ServiceConfig):
+    """Pydantic base class for service configurations."""
 
-    .. _Pydantic: https://docs.pydantic.dev/latest/
```
ryneeverett, line 342: Please restore this comment so we get a hyperlink to pydantic in the docs: https://bugwarrior.readthedocs.io/en/stable/other-services/api.html#bugwarrior.config.ServiceConfig

bugwarrior/config/schema.py
```diff
@@ -383,31 +348,39 @@ def compute_templates(cls, values):
         generated issue was.
 
         """
-        for key in taskw.task.Task.FIELDS.keys():
-            template = values.get(f'{key}_template')
-            if template is not None:
-                values['templates'][key] = template
+        if isinstance(values, dict):
```
ryneeverett: Is it actually possible for `values` to not be a dict? I see we have no such guard in the `compute_target` model_validator, is this case different?

tests/config/test_schema.py
```diff
@@ -4,50 +4,27 @@
...
+import pydantic
+from pydantic import TypeAdapter
 
 from bugwarrior.config import schema
 
 from ..base import ConfigTest
 
 
-class TestLoggingPath(unittest.TestCase):
+class TestConfigList(unittest.TestCase):
```
ryneeverett, line 37: Seems like we were indirectly testing ExpandedPath here.

```suggestion
class TestExpandedPath(unittest.TestCase):
    def setUp(self):
        self.adapter = TypeAdapter(schema.ExpandedPath)
        self.dir = os.getcwd()
        os.chdir(os.path.expanduser('~'))
        self.log = pathlib.Path('./bugwarrior.log').absolute()

    def test_log(self):
        filename = os.path.join(os.path.expandvars('$HOME'), self.log)
        self.assertEqual(self.adapter.validate_python(filename), self.log)

    def test_log_userhome(self):
        self.assertEqual(self.adapter.validate_python('~/bugwarrior.log'), self.log)

    def test_log_envvar(self):
        self.assertEqual(self.adapter.validate_python('$HOME/bugwarrior.log'), self.log)

    def tearDown(self):
        os.chdir(self.dir)


class TestConfigList(unittest.TestCase):
```

ryneeverett (changes requested):

bugwarrior/config/schema.py
Lotram: In the example you shared, there is no default value because `foo` and `bars` are passed as parameters: 
```
m = Spam(foo={'count': 4}, bars=[{'apple': 'x1'}, {'apple': 'x2'}])
```
That's not case here.

Though we should do `bugwarrior_config_model.model_validate(config)` insted of `bugwarrior_config_model(**config)`, which is a bit cleaner IMO

We can also do:
```
config.setdefault("hooks", {})
config.setdefault("notifications", {})
```
just before, to avoid using a field_factory, but I'm not sure that's better, what do you think ? 

In any case, we can do:
```
hooks: Hooks = Hooks()
```
simpler than using a Field + default factory

bugwarrior/config/schema.py
Lotram: Nope, you're right

bugwarrior/config/schema.py
ryneeverett: You've now reverted back to the way I had it before which still seem like the simplest of the options to me. Thanks.

ryneeverett (approved):
I'll probably give the whole thing another look-over in the next day or two and it's possible that I'll have a few more comments. Thanks so much for taking the lead on this!

bugwarrior/services/linear.py
```diff
@@ -17,19 +17,22 @@ class LinearConfig(config.ServiceConfig):
     api_token: str
 
     host: config.StrippedTrailingSlashUrl = config.StrippedTrailingSlashUrl(
-        "https://api.linear.app/graphql", scheme="https", host="api.linear.app"
+        "https://api.linear.app/graphql"
```
ryneeverett: I don't think the string needs to be passed into StrippedTrailingSlashUrl any more.

bugwarrior/services/linear.py
```diff
@@ -17,19 +17,22 @@ class LinearConfig(config.ServiceConfig):
...
     import_labels_as_tags: bool = False
     label_template: str = "{{label|replace(' ', '_')}}"
     also_unassigned: config.UnsupportedOption[bool] = False
 
-    @pydantic.v1.root_validator
+    @model_validator(mode='before')
+    @classmethod
     def statuses_or_status_types(cls, values):
-        if values["statuses"] and values["status_types"]:
+        statuses = values.get("statuses", [])
```
ryneeverett: Doesn't seem like there's any reason for this one to get a different default than the next one. They're both only used in boolean expressions.

```suggestion
        statuses = values.get("statuses")
```

PR #1164, opened 2025-12-19
chore: remove pytz dependency

Lotram:
Some context about this MR:

I stumbled the `Issue.parse_date`, and noticed the `timezone` behaviour is really ambiguous:
```
    def parse_date(self, date, timezone='UTC'):


        """
        if date:
            date = parse_date(date)
            if not date.tzinfo:
                if timezone == '':
                    tzinfo = tzlocal()
                else:
                    tzinfo = pytz.timezone(timezone)
                date = date.replace(tzinfo=tzinfo)
            return date.replace(microsecond=0)
        return None
```

from looking at the function body, we could think the default timezone is `tzlocal()`, but the default value is actually `UTC`. Someone would need to pass `timezone=''` to enter this if block, which seems unlikely. 

I checked, and the `timezone` parameter is actually never used. 
There are only two places (based on the tests) where the `date` string does not already have a timezone, it's:
* Gitlab `duedate`, which is actually a date, not a datetime, so we don't really care about the timezone
* Gmail's get_entry. As the InternalDate is actually a timestamp, I switch to use `datetime.fromtimestamp` (already used in multiple places)

So I ended up removing the `timezone` parameter of `Issue.parse_date` (which should be called `parse_datetime` IMO)

I then noticed we were using `pytz` in several places, just to get the utc timezone, so I replaced them with `timezone.utc`, so pytz can be removed as a dependency.

If we actually want to keep the `timezone` parameter in `Issue.parse_date`, I suggest using `zoneinfo.ZoneInfo` which in the standard library, instead of relying on the third party lib.

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -4,12 +4,13 @@ Service API
 """
 
 import abc
-from datetime import datetime, timezone
+import datetime
 import logging
 import math
 import os
 import re
 import typing
+import zoneinfo
 
 from dateutil.parser import parse as parse_date
 import dogpile.cache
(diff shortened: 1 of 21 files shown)


ryneeverett:
FYI I made a pull request against this branch: https://github.com/Lotram/bugwarrior/pull/1.

ryneeverett (approved):

PR #1165, opened 2025-12-20
pyproject: Enforce 100 character max line length

ryneeverett:
We had this before with flake8 and I didn't realize that ruff wasn't enforcing this.


ryneeverett:
@Lotram Any comment?

Lotram:
> @Lotram Any comment?

Nope, seems like a good idea, especially if the rule was there with flake8

Issue #1166, opened 2025-12-20
Flavor feature seems broken


Lotram:
I've noticed there is no test about flavors, so I wanted to add a simple one:
```
    def test_flavors(self):
        self.config["general"]["interactive"] = False
        self.config["flavor"] = {
            "work": {"targets": ["my_service", "my_gitlab"], "interactive": False},
            "personal": {"targets": ["my_kan"], "interactive": False},
        }

        conf = self.validate()
```
This breaks in `validate_config`. I checked, and this was already the case before I migrated the code to pydantic v2. 
Maybe I'm wrong with my test, and the flavors are not passed like that ? I think it's the case, but I'm not sure.

We should add some tests about flavors anyway

Lotram:
Note that it looks like the fix is simply to remove the `.values()` here:
```
flavor=(
            dict[str, MainSectionConfig],
            {
                flavor: (MainSectionConfig, ...)
                for flavor in config.get('flavor', {}).values()
            },
        )
```

Lotram:
Testing this feature with the packaged 2.0.0 version, I can confirm it does not work

Lotram:
The problem is actually  deeper than I first thought. It might be a good time to refactor this a bit ? 
In the code, there is no reason to treat `general` and other flavors differently, IMO the Config pydantic model does not need to know about flavors, as we know from the start which flavor to use. It can become main configuration right there.

ryneeverett:
I'm also noticing that ini2toml quotes flavor sections which I don't believe will work because we rely on them being parsed into a nested 'flavor' dict.

```py
>>> import tomllib
>>> tomllib.loads('[flavor.myflavor]')
{'flavor': {'myflavor': {}}}
>>> tomllib.loads('["flavor.myflavor"]')
{'flavor.myflavor': {}}
```

PR #1167, opened 2025-12-20
Simplify flavors

Lotram:
related to #1166 

I noticed the flavors were broken. 

After trying several things, I realized we could simply define the "general" section as the flavor being used, instead of passing the `main_section` param everywhere.

I think it's a huge simplification, the only (small) drawback is we currently don't validate other flavors (i.e if I'm using "work",  I won't validate either "general" nor "personal" are valid). It's not a big loss IMO, but if we really want to validate them all, we could either add a "validate_config" command, or simply iterate over flavors, build a config with each one, and call `validate_config`. 

I think the `flavor_name` field  (used to log errors) I added to the pydantic model will appear in the docs, which we probably don't want, how to change that ?

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -19,15 +19,9 @@ log = logging.getLogger(__name__)
 lst = list
 
 
-def _get_section_name(flavor):
-    if flavor:
-        return 'flavor.' + flavor
-    return 'general'
-
-
-def _try_load_config(main_section, interactive=False, quiet=False):
+def _try_load_config(flavor, interactive=False, quiet=False):
     try:
-        return load_config(main_section, interactive, quiet)
+        return load_config(flavor, interactive, quiet)
     except OSError:
         # Our standard logging configuration depends on the bugwarrior
         # configuration file which just failed to load.
@@ -82,7 +76,7 @@ def cli():
 
 @cli.command()
 @click.option('--dry-run', is_flag=True)
-@click.option('--flavor', default=None, help='The flavor to use')
+@click.option('--flavor', default="general", help='The flavor to use')
 @click.option('--interactive', is_flag=True)
 @click.option(
     '--debug', is_flag=True, help='Do not use multiprocessing (which breaks pdb).'
@@ -96,20 +90,17 @@ def pull(dry_run, flavor, interactive, debug, quiet):
     """
 
     try:
-        main_section = _get_section_name(flavor)
-        config = _try_load_config(main_section, interactive, quiet)
+        config = _try_load_config(flavor, interactive, quiet)
 
-        lockfile_path = os.path.join(
-            config[main_section].data.path, 'bugwarrior.lockfile'
-        )
+        lockfile_path = os.path.join(config["general"].data.path, 'bugwarrior.lockfile')
         lockfile = PIDLockFile(lockfile_path)
         lockfile.acquire(timeout=10)
         try:
             # Get all the issues.  This can take a while.
-            issue_generator = aggregate_issues(config, main_section, debug)
+            issue_generator = aggregate_issues(config, debug)
 
             # Stuff them in the taskwarrior db as necessary
-            synchronize(issue_generator, config, main_section, dry_run)
+            synchronize(issue_generator, config)
         finally:
             lockfile.release()
     except LockTimeout:
@@ -187,7 +178,7 @@ def set(target, username):
 
 
 @cli.command()
-@click.option('--flavor', default=None, help='The flavor to use')
+@click.option('--flavor', default="general", help='The flavor to use')
 @_legacy_cli_deprecation_warning
 def uda(flavor):
     """
@@ -211,10 +202,9 @@ def uda(flavor):
        may prevent you from changing the values of those fields or using them
        in filter expressions.
     """
-    main_section = _get_section_name(flavor)
-    conf = _try_load_config(main_section)
+    conf = _try_load_config(flavor)
     print("# Bugwarrior UDAs")
-    for uda in get_defined_udas_as_strings(conf, main_section):
+    for uda in get_defined_udas_as_strings(conf):
         print(uda)
     print("# END Bugwarrior UDAs")
 
--- a/tests/config/test_load.py
+++ b/tests/config/test_load.py
@@ -236,3 +236,39 @@ class TestParseFile(LoadTest):
 
         with self.assertRaises(SystemExit):
             load.parse_file(config_path)
+
+    def test_toml_flavors(self):
+        config_path = self.create('.bugwarrior.toml')
+        with open(config_path, 'w') as fout:
+            fout.write(
+                textwrap.dedent("""
+                [general]
+                targets = ["my_github"]
+
+                [flavor.work]
+                targets = ["my_jira"]
+
+                [flavor.personal]
+                targets = ["my_github"]
+                shorten = true
+
+                [my_github]
+                service = "github"
+                login = "test"
+                username = "test"
+                token = "abc"
+
+                [my_jira]
+                service = "jira"
+                base_uri = "https://jira.example.org"
+                username = "test"
+                password = "abc"
+            """)
+            )
+        config = load.parse_file(config_path)
+        self.assertIn('flavor', config)
+        self.assertIn('work', config['flavor'])
+        self.assertIn('personal', config['flavor'])
+        self.assertEqual(config['flavor']['work']['targets'], ['my_jira'])
+        self.assertEqual(config['flavor']['personal']['targets'], ['my_github'])
+        self.assertTrue(config['flavor']['personal']['shorten'])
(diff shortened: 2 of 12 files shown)


tests/config/test_load.py
```diff
@@ -236,3 +236,39 @@ def test_ini_wrong_prefix(self):
...
+                password = "abc"
+            """)
+            )
+        config = load.parse_file(config_path)
+        self.assertIn('flavor', config)
+        self.assertIn('work', config['flavor'])
+        self.assertIn('personal', config['flavor'])
+        self.assertEqual(config['flavor']['work']['targets'], ['my_jira'])
+        self.assertEqual(config['flavor']['personal']['targets'], ['my_github'])
+        self.assertTrue(config['flavor']['personal']['shorten'])
```
ryneeverett, line 274: Unless I'm mistaken, this seems to only really be testing tomllib. Would this test fail on `develop`?

bugwarrior/command.py
```diff
@@ -96,20 +90,17 @@ def pull(dry_run, flavor, interactive, debug, quiet):
...
         lockfile = PIDLockFile(lockfile_path)
         lockfile.acquire(timeout=10)
         try:
             # Get all the issues.  This can take a while.
-            issue_generator = aggregate_issues(config, main_section, debug)
+            issue_generator = aggregate_issues(config, debug)
 
             # Stuff them in the taskwarrior db as necessary
-            synchronize(issue_generator, config, main_section, dry_run)
+            synchronize(issue_generator, config)
```
ryneeverett, line 103: You left out dry_run! :scream: It would probably be safer if that parameter was required.

ryneeverett (commented):
Frankly, I'm not sure I see this as an overall simplification. The only thing that strikes me as simpler is that the schema no longer covers inactive flavors, which as you noted means they're not validated.

I don't find that passing the main_section variable around is much of a problem and seems a bit cleaner to me than hardcoding the "general" string everywhere and copying the flavor to the "general" section. IMO we should also minimize modification of rawconfig and adding a flavor_name field just for one error message is a bit smelly.

> I think the flavor_name field (used to log errors) I added to the pydantic model will appear in the docs

Nope, only fields which have an [autodoc comment](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings) (starting with `#:`) are documented.

bugwarrior/command.py
Lotram: Good catch

Lotram:
To me, this is a huge gain, because:
1. It means we can further decouple the config parsing and the processing. If we ever decide to change the flavor config, it does not impact the rest at all. 
2. It also means we don't have to add any tests regarding flavors after the parsing phase, as there is nothing to test. 
3. The less parameters in functions, the less reasons they have to fail. It becomes less likely to introduce a breaking change, functions are more specialized.

I really think it's better. 

That said, I also agree that hardcoding "general" everywhere is not the cleanest  solution. I did this to avoid bigger changes. I think the cleanest solution would be to pass a `MainSectionConfig` object, better for typing, easier to understand for devs, easier for modern IDEs to analyze the code. In `validate_config`, instead of returning a dict, we could simply return the pydantic object we've just created

A the whole code is executed by a click command, we could also attach this config to click's context, and access it when needed, though I'm not convinced by this solution.

tests/config/test_load.py
Lotram: I added this test mostly because I wanted to be sure how the flavors were parsed. I can remove it

ryneeverett:
> 1. It means we can further decouple the config parsing and the processing. If we ever decide to change the flavor config, it does not impact the rest at all.

I'm sorry but I'm not following. Can you give an example of the kind of change this facilitates?

> I think the cleanest solution would be to pass a `MainSectionConfig` object

Do you mean passing a `MainSectionConfig` object rather than the `main_section` string? It seems redundant to pass the whole `config` alongside a pydantic object which is a view into a portion of that config, but that isn't a big deal if it improves something else.

> easier for modern IDEs to analyze the code

Does this happen to be your primary motivation? Perhaps if I had more experience with IDEs I'd be more sympathetic to changes like this.

> In `validate_config`, instead of returning a dict, we could simply return the pydantic object we've just created

My recollection is that the reason we return a dict is that it makes the testing simpler because we can just deal with dicts rather than pydantic objects. This is probably also not a big deal to change though.

Lotram:
> I'm sorry but I'm not following. Can you give an example of the kind of change this facilitates?

For now, all functions using `conf` and `main_section` parameters assume that the `conf`dictionnary a key that matches `main_section` and it has the right format, so callers need to make sure of that. 

By removing the `main_section` parameter, we now just have to ensure the conf is valid, no need to ensure consistency with the second parameter. At that point, the config does not need to know about what flavors are, and how we retrieve them from the config file. 
Maybe I'm not clear enough in my explanations, but IMO, all else being equal, removing a function parameter is always a good thing

> Do you mean passing a MainSectionConfig object rather than the main_section string? It seems redundant to pass the whole config alongside a pydantic object which is a view into a portion of that config, but that isn't a big deal if it improves something else.

When there is also a `conf` parameter, I would just pass the entire conf. When we need to use the "main_section" config, we would just need to do `conf.main` or `conf.general` (or whatever name we use)

I think we could have a simpler Config class, something like 
```
class Config(BaseConfig):
    # Allow extra top-level sections so all targets don't have to be selected.
    model_config = ConfigDict(extra="ignore")

    hooks: Hooks = Hooks()
    notifications: Notifications = Notifications()
    main: MainSectionConfig
    service_configs: list[ServiceConfig] 

```
so we don't have to rely on dynamic names (both for services and flavor). This once agains allows for better typing, and better python language server integration, which brings us to the last point:


> Does this happen to be your primary motivation? Perhaps if I had more experience with IDEs I'd be more sympathetic to changes like this.

Primary motivation, I don't know, but one of the reasons yes. If we have a proper Config object passed as an argument, it's much easier, as someone wanting to contribute, to understand what is expected, what fields can I use, etc...
It also means the Language Server (let's say for VSCode and IntelliJ, among the main IDEs), can directly show the dev the class definition, and add some linting / error detection. 
This is much harder when using simple dicts, because there is no guarantee on the keys (except if we want to use TypedDict, also a possibility, but not as good as pydantic)

But once again, even though I really think that using pydantic models instead of dicts would be a developer experience / QoL improvement, simply removing a parameter from a big part of the code is itself an improvement.

ryneeverett:
> I think we could have a simpler Config class, something like
> 
> ```
> class Config(BaseConfig):
>     # Allow extra top-level sections so all targets don't have to be selected.
>     model_config = ConfigDict(extra="ignore")
> 
>     hooks: Hooks = Hooks()
>     notifications: Notifications = Notifications()
>     main: MainSectionConfig
>     service_configs: list[ServiceConfig] 
> ```
> 
> so we don't have to rely on dynamic names (both for services and flavor).

I do see the appeal of this outcome but I'm not convinced that achieving it is worth the tradeoff.

I think the fundamental things we're butting heads on is prioritization of the two roles that pydantic is fulfilling in bugwarrior -- configuration validation and providing a "view" into parsed values. From my perspective, describing and enforcing a complete schematic of the configuration file is the primary role, while you are more concerned with the pydantic model from the consumer's perspective.

I'm not sure I see an opportunity to improve one without taking from the other. The fact is that our configuration schema currently *is* dynamic. (Had we been using pydantic when it was designed I believe we would have avoided this.) There doesn't seem to be a good way to implement a dynamic schema with a non-dynamic pydantic model.

Maybe we're asking too much from pydantic to provide the ideal configuration interface for both inputs and outputs. Maybe we should consider another step in which the validation model is transformed into a view (like your quoted example) that is better optimized for consumers of the model. I'm not entirely convinced that such an improvement of the view justifies adding another transformation step but if it is as simple as I suspect it would be I'd be inclined to defer to your judgement.

Lotram:
> The fact is that our configuration schema currently is dynamic. (Had we been using pydantic when it was designed I believe we would have avoided this.) There doesn't seem to be a good way to implement a dynamic schema with a non-dynamic pydantic model.

I agree on your analysis (i.e. we're using pydantic for 2 different things), but I do think we can make the pydantic model static, while keeping the configuration dynamic. 

My initial suggestion was to shape the raw object even before passing it to the (hopefully static) pydantic model, but if we want to validate the raw config structure without any modification, we could indeed use 2 separate models. 


No matter the solution we choose, do you agree on using pydantic objects instead of dicts in collect.py ?

ryneeverett:
> ...I do think we can make the pydantic model static, while keeping the configuration dynamic.

Coming back to this after not thinking about it for a while, I suspect you're right.

Maybe rather than only validate the selected main configuration, we could have a `main_configs: list[MainSectionConfig]` field and then add a `main` property (not as part of the schema) that returns the correct one.

> No matter the solution we choose, do you agree on using pydantic objects instead of dicts in collect.py ?

So long as the model is dynamic it seems like it would be less readable because we'd be replacing the dict lookups with `getattr`'s, so I think making the pydantic model static is a prerequisite.

PR #1172, opened 2025-12-29
Fix --flavor feature. Resolve #1166.

### LINKED ISSUE #1166: Flavor feature seems broken

ryneeverett:
This regression was probably introduced when adding toml support. The main issue is that the code is "indecisive" about whether flavors should be stored as top-level fields (`"flavor.myflavor"`) or a nested dictionary field (`"flavor": { "myflavor": ... }`). While unquoted sections are parsed into the nested structure in toml, top-level sections seem to be a bit easier to work with, so we now ensure that the data structure is flattened into top-level flavor fields.

Additionally:
- ini2toml now does not result in quoted flavor sections. This makes no functional difference but it does align with the docs and IMO is a bit more attractive.
- `interactive` is now optional because otherwise we'd have to set it for every flavor including "inactive" ones.
- Flavor configuration is now tested in several ways whereas it was not exercised in the test suite at all before.

### DIFF
--- a/bugwarrior/config/ini2toml_plugin.py
+++ b/bugwarrior/config/ini2toml_plugin.py
@@ -123,7 +123,14 @@ def process_values(doc: IntermediateRepr) -> IntermediateRepr:
     return doc
 
 
+def unquote_flavors(file_contents: str) -> str:
+    return re.sub(
+        r'\n\["flavor\.(?P<flavor>[^"]*)"\]', r'\n[flavor.\g<flavor>]', file_contents
+    )
+
+
 def activate(translator: Translator):
     profile = translator["bugwarriorrc"]
     profile.description = "Convert 'bugwarriorrc' files to 'bugwarrior.toml'"
     profile.intermediate_processors.append(process_values)
+    profile.post_processors.append(unquote_flavors)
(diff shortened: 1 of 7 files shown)


ryneeverett:
@Lotram Sorry if I burned you out in #1167. This is a more minimalist/direct approach to fixing the regression but it doesn't preclude the refactoring discussed there. I'd like to get 2.1.0 released in a week or so and would like to get the known 2.0.0 regressions fixed. Any comment?

Lotram:
Sorry I didn't open GitHub in the last 2 weeks. Sure let's merge this, having actual tests about flavors is already a win!

ryneeverett:
> Sorry I didn't open GitHub in the last 2 weeks.

No apology needed. A lot of people go silent for a couple weeks this time of year.

PR #1174, opened 2026-01-04
2.1.0 changelog

ryneeverett:
Drafted until [milestones](https://github.com/GothenburgBitFactory/bugwarrior/milestone/2) are complete.


ryneeverett:
@Lotram I think we're in good shape for a release. What do you think?

PR #1177, opened 2026-01-06
remove dependency to importlib-metadata

Lotram:
Since python 3.8, this package is in the standard library.


ryneeverett (approved):
The CI check failure is due to the flaky writing of the test -- it installs the PR branch and then later tries to build the develop branch. Well if a PR removes any dependency then they're going to be missing.

Issue #1134, opened 2025-11-14
Mock out optional service libraries from test suite


ryneeverett:
I don't believe they're exercised at all during tests so it should be a simple replacement with a generic stub. We can then update the developer docs to only require installing `[test]` dependencies rather than `[all]`.

We should add a CI job to install all optional dependencies to make sure they continue to be resolvable once we're no longer doing so in development.

Lotram:
On a related subject: dev dependencies could be moved to a "dependency-group" instead of an optional dependency, as defined in [this PEP](https://peps.python.org/pep-0735/#limitations-of-extras). 
[uv fully supports this](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups)

Issue #955, opened 2022-09-01
Easier Per-Service Configuration Testing


ryneeverett:
Per https://github.com/ralphbean/bugwarrior/pull/954#issuecomment-1228110944, we could make it simpler and more straightforward to test different service configuration values if we replaced our per-service `self.config` definition with some magic in the base class that allowed for better ergonomics, e.g.:

```
def test_filter_repos_include(self):
    self.service.config['deck.include_board_ids'] = ['5']
    self.assertTrue(self.service.filter_repos({'title': 'testboard', 'id': 5}))
    self.assertFalse(self.service.filter_repos({'title': 'testboard', 'id': 6}))

def test_filter_repos_exclude(self):
    self.service.config['deck.exclude_board_ids'] = ['5']
    self.assertFalse(self.service.filter_repos({'title': 'testboard', 'id': 5}))
    self.assertTrue(self.service.filter_repos({'title': 'testboard', 'id': 6}))
```

It may or may not make sense to put `config` in the top level namespace instead of within `service`.

ryneeverett:
This will be resolved by #1225.

PR #1181, opened 2026-01-14
Static config model

Lotram:
Before changing the actual config validation implementation, I think it's easier to use another `Config` model once everything has been validated, and use this model downstream. 
I plan to change the validate_config implementation once we agree on these changes, and this `Config` model.

What do you think @ryneeverett  ? 

There is a failing test related to flavors in test_load.py, It will be fixed once I modify the validation step

### DIFF
--- a/bugwarrior/config/load.py
+++ b/bugwarrior/config/load.py
@@ -9,7 +9,7 @@ try:
 except ImportError:
     import tomli as tomllib  # backport
 
-from bugwarrior.config.validation import Config, validate_config
+from .validation import Config, validate_config
 
 # The name of the environment variable that can be used to ovewrite the path
 # to the bugwarriorrc file
@@ -54,26 +54,15 @@ def get_config_path():
 
 
 def format_config(config: dict) -> dict[str, Any]:
-    # Build flavors from 'flavor' table
-    flavors = {
-        name: flavor_config for name, flavor_config in config.pop("flavor", {}).items()
-    }
-
-    # Handle 'general' as top-level key (TOML format, tests)
     if "general" in config:
-        flavors["general"] = config.pop("general")
+        config.setdefault("flavor", {})["general"] = config.pop("general")
 
-    services = {
-        section: {**config.pop(section), "target": section}
+    config["services"] = [
+        {**config.pop(section), "target": section}
         for section in list(config)
-        if section not in {"hooks", "notifications"}
-    }
-
-    return {
-        "flavors": flavors,
-        "services": services,
-        **config,  # remaining: "hooks" and "notifications" (if present)
-    }
+        if section not in {"hooks", "notifications", "flavor"}
+    ]
+    return config
 
 
 def parse_toml_file(configpath: str) -> dict:
@@ -130,7 +119,7 @@ def parse_file(configpath: str) -> dict:
 def load_config(main_section, interactive, quiet) -> Config:
     configpath = get_config_path()
     rawconfig = parse_file(configpath)
-    rawconfig['flavors'][main_section]['interactive'] = interactive
+    rawconfig['flavor'][main_section]['interactive'] = interactive
     config = validate_config(rawconfig, main_section, configpath)
     configure_logging(
         config.main.log_file, 'WARNING' if quiet else config.main.log_level
--- a/bugwarrior/config/load.py
+++ b/bugwarrior/config/load.py
@@ -1,14 +1,15 @@
-import codecs
 import configparser
 import logging
 import os
+from pathlib import Path
+from typing import Any
 
 try:
     import tomllib  # python>=3.11
 except ImportError:
     import tomli as tomllib  # backport
 
-from . import schema
+from bugwarrior.config.validation import Config, validate_config
 
 # The name of the environment variable that can be used to ovewrite the path
 # to the bugwarriorrc file
@@ -52,58 +53,87 @@ def get_config_path():
     return paths[0]
 
 
-def parse_file(configpath: str) -> dict:
-    if os.path.splitext(configpath)[-1] == '.toml':
-        with open(configpath, 'rb') as f:
-            config = tomllib.load(f)
-        # Flatten flavors into top-level sections (if they're unquoted).
-        for k, v in config.get('flavor', {}).items():
-            config[f'flavor.{k}'] = v
-        config.pop('flavor', None)
-    else:
-        rawconfig = BugwarriorConfigParser()
-        with codecs.open(configpath, "r", "utf-8") as buff:
-            rawconfig.read_file(buff)
-        config = {}
-        for section in rawconfig.sections():
-            if section in ['hooks', 'notifications']:
-                config[section] = dict(rawconfig[section])
-            elif section == 'general':
-                config[section] = {
-                    k.replace('log.', 'log_'): v for k, v in rawconfig[section].items()
-                }
-            elif section.startswith('flavor.'):
-                config[section] = {
-                    k.replace('.', '_'): v for k, v in rawconfig[section].items()
-                }
-            else:
-                service = rawconfig[section].pop('service')
-                service_prefix = 'ado' if service == 'azuredevops' else service
-                config[section] = {'service': service}
-                for k, v in rawconfig[section].items():
-                    try:
-                        prefix, key = k.split('.')
-                    except ValueError:  # missing prefix
-                        prefix = None
-                        key = k
-                    if prefix != service_prefix:
-                        raise SystemExit(
-                            f"[{section}]\n{k} <-expected prefix "
-                            f"'{service_prefix}': did you mean "
-                            f"'{service_prefix}.{key}'?"
-                        )
-                    config[section][key] = v
+def format_config(config: dict) -> dict[str, Any]:
+    # Build flavors from 'flavor' table
+    flavors = {
+        name: flavor_config for name, flavor_config in config.pop("flavor", {}).items()
+    }
+
+    # Handle 'general' as top-level key (TOML format, tests)
+    if "general" in config:
+        flavors["general"] = config.pop("general")
+
+    services = {
+        section: {**config.pop(section), "target": section}
+        for section in list(config)
+        if section not in {"hooks", "notifications"}
+    }
+
+    return {
+        "flavors": flavors,
+        "services": services,
+        **config,  # remaining: "hooks" and "notifications" (if present)
+    }
+
+
+def parse_toml_file(configpath: str) -> dict:
+    with open(configpath, 'rb') as file:
+        return tomllib.load(file)
+
+
+def parse_ini_file(configpath: str) -> dict:
+    rawconfig = BugwarriorConfigParser()
+    with open(configpath, encoding="utf-8") as buff:
+        rawconfig.read_file(buff)
+
+    config = {"flavor": {}}
+    for section in rawconfig.sections():
+        if section in ['hooks', 'notifications']:
+            config[section] = dict(rawconfig[section])
+        elif section == 'general' or section.startswith('flavor.'):
+            name = section.removeprefix('flavor.')
+            config["flavor"][name] = {
+                key.replace('.', '_'): value
+                for key, value in rawconfig[section].items()
+            }
+
+        # All other sections are assumed to be services
+        else:
+            service = rawconfig[section].pop('service')
+            service_prefix = 'ado' if service == 'azuredevops' else service
+            config[section] = {'service': service}
+            for key, value in rawconfig[section].items():
+                try:
+                    prefix, unprefixed_key = key.split('.')
+                except ValueError:  # missing prefix
+                    prefix = None
+                    unprefixed_key = key
+                if prefix != service_prefix:
+                    raise SystemExit(
+                        f"[{section}]\n{key} <-expected prefix "
+                        f"'{service_prefix}': did you mean "
+                        f"'{service_prefix}.{unprefixed_key}'?"
+                    )
+                config[section][unprefixed_key] = value
+
     return config
 
 
-def load_config(main_section, interactive, quiet) -> dict:
+def parse_file(configpath: str) -> dict:
+    if Path(configpath).suffix == '.toml':
+        config = parse_toml_file(configpath)
+    else:
+        config = parse_ini_file(configpath)
+    return format_config(config)
+
+
+def load_config(main_section, interactive, quiet) -> Config:
     configpath = get_config_path()
     rawconfig = parse_file(configpath)
-    rawconfig[main_section]['interactive'] = interactive
-    config = schema.validate_config(rawconfig, main_section, configpath)
+    rawconfig['flavors'][main_section]['interactive'] = interactive
+    config = validate_config(rawconfig, main_section, configpath)
     configure_logging(
-        config[main_section].log_file,
-        'WARNING' if quiet else config[main_section].log_level,
+        config.main.log_file, 'WARNING' if quiet else config.main.log_level
     )
     return config
 
--- a/bugwarrior/config/schema.py
+++ b/bugwarrior/config/schema.py
@@ -197,6 +197,7 @@ class ServiceConfig(_ServiceConfig):
     .. _Pydantic: https://docs.pydantic.dev/latest/
     """
 
+    # Added before validation (computed field)
     service: str
     target: str
 
--- a/bugwarrior/config/schema.py
+++ b/bugwarrior/config/schema.py
@@ -2,7 +2,6 @@ import logging
 import os
 from pathlib import Path
 import re
-import sys
 import typing
 from typing import Annotated, Any, Generic, Literal
 
@@ -13,7 +12,6 @@ from pydantic import (
     BeforeValidator,
     ConfigDict,
     Field,
-    TypeAdapter,
     ValidationInfo,
     computed_field,
     field_validator,
@@ -21,8 +19,6 @@ from pydantic import (
 )
 from pydantic_core import PydanticCustomError
 import taskw
-
-from bugwarrior.collect import get_service
 
 from .data import BugwarriorData, get_data_path
 
@@ -184,117 +180,6 @@ class Notifications(BaseConfig):
     only_on_new_tasks: bool = False
 
 
-class SchemaBase(BaseConfig):
-    # Allow extra top-level sections so all targets don't have to be selected.
-    model_config = ConfigDict(extra="ignore")
-
-    hooks: Hooks = Hooks()
-    notifications: Notifications = Notifications()
-
-
-def get_validation_error_enhanced_messages(
-    error: pydantic.ValidationError,
-) -> list[str]:
-    errors = []
-    for _error in error.errors():
-        msg = _error["msg"]
-        if _error["type"] == "extra_forbidden":
-            msg = "unrecognized option"
-        loc = _error["loc"]
-        loc_len = len(loc)
-
-        if loc_len == 1 or (loc_len > 1 and loc[1] == "__root__"):
-            formatted_error_loc = f"[{loc[0]}]"
-        elif loc_len == 2:
-            formatted_error_loc = f"[{loc[0]}]\n{loc[1]}"
-            if _error["type"] != "missing":
-                formatted_error_loc = f"{formatted_error_loc} = {_error['input']}"
-        else:
-            raise ValueError(
-                "Configuration should not be nested more than two layers deep."
-            )
-        errors.append(f"{formatted_error_loc}  <- {msg}\n")
-    return errors
-
-
-def raise_validation_error(msg, config_path, no_errors=1) -> typing.NoReturn:
-    log.error(
-        ("Validation error" if no_errors == 1 else f"{no_errors} validation errors")
-        + f" found in {config_path}\n"
-        f"See https://bugwarrior.readthedocs.io\n\n{msg}"
-    )
-    sys.exit(1)
-
-
-def get_target_validator(targets):
-    @model_validator(mode='before')
-    @classmethod
-    def compute_target(cls, values):
-        for target in targets:
-            values[target]['target'] = target
-        return values
-
-    return compute_target
-
-
-def validate_config(config: dict, main_section: str, config_path: str) -> dict:
-    # Pre-validate the minimum requirements to build our pydantic models.
-    try:
-        main = config[main_section]
-    except KeyError:
-        raise_validation_error(f"No section: '{main_section}'", config_path)
-    try:
-        targets = TypeAdapter(ConfigList).validate_python(main['targets'])
-    except KeyError:
-        raise_validation_error(
-            f"No option 'targets' in section: '{main_section}'", config_path
-        )
-    try:
-        configmap = {target: config[target] for target in targets}
-    except KeyError as e:
-        raise_validation_error(f"No section: '{e.args[0]}'", config_path)
-    servicemap = {}
-    for target, serviceconfig in configmap.items():
-        try:
-            servicemap[target] = serviceconfig['service']
-        except KeyError:
-            raise_validation_error(
-                f"No option 'service' in section: '{target}'", config_path
-            )
-
-    # Construct Service Models
-    target_schemas = {
-        target: (get_service(service).CONFIG_SCHEMA, ...)
-        for target, service in servicemap.items()
-    }
-
-    # Construct Flavors
-    flavor_schemas = {
-        section: (MainSectionConfig, ...)
-        for section in config.keys()
-        if section.startswith('flavor.')
-    }
-
-    # Construct Validation Model
-    bugwarrior_config_model = pydantic.create_model(
-        'bugwarriorrc',
-        __base__=SchemaBase,
-        __validators__={'compute_target': get_target_validator(targets)},
-        general=(MainSectionConfig, ...),
-        **flavor_schemas,
-        **target_schemas,
-    )
-
-    # Validate
-    try:
-        # Convert top-level model to dict since target names are dynamic and
-        # a bunch of calls to getattr(config, target) inhibits readability.
-        return dict(bugwarrior_config_model.model_validate(config))
-    except pydantic.ValidationError as e:
-        errors = get_validation_error_enhanced_messages(e)
-        raise_validation_error("\n".join(errors), config_path, no_errors=len(errors))
-
-
 # Dynamically add template fields to model.
 _ServiceConfig = pydantic.create_model(
     "_ServiceConfig",
@@ -312,9 +197,11 @@ class ServiceConfig(_ServiceConfig):
     .. _Pydantic: https://docs.pydantic.dev/latest/
     """
 
+    service: str
+    target: str
+
     # Added during validation (computed field)
     templates: dict = {}
-    target: typing.Optional[str] = None
 
     # Optional fields shared by all services.
     only_if_assigned: str = ""
--- a/bugwarrior/config/validation.py
+++ b/bugwarrior/config/validation.py
@@ -16,12 +16,12 @@ if TYPE_CHECKING:
 log = logging.getLogger(__name__)
 
 
-def _format_section_error(section: str | int, msg: str) -> str:
+def _format_section_error(section: str, msg: str) -> str:
     return f"[{section}]  <- {msg}\n"
 
 
 def _format_field_error(
-    section: str | int, field: str | int, msg: str, error: ErrorDetails | dict
+    section: str, field: str, msg: str, error: ErrorDetails | dict
 ) -> str:
     formatted = f"[{section}]\n{field}"
     if error["type"] != "missing":
@@ -56,6 +56,7 @@ def _format_service_error(
     if len(loc) == 2 or loc[-1] == "__root__":
         return _format_section_error(target, msg)
 
+    assert isinstance(loc[2], str)
     return _format_field_error(target, loc[2], msg, error)
 
 
@@ -70,6 +71,7 @@ def _format_flavor_error(error: ErrorDetails | dict) -> str:
     msg = error["msg"]
 
     flavor_name = loc[0]
+    assert isinstance(flavor_name, str)
 
     if len(loc) == 1:
         raise ValueError(f"Unexpected error loc with single element: {loc}")
@@ -77,6 +79,7 @@ def _format_flavor_error(error: ErrorDetails | dict) -> str:
     if loc[-1] == "__root__":
         return _format_section_error(flavor_name, msg)
 
+    assert isinstance(loc[1], str)
     return _format_field_error(flavor_name, loc[1], msg, error)
 
 
@@ -100,6 +103,7 @@ def _format_extra_section_error(error: ErrorDetails | dict) -> str:
     if loc[-1] == "__root__":
         return _format_section_error(section_name, msg)
 
+    assert isinstance(loc[1], str)
     return _format_field_error(section_name, loc[1], msg, error)
 
 
@@ -147,7 +151,7 @@ def validate_config(config: dict, main_section: str, config_path: str) -> "Confi
         error_messages.extend(_format_flavor_error(err) for err in error.errors())
 
     # Check for misquoted flavor sections (e.g., ["flavor.myflavor"] instead of [flavor.myflavor])
-    services_list = []
+    raw_service_configs = []
     for service in config.pop("services", []):
         target = service["target"]
         if target.startswith("flavor."):
@@ -156,17 +160,17 @@ def validate_config(config: dict, main_section: str, config_path: str) -> "Confi
                 "Use [flavor.name] (not quoted) to define a flavor.\n"
             )
         else:
-            services_list.append(service)
+            raw_service_configs.append(service)
 
     # Validate service configs
-    ServiceConfigType = get_service_config_union_type(services_list)
+    ServiceConfigType = get_service_config_union_type(raw_service_configs)
     try:
         service_configs = TypeAdapter(list[ServiceConfigType]).validate_python(
-            services_list
+            raw_service_configs
         )
     except ValidationError as error:
         error_messages.extend(
-            _format_service_error(err, services_list) for err in error.errors()
+            _format_service_error(err, raw_service_configs) for err in error.errors()
         )
         service_configs = []
 
@@ -175,7 +179,9 @@ def validate_config(config: dict, main_section: str, config_path: str) -> "Confi
         error_messages.append(f"No section: '{main_section}'\n")
 
     # Check targets exist for all flavors
-    available_targets = {service_config["target"] for service_config in services_list}
+    available_targets = {
+        service_config["target"] for service_config in raw_service_configs
+    }
     for flavor_name, flavor in flavors.items():
         missing = sorted(set(flavor.targets) - available_targets)
         for target in missing:
--- a/bugwarrior/config/validation.py
+++ b/bugwarrior/config/validation.py
@@ -31,21 +31,24 @@ def _format_field_error(
     return f"{formatted}  <- {msg}\n"
 
 
-def _format_service_error(error: ErrorDetails | dict) -> str:
+def _format_service_error(
+    error: ErrorDetails | dict, services: list[dict[str, Any]]
+) -> str:
     """Format validation error for service configs (discriminated union).
 
     loc structure:
-    - (target,) - discriminator error (missing 'service')
-    - (target, service_type) - model-level error
-    - (target, service_type, '__root__') - model validator
-    - (target, service_type, field) - field error
+    - (index,) - discriminator error (missing 'service')
+    - (index, service_type) - model-level error
+    - (index, service_type, '__root__') - model validator
+    - (index, service_type, field) - field error
     """
     loc = error["loc"]
     msg = error["msg"]
     if error["type"] == "extra_forbidden":
         msg = "unrecognized option"
 
-    target = loc[0]
+    index = int(loc[0])
+    target = services[index].get("target", f"service[{index}]")
 
     if len(loc) == 1:
         return _format_section_error(
@@ -76,17 +79,24 @@ def _format_flavor_error(error: ErrorDetails | dict) -> str:
     return _format_field_error(flavor_name, loc[1], msg, error)
 
 
-def _format_simple_section_error(error: ErrorDetails | dict, section_name: str) -> str:
-    """Format validation error for simple sections (hooks, notifications)."""
+def _format_extra_section_error(error: ErrorDetails | dict) -> str:
+    """Format validation error for extra options (hooks, notifications).
+
+    loc structure:
+    - (section_name, field) - field error
+    - (section_name, '__root__') - model validator
+    """
     loc = error["loc"]
     msg = error["msg"]
     if error["type"] == "extra_forbidden":
         msg = "unrecognized option"
 
-    if not loc or loc[-1] == "__root__":
+    section_name = str(loc[0])
+
+    if len(loc) == 1 or loc[-1] == "__root__":
         return _format_section_error(section_name, msg)
 
-    return _format_field_error(section_name, loc[0], msg, error)
+    return _format_field_error(section_name, loc[1], msg, error)
 
 
 def raise_validation_error(msg, config_path, error_count=1) -> typing.NoReturn:
@@ -98,10 +108,20 @@ def raise_validation_error(msg, config_path, error_count=1) -> typing.NoReturn:
     sys.exit(1)
 
 
-def get_service_config_union_type(services: dict[str, dict[str, Any]]):
+def get_service_config_union_type(services: list[dict[str, Any]]):
+    """
+    Return a Union type of the ServiceConfig subclasses of the services actually configured.
+
+    We don't want to include all available services because each service must be imported
+    which could have side-effects.
+
+    The returned type takes advantaged of Pydantic's "Discriminated Unions" feature
+    to ensure that each service configuration is validated against
+    the ServiceConfig corresponding to the correct service.
+    """
     service_config_classes = tuple(
         get_service(service["service"]).CONFIG_SCHEMA
-        for service in services.values()
+        for service in services
         if "service" in service
     )
 
@@ -114,62 +134,44 @@ def get_service_config_union_type(services: dict[str, dict[str, Any]]):
 
 def validate_config(config: dict, main_section: str, config_path: str) -> "Config":
     error_messages: list[str] = []
-
+    raw_flavors = config.pop("flavor", {})
     # Validate flavors
-    flavors: dict[str, MainSectionConfig] = {}
     try:
-        flavors = TypeAdapter(dict[str, MainSectionConfig]).validate_python(
-            config.get("flavors", {})
-        )
+        flavors = TypeAdapter(dict[str, MainSectionConfig]).validate_python(raw_flavors)
     except pydantic.ValidationError as error:
+        flavors = {}
         error_messages.extend(_format_flavor_error(err) for err in error.errors())
 
     # Check for misquoted flavor sections (e.g., ["flavor.myflavor"] instead of [flavor.myflavor])
-    services_config = config.get("services", {})
-    misquoted_flavors = [name for name in services_config if name.startswith("flavor.")]
-    for section in misquoted_flavors:
-        error_messages.append(
-            f'["{section}"]  <- Did you mean [{section}]?\n'
-            "Use [flavor.name] (not quoted) to define a flavor.\n"
-        )
-        del services_config[section]
+    services_list = []
+    for service in config.pop("services", []):
+        target = service["target"]
+        if target.startswith("flavor."):
+            error_messages.append(
+                f'["{target}"]  <- Did you mean [{target}]?\n'
+                "Use [flavor.name] (not quoted) to define a flavor.\n"
+            )
+        else:
+            services_list.append(service)
 
     # Validate service configs
-    ServiceConfigType = get_service_config_union_type(services_config)
-    service_configs: dict[str, ServiceConfigType] = {}
+    ServiceConfigType = get_service_config_union_type(services_list)
     try:
-        service_configs = TypeAdapter(dict[str, ServiceConfigType]).validate_python(
-            services_config
-        )
-    except pydantic.ValidationError as error:
-        error_messages.extend(_format_service_error(err) for err in error.errors())
-
-    # Validate hooks
-    hooks = Hooks()
-    try:
-        hooks = TypeAdapter(Hooks).validate_python(config.get("hooks", {}))
-    except pydantic.ValidationError as error:
-        error_messages.extend(
-            _format_simple_section_error(err, "hooks") for err in error.errors()
-        )
-
-    # Validate notifications
-    notifications = Notifications()
-    try:
-        notifications = TypeAdapter(Notifications).validate_python(
-            config.get("notifications", {})
+        service_configs = TypeAdapter(list[ServiceConfigType]).validate_python(
+            services_list
         )
     except pydantic.ValidationError as error:
         error_messages.extend(
-            _format_simple_section_error(err, "notifications") for err in error.errors()
+            _format_service_error(err, services_list) for err in error.errors()
         )
+        service_configs = []
 
     # Check main_section exists in flavors
-    if main_section not in flavors:
+    if main_section not in raw_flavors:
         error_messages.append(f"No section: '{main_section}'\n")
 
     # Check targets exist for all flavors
-    available_targets = set(service_configs.keys())
+    available_targets = {service_config["target"] for service_config in services_list}
     for flavor_name, flavor in flavors.items():
         missing = set(flavor.targets) - available_targets
         if missing:
@@ -177,20 +179,25 @@ def validate_config(config: dict, main_section: str, config_path: str) -> "Confi
                 f"[{flavor_name}] missing targets: {', '.join(sorted(missing))}\n"
             )
 
+    main = flavors.get(main_section, MainSectionConfig(targets=[]))
+    filtered_service_configs = [
+        service_config
+        for service_config in service_configs
+        if service_config.target in main.targets
+    ]
+
+    try:
+        _config = Config(service_configs=filtered_service_configs, main=main, **config)
+    except pydantic.ValidationError as error:
+        error_messages.extend(
+            _format_extra_section_error(err) for err in error.errors()
+        )
+
     if error_messages:
         raise_validation_error(
             "".join(error_messages), config_path, error_count=len(error_messages)
         )
-
-    main = flavors[main_section]
-    filtered_service_configs = [service_configs[target] for target in main.targets]
-
-    return Config(
-        service_configs=filtered_service_configs,
-        main=main,
-        hooks=hooks,
-        notifications=notifications,
-    )
+    return _config
 
 
 class Config(BaseConfig):
--- a/bugwarrior/config/validation.py
+++ b/bugwarrior/config/validation.py
@@ -1,10 +1,8 @@
 import logging
 import sys
-import typing
-from typing import TYPE_CHECKING, Annotated, Any, Union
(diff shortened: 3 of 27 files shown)


bugwarrior/config/schema.py
```diff
@@ -387,3 +399,32 @@ def deprecate_project_name(cls, value):
...
+    # so I'm not sure we should keep this and get_service_config
+    @cached_property
+    def service_config_map(self) -> dict[str, ServiceConfig]:
+        return {
+            service_config.target: service_config
+            for service_config in self.service_configs
+        }
+
+    def get_service_config(self, target: str) -> ServiceConfig:
+        return self.service_config_map[target]
```
ryneeverett: I'm not sure what functionality you're conserving here, but if we're not using it I'd agree to delete it.

ryneeverett (commented):
It seems like you're going with my "consumer view" proposal from #1167:

> Maybe we're asking too much from pydantic to provide the ideal configuration interface for both inputs and outputs. Maybe we should consider another step in which the validation model is transformed into a view (like your quoted example) that is better optimized for consumers of the model.

That's fine and it's looking good. Some thoughts:

- I think `Config` should be renamed to `ConfigView` or something that more particularly describes its specific purpose. Of course, once it is passed out of bugwarrior/config to other parts of the code base it can be referred to simply as "config", but within the bugwarrior.config package that is too vague of a name. 
- it would be nice to keep the configuration "view" separate from the configuration file validation (schema.py). So maybe `validate_config` returns the validated `bugwarrior_config_model` and `Config` moves to load.py and is instantiated from `validated_data` in `load_config`.
- Is the `get_service()` -> `ServiceConfig.service_class` change necessary? I'd rather not require services to verbosely add that boilerplate, and it also violates the separation of concerns in service implementations -- ServiceConfig is for user configuration.

Lotram:
>  think Config should be renamed to ConfigView or something that more particularly describes its specific purpose. Of course, once it is passed out of bugwarrior/config to other parts of the code base it can be referred to simply as "config", but within the bugwarrior.config package that is too vague of a name.

I have no problem with this, except that my plan is to remove the existing dynamic config object, so there would be only 1 Config model after all my changes. Still, I can rename `ConfigView` if you prefere

> it would be nice to keep the configuration "view" separate from the configuration file validation (schema.py). So maybe validate_config returns the validated bugwarrior_config_model and Config moves to load.py and is instantiated from validated_data in load_config.

Sure. 

> Is the get_service() -> ServiceConfig.service_class change necessary? I'd rather not require services to verbosely add that boilerplate, and it also violates the separation of concerns in service implementations -- ServiceConfig is for user configuration.

So I'm not entirely satisfied with this solution either, I agree about the boilerplate. Not so much about separation of concerns, as the `Config` and `Service` objects are tightly coupled anyway, but I get your point

My main concern with the existing implementation is that we need to call entrypoints (through importlib) everytime we want to retrieve a `Service` class, but IMO we should do it only once at the beginning, and then have the classes available somewhere.

We could do

```
class ServiceConfig(BaseModel):
   @cached_property
   def service_class(self):
        return get_service(self.service)
```
which does not really change anything about separation of concerns, but at least remove any boilerplate (at the cost of a slightly less precise typing, but that's not really a problem).

I might be able to avoid some usees of `service_class` by iterating over `Config.services` instead (but then I need to find a way to build this `services` property without `ServiceConfig.service_class`)

ryneeverett:
I guess I misinterpreted your plan. Forget about the naming and code moving suggestions since they seem to be based on scaffolding.

> I plan to change the validate_config implementation once we agree on these changes, and this Config model.

I don't have any objections to the Config model at the moment, but it seems like it's going to have to change a lot. As long as the configuration file is fully validated by pydantic I don't foresee any objections.

> Not so much about separation of concerns, as the Config and Service objects are tightly coupled anyway

Perhaps I used the wrong terminology, but my concern here was not from an architectural point of view but from a service implementer's point of view. Somebody writing a service (who might not know python or have much of any programming experience) should be able to write their Config class as a simple data structure describing user configuration.

However, I would distinguish between unidirectional coupling and bidirectional coupling, the latter being worse. Config doesn't need to know about Service, which I think is another valid point against the current `service_class` methods.

> My main concern with the existing implementation is that we need to call entrypoints (through importlib) everytime we want to retrieve a Service class

The same Service class is only being retrieved more than once if a user has multiple instances of the same service configured, right? Why couldn't we just make `get_service` a cached function if this is a concern?

Lotram:
@ryneeverett I'd like your opinion on commit c6e2a8. My opinion is the current implementation relies on a mix of function validations and some pydantic model validations, I think we should choose either one method or the other, so I suggest two alternatives for `validate_config`.
* `validate_config_using_type_adapters`, that uses only `TypeAdapter` and some other checks, since we don't really need to create a model to validate objects that we won't use afterwards (i.e. other flavors, and their targets)
* `validate_config_using_model`, that only uses a pydantic model with some `model_validator` functions. Opposite solution, almost everything is done in the model, we just need a function to set the type of `service_configs` list

Since we want to validate all flavors, I think we should also validate all service configs, not only the ones targeted by the choosen flavor. 

A note on `get_service_config_union_type`: it can seems complex at first sight, but using a [union discriminator](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) is actually the most pydantic way to have polymorphism, IMO. We could define this union at import time, but this would require to load all available services, not only the ones referenced in the config, I'm not sure it's worth it.

Also, error messages may be wrong for now, I'll make sure they're are close as possible to current ones when we choose one implem or the other

bugwarrior/config/schema.py
```diff
@@ -405,6 +410,97 @@ def service_class(self) -> type["Service"]:
...
+        ],
+        **config,  # only remaining keys are "hooks" and "notifications" (if present)
+    }
+
+
+def get_service_config_union_type(services: Iterable[dict[str, Any]]):
+    service_config_classes = [
+        get_service(service["service"]).CONFIG_SCHEMA
+        for service in services
+        if "service" in service
```
ryneeverett: I take it this line is to make type checking happy? Would raising an exception if there is no "service" in service satisfy the type checker? We shouldn't be passing over something like this silently.

bugwarrior/config/schema.py
```diff
@@ -405,6 +410,97 @@ def service_class(self) -> type["Service"]:
...
+    ]
+    return Annotated[
+        reduce(or_, service_config_classes), Field(discriminator='service')
```
ryneeverett: I don't get what's going on here. Seems like we want to deduplicate service classes but I don't follow how the bitwise OR operator does that or why there isn't an explicit Union type.

bugwarrior/config/schema.py
```diff
@@ -416,14 +512,70 @@ def services(self) -> "list[Service]":
...
-    # so I'm not sure we should keep this and get_service_config
+
+def validate_config_using_model(
+    config: dict, main_section: str, config_path: str
+) -> "WholeConfig":
+    config = format_config(config)
+    # NOTE: I think this is an elegant way to 'hide' the dynamical part of the config definition
+    # but it might be a bit too implicit
+    # for more info, see https://docs.pydantic.dev/latest/internals/resolving_annotations/#resolving-annotations-at-class-definition
+    ServiceConfigType = get_service_config_union_type(config["services"])
```
ryneeverett: I'm not entirely following. is the relevant example in the linked docs 'UnknownType`? It isn't clear to me from the docs that the namespace in which a model is instantiated is included when looking up types. (And how do type checkers handle this?)

ryneeverett (commented):
> My opinion is the current implementation relies on a mix of function validations and some pydantic model validations, I think we should choose either one method or the other...

My impression is that the type adapter approach is a more consistent and complete implementation of the current approach of "pre-validation" followed by the model validation. It seems like an improvement, but I'm not sure it satisfies your either/or principle.

The WholeConfig would actually replace Config, right? And with the WholeConfig approach, we eliminate the "stages" of validation, right? This seems nicer for users because they get all the validation errors at once.

> We could define this union at import time, but this would require to load all available services, not only the ones referenced in the config, I'm not sure it's worth it.

I would tend to agree. It would be nice if unconfigured services weren't imported since this could result in side effects or exceptions. We shouldn't insist that users have a perfect installation and python environment when it comes to features they're not using.

> Since we want to validate all flavors, I think we should also validate all service configs, not only the ones targeted by the choosen flavor.

:+1: Shouldn't we also be validating that all targets are defined in all flavors rather than just the selected one?

Lotram:
> My impression is that the type adapter approach is a more consistent and complete implementation of the current approach of "pre-validation" followed by the model validation. It seems like an improvement, but I'm not sure it satisfies your either/or principle.

Yes, with that option, we would have the `validate` part using type adapters, no models, ensuring everything is ok. The model would then only be used for downstream processing (kind of like the config view you were talking earlier). The biggest improvement IMO is we don't have to create a dynamic model simply for validation

> The WholeConfig would actually replace Config, right? And with the WholeConfig approach, we eliminate the "stages" of validation, right? 

Yes exactly, with that solution, validation is almost entirely included in the model. IMO The small downside is: this object is carrying a lot of data we won't ever use after validation, making the developer experience a bit more complex, but that's probably fine. 

> This seems nicer for users because they get all the validation errors at once.

In both cases, I think we can indeed aim for that. That's what I was starting to do with the `validation_errors` in `validate_config_using_type_adapters`, gathering all errors and raising them all at once. 
I think either solution should be completely transparent for the end user, and we should choose the simplest / most readable / easiest to maintain implementation from a developer perspective (as long as we can indeed guarantee it won't change anything for the user)

bugwarrior/config/schema.py
Lotram: No, this line is to avoid raising an error right here because this section has no `service` key. We still want to raise an error, but that will be done by the type validation (either through TypeAdapter or the model field), so we can get a proper pydantic error, more consistent with the rest. 

In case it's missing, pydantic will complain the discriminator field is missing.

bugwarrior/config/schema.py
Lotram: I admit it's not really explicit. Basically, let's say our targets are a GitHub and Jira services, we want to build the following type: `GitHubServiceConfig | JiraServiceConfig`, so pydantic can use the proper subclass when instanciating each service config, using the discriminator. 

the `or_` function is simply the function called by the `|` operator. The `Union` is created by doing `type1 | type2`

I need to check, but I may be able to do `Union[*service_config_classes]` instead of `reduce(or_, service_config_classes)`, result is the same, but probably easier to understand

Edit, yes it works:
```
In [9]: Union[*service_config_classes] == reduce(or_, service_config_classes)
Out[9]: True
```

Lotram: I simply need to check it works in all supported python versions

bugwarrior/config/schema.py
Lotram: Yep the relevant example here is `UnknownType` (+ explanations about how types are resolved)

> it isn't clear to me from the docs that the namespace in which a model is instantiated is included when looking up types.

Among other thing, the doc says:

> For the locals argument, Pydantic will try to resolve symbols in the following namespaces, sorted by highest priority:
>  *  A namespace created on the fly, containing the current class name (`{cls.__name__: cls}`). 

which is exactly the point of this line. Afterwards, we have:
```
In [1]: locals()["ServiceConfigType"]
Out[1]: typing.Annotated[typing.Union[bugwarrior.services.github.GithubConfig, .... (other config types)], FieldInfo(annotation=NoneType, required=True, discriminator='service') 
```
so the type can be resolved when building the `WholeConfig` object. 

I admit it's not explicit at all, but unfortunately I found no other way to override the type of `all_service_configs` at runtime (other alternative being the one we don't want: building this type before model creation, but using all service config classes, not just the ones targeted in the config).

Another solution would be to create a new model with `create_model`:
```
    ServiceConfigType = get_service_config_union_type(config["services"])

    _WholeConfig = create_model(
        "WholeConfig", __base__=WholeConfig, all_service_configs=list[ServiceConfigType]
    )
    return _WholeConfig(
        all_service_configs=config.pop("services"),
        main_section=main_section,
        flavors=config.pop("flavors"),
        config_path=config_path,
        **config,
    )
```

> And how do type checkers handle this?

Simple trick (which I should have put directly above `WholeConfig` instead):
```
if TYPE_CHECKING:
    ServiceConfigType = ServiceConfig
```

bugwarrior/config/schema.py
ryneeverett: Ah, I wasn't realizing that pipes were overloaded to make Unions in type annotations so it makes sense now. Nonetheless, the `Union[*service_config_classes]` form is preferable if it works.

bugwarrior/config/schema.py
ryneeverett: > > For the locals argument, Pydantic will try to resolve symbols in the following namespaces, sorted by highest priority:
> > 
> > * A namespace created on the fly, containing the current class name (`{cls.__name__: cls}`).
> 
> which is exactly the point of this line.

I can't find any way to read this that indicates that the `locals` of the namespace in which the model is instantiated is searched for annotations. My interpretation of this particular line is that `{'WholeConfig': WholeConfig}` is included -- purely for recursive references, like the docs say.

> I admit it's not explicit at all, but unfortunately I found no other way to override the type of `all_service_configs` at runtime (other alternative being the one we don't want: building this type before model creation, but using all service config classes, not just the ones targeted in the config).
> 
> Another solution would be to create a new model with `create_model`:
> 
> ```
>     ServiceConfigType = get_service_config_union_type(config["services"])
> 
>     _WholeConfig = create_model(
>         "WholeConfig", __base__=WholeConfig, all_service_configs=list[ServiceConfigType]
>     )
>     return _WholeConfig(
>         all_service_configs=config.pop("services"),
>         main_section=main_section,
>         flavors=config.pop("flavors"),
>         config_path=config_path,
>         **config,
>     )
> ```

If this behavior is in fact documented then I'm fine with leaving it as is and just adding a comment explaining that the local namespace is searched for type annotations when instantiating a model. If it is undocumented then I'd be more in favor of this `create_model` solution.

ryneeverett:
> IMO The small downside is: this object is carrying a lot of data we won't ever use after validation, making the developer experience a bit more complex, but that's probably fine.

Hmm, I was hoping we could move the data that is purely for validation purposes into "private" fields to make it clear that they aren't meant to be consumed but it seems that Pydantic considers any underscored fields to not be part of the model for validation either. A couple alternative ideas:

- delete that data from the model before returning it
- prefix with `private_` rather than `_`

> In both cases, I think we can indeed aim for that. That's what I was starting to do with the `validation_errors` in `validate_config_using_type_adapters`, gathering all errors and raising them all at once. I think either solution should be completely transparent for the end user, and we should choose the simplest / most readable / easiest to maintain implementation from a developer perspective (as long as we can indeed guarantee it won't change anything for the user)

:+1:

bugwarrior/config/schema.py
Lotram: You're right, I quoted the wrong bullet point, I should have used this one:
> The parent namespace of the class, if different from the globals described above. This is the [locals](https://docs.python.org/3/reference/datamodel.html#frame.f_locals) of the frame where the class is being defined. For Base, because the class is being defined in the module directly, this namespace won't be used as it will result in the globals being used again. For `Model`, the parent namespace is the locals of the frame of inner().

Our use-case is similar to their `Model` one, because in our case, the `WholeConfig` model is being built inside the `validate_config_using_model` (even though the class itself is defined outside).

Lotram:
So just to be clear, do you have any preference between the two solutions ? If not, I think I'll go with the `validate_config_using_type_adapters` one, so we can have a lighter model.

bugwarrior/config/schema.py
ryneeverett: Sorry but I can't agree with your interpretation. The parent namespace of the class is not "different from the globals described above" -- it is the module's `__dict__`. The docs clearly say "the locals of the frame where the class is being *defined*". That the locals in which the class is being *built/instantiated* are included seems to be undocumented behavior.

ryneeverett:
> So just to be clear, do you have any preference between the two solutions ? If not, I think I'll go with the `validate_config_using_type_adapters` one, so we can have a lighter model.

From the perspective of the maintainability of schema.py, I prefer WholeConfig. I would characterize model definitions as pydantic's high level API, whereas type adapters are a lower level API that people are less familiar with and is less intuitively obvious. I also think lengthy functions like `validate_config_using_type_adapters` are a bit of a code smell, but to be fair it is less smelly than the current `validate_config`.

From the perspective of the maintainability of bugwarrior as a whole, I'd rather compartmentalize implementation details within schema.py than let them leak out, so I agree that having a cleaner model to pass on to consumers is desirable. Can we not delete the fields we don't want to expose from the model before returning it? If there's no decent way to return a "lighter model" to consumers then I'd agree that the type adapter approach is preferable.

ryneeverett:
On the other hand, the two compromises we'd have to make for WholeConfig would seem to undercut its advantages:

- Deleting fields is probably a violation of pydantic's high level API, makes the code less intuitively obvious, and is arguably a code smell.
- In order for ServiceConfigType to work, we either need to rely on namespace magic (regardless of whether it's documented) or add an intermediary model using create_model, which arguably isn't pydantic's high level API.

So I guess I'd be in favor of the type adapter approach too.

Lotram:
> From the perspective of the maintainability of bugwarrior as a whole, I'd rather compartmentalize implementation details within schema.py than let them leak out, so I agree that having a cleaner model to pass on to consumers is desirable. Can we not delete the fields we don't want to expose from the model before returning it? If there's no decent way to return a "lighter model" to consumers then I'd agree that the type adapter approach is preferable.

If we think the model validation is cleaner, we could also use a first model to validate everything, and have another one afterwards (coming back to the "view" idea). 

> In order for ServiceConfigType to work, we either need to rely on namespace magic (regardless of whether it's documented) or add an intermediary model using create_model, which arguably isn't pydantic's high level API.

Since the only non-static part of the model is the list of service config classes, we could use a type adapter only for this part. Something like :

```
class WholeConfig(SchemaBase):
    all_service_configs: list[ServiceConfig]
    main_section: str
    config_path: str
    flavors: list[MainSectionConfig]

def validate_config_using_model(
    config: dict, main_section: str, config_path: str
) -> "WholeConfig":
    config = format_config(config)

    ServiceConfigType = get_service_config_union_type(config["services"])

    all_service_configs = TypeAdapter(list[ServiceConfigType]).validate_python(
        config["services"]
    )
    return WholeConfig(
        all_service_configs=all_service_configs,
        main_section=main_section,
        flavors=config.pop("flavors"),
        config_path=config_path,
        **config,
    )
```

Other alternative, regarding the pydantic doc discussion, since it does use "The locals of the current class", we could do
```
WholeConfig.ServiceConfigType = get_service_config_union_type(config["services"])
# instead of  ServiceConfigType = get_service_config_union_type(config["services"])
```

bugwarrior/config/schema.py
```diff
@@ -387,3 +404,178 @@ def deprecate_project_name(cls, value):
...
+class Config(SchemaBase):
+    service_configs: list[ServiceConfig]
+    main: MainSectionConfig
+
+    @cached_property
+    def services(self) -> "list[Service]":
+        return [
+            service_config.service_class(service_config, self.main)
+            for service_config in self.service_configs
+        ]
```
ryneeverett: I don't think I'd noticed this before, but why do we want to instantiate our service classes here? IMO collect.py is the sensible place for this. (I realize this is applicable to WholeConfig as well.)

ryneeverett:
If we did go with the two-model idea, it seems like we'd want to build the "view" out of a generic dataclass rather than a pydantic model. Otherwise, it seems like we'd have to feed the raw data into both models and validate/parse them both which seems wrong and also opens up the theoretical possibility of "passing" the validation stage and then failing the view-model validation stage. I suppose we'd only need to define the top-level dataclass though and could simply assign the top-level fields we want from the pydantic model, right?

Lotram:
>If we did go with the two-model idea, it seems like we'd want to build the "view" out of a generic dataclass rather than a pydantic model. Otherwise, it seems like we'd have to feed the raw data into both models and validate/parse them both which seems wrong and also opens up the theoretical possibility of "passing" the validation stage and then failing the view-model validation stage. 

I'm not sure I understand what you're saying here. While we can use a dataclass, I don't see why we can't use another pydantic model, by doing something like:
```
class ConfigView(SchemaBase):
    service_configs: list[ServiceConfig]
    main: MainSectionConfig


class WholeConfig(SchemaBase):
    ... 
    def to_config_view(self):
        return ConfigView(
            service_configs=self.service_configs,
            main=self.main,
            hooks=self.hooks,
            notifications=self.notifications,
        )
```
There is no additional validation, except for the types, and no parsing.

I'll work on a final version this week-end.

ryneeverett:
I see. I hadn't realized that pydantic models accept instantiated submodels as inputs just as well as the data to instantiate those submodels. But that does seem to be the case:

```py
>>> from typing import List, Optional
...
... from pydantic import BaseModel
...
...
... class Foo(BaseModel):
...     count: int
...     size: Optional[float] = None
...
...
... class Bar(BaseModel):
...     apple: str = 'x'
...     banana: str = 'y'
...
...
... class Spam(BaseModel):
...     foo: Foo
...     bars: List[Bar]
...
>>> foo = Foo(count='4')
>>> n = Spam(foo=foo, bars=[{'apple': 'x1'}, {'apple': 'x2'}])
>>> m = Spam(foo={'count': 4}, bars=[{'apple': 'x1'}, {'apple': 'x2'}])
>>> n == m
True
```

tests/config/test_load.py
```diff
@@ -185,14 +169,15 @@ def test_ini_invalid(self):
 
     def test_toml_flavors(self):
         config_path = self.create('.bugwarrior.toml')
-        for section in ('["flavor.myflavor"]', '[flavor.myflavor]'):
```
Lotram, line 188: IMO, we should not have to  handle the `["flavor.myflavor"]` case. It's explicitly another syntax, the user wants to have a key that is not `[flavor.myflavor]`, I think it should be ok to raise an error here.

bugwarrior/services/jira.py
```diff
@@ -337,6 +337,10 @@ class JiraService(Service):
     CONFIG_SCHEMA = JiraConfig
 
     def __init__(self, *args, **kw):
+        # NOTE: this is only used in tests.
```
Lotram: Not really related to this MR, I'll remove this.

bugwarrior/config/load.py
```diff
@@ -52,58 +53,89 @@ def get_config_path():
...
+def parse_ini_file(configpath: str) -> dict:
+    rawconfig = BugwarriorConfigParser()
+    with open(configpath, encoding="utf-8") as buff:
+        rawconfig.read_file(buff)
+
+    config = {"flavor": {}}
+    for section in rawconfig.sections():
+        if section in ['hooks', 'notifications']:
+            config[section] = dict(rawconfig[section])
+        elif section == 'general' or section.startswith('flavor.'):
```
Lotram, line 82: This used to be 2 distinct `elif`, with `k.replace('log.', 'log_')` for `general`, and `k.replace('.', '_')` for other flavors. I couldn't see a reason for that, maybe I'm missing something

bugwarrior/services/jira.py
ryneeverett: It doesn't block a minor improvement (in a separate MR), but I would think that mocking out jira (#1134) would eliminate the need for this mechanism of avoiding real network access.

tests/config/test_load.py
ryneeverett: Ok, that seems reasonable. But if we're going to eliminate this test can we add a test that `["flavor.myflavor"]` gives an appropriate error message?

ryneeverett:
Breaking validation out of schema.py seems good. I'm thinking Config should go with it since it's not part of the configuration schema per se and load.py already imports it from validation.py

Lotram:
@ryneeverett I squashed most of my commits, and applied your comments.

bugwarrior/config/load.py
```diff
@@ -52,58 +53,87 @@ def get_config_path():
     return paths[0]
 
 
+def format_config(config: dict) -> dict[str, Any]:
+    # Build flavors from 'flavor' table
+    flavors = {
+        name: flavor_config for name, flavor_config in config.pop("flavor", {}).items()
+    }
```
ryneeverett:
```suggestion
    flavors = config.pop("flavor", {})
```
Or maybe just leave this key called "flavor"? Am I missing something here?

bugwarrior/config/load.py
```diff
@@ -52,58 +53,87 @@ def get_config_path():
...
+
+    # Handle 'general' as top-level key (TOML format, tests)
```
ryneeverett: "(TOML format, tests)" ?

bugwarrior/config/load.py
```diff
@@ -52,58 +53,87 @@ def get_config_path():
...
+    if "general" in config:
+        flavors["general"] = config.pop("general")
```
ryneeverett: I'm believe 'general' was a required field, the lack of which would have thrown an error before. I think it's ok if this no longer throws an error, but have we checked the behavior if no flavor is specified and no 'general' section exists?

bugwarrior/config/load.py
```diff
@@ -1,14 +1,15 @@
...
+from pathlib import Path
+from typing import Any
 
 try:
     import tomllib  # python>=3.11
 except ImportError:
     import tomli as tomllib  # backport
 
-from . import schema
+from bugwarrior.config.validation import Config, validate_config
```
ryneeverett: For consistency with the rest of the `bugwarrior.config` package.

```suggestion
from .validation import Config, validate_config
```

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,200 @@
...
+
+    main = flavors[main_section]
+    filtered_service_configs = [service_configs[target] for target in main.targets]
+
+    return Config(
+        service_configs=filtered_service_configs,
+        main=main,
+        hooks=hooks,
+        notifications=notifications,
+    )
```
ryneeverett: Rather than *completely* doing validation prior to constructing the model, couldn't we put the Config instantiation in a try/except block and then append any errors to error_messages? Seems like this would eliminate a lot of the repetitive code here, would avoid further repetition if we were to add more fields to Config some day, and would keep our aberration from typical pydantic usage to a minimum.

bugwarrior/config/load.py
```diff
@@ -52,58 +53,87 @@ def get_config_path():
...
+
+    # Handle 'general' as top-level key (TOML format, tests)
+    if "general" in config:
+        flavors["general"] = config.pop("general")
+
+    services = {
+        section: {**config.pop(section), "target": section}
+        for section in list(config)
+        if section not in {"hooks", "notifications"}
+    }
```
ryneeverett: Did you consider making this a list? It appears that we're never doing lookups and the only places we're using the key name we could just as easily use the "target" value.

```suggestion
    services = [
        {**config.pop(section), "target": section}
        for section in list(config)
        if section not in {"hooks", "notifications"}
    ]
```

bugwarrior/config/schema.py
```diff
@@ -312,9 +197,11 @@ class ServiceConfig(_ServiceConfig):
     .. _Pydantic: https://docs.pydantic.dev/latest/
     """
 
+    service: str
+    target: str
```
ryneeverett, line 202: An explanatory comment above this block of fields would be nice. Maybe "Added before validation (computed field)".

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,200 @@
...
+def raise_validation_error(msg, config_path, error_count=1) -> typing.NoReturn:
+    log.error(
+        ("Validation error" if error_count == 1 else f"{error_count} validation errors")
+        + f" found in {config_path}\n"
+        f"See https://bugwarrior.readthedocs.io\n\n{msg}"
+    )
+    sys.exit(1)
+
+
+def get_service_config_union_type(services: dict[str, dict[str, Any]]):
```
ryneeverett: You don't have to adopt this suggested wording, but I think the main things a doc-comment ought to cover here are:
- What is the purpose of this function? (Typing isn't usually this complicated or verbose. Why is this the exception?)
- What is the return value of this function? (I'm not sure a type annotation would be very helpful to somebody trying to understand what's going on here.)

```suggestion
def get_service_config_union_type(services: dict[str, dict[str, Any]]):
    """
    Return a Union type of the ServiceConfig subclasses of the services actually configured.
    
    We don't want to include all available services because each service must be imported which could have side-effects.
    
    The returned type takes advantaged of Pydantic's "Discriminated Unions" feature to ensure that each service configuration is validated against the ServiceConfig corresponding to the correct service.
    """
```

bugwarrior/config/validation.py
Lotram: Actually yes I thought about this, but ensuring we raise all errors, while avoiding duplicates, is actually tricky:

We could definitely avoid pre-validating hooks and notifications, but then: if there are some service-related errors, we would catch them both during pre-validation and `Config` validation (same for flavors). Avoiding this would require further checks, not sure that's better. 

I wanted to avoid mixing  type adapter and model validations, that's why I went with the full type adapter validation, then passing it to the model. I still can revert it back to using the `WholeConfig` model validation. 

I agree the validation repetition does not feel right though

bugwarrior/config/load.py
Lotram: yes, my first implementation was using lists, for both services and flavors. The main issue with this is we can't retrieve the service target  / flavor name from any validation error, that's why I switched to a dict, so that it explicitly appears in the `loc` attribute

bugwarrior/config/load.py
Lotram: No you're right, I messed up while moving this from validation.py . Simplified

bugwarrior/config/load.py
Lotram: This comment actually means the "general" section is already moved to flavor dict, for ini files. But that's confusing, I'll remove it

bugwarrior/config/validation.py
ryneeverett: As it currently stands, if errors are encountered then the values we're passing into `Config` are undefined. The only way we'd get repeated errors is if we passed the same user input into the model in these cases. Rather than doing that, couldn't we define them as generic dummy values that we know will not generate any errors? For example, in the except block after validating service configs, we could define `service_configs = []`.

*Edit: I just realized that my comment about them being undefined is untrue because you're initializing them first. I think the point stands though.*

bugwarrior/config/load.py
ryneeverett: I do see the convenience of having the target already in the `loc` attribute, but we could still get the target by passing the list into the error formatter and accessing it with the index in `loc`. A dirty demo:

```diff
-def _format_service_error(error: ErrorDetails | dict) -> str:
+def _format_service_error(error: ErrorDetails | dict, services) -> str:
     """Format validation error for service configs (discriminated union).

     loc structure:
@@ -45,7 +45,7 @@ def _format_service_error(error: ErrorDetails | dict) -> str:
     if error["type"] == "extra_forbidden":
         msg = "unrecognized option"

-    target = loc[0]
+    target = services[loc[0]]["target"]

     if len(loc) == 1:
         return _format_section_error(
@@ -136,13 +136,16 @@ def validate_config(config: dict, main_section: str, config_path: str) -> "Confi

     # Validate service configs
     ServiceConfigType = get_service_config_union_type(services_config)
-    service_configs: dict[str, ServiceConfigType] = {}
+    service_configs: list[ServiceConfigType] = []
     try:
-        service_configs = TypeAdapter(dict[str, ServiceConfigType]).validate_python(
-            services_config
+        service_configs = TypeAdapter(list[ServiceConfigType]).validate_python(
+            services_config.values()
         )
     except pydantic.ValidationError as error:
-        error_messages.extend(_format_service_error(err) for err in error.errors())
+        error_messages.extend(
+            _format_service_error(err, list(services_config.values()))
+            for err in error.errors()
+        )
```

Lotram: I tried something similar too, I found it more complex and not as readable as using a dict, don't you ? I prefer that the _format_* only rely on the errors, but if you prefer using list, I'll change it.

bugwarrior/config/validation.py
Lotram: I could pass `service_configs = []` , but `main` still needs to be defined though. What should happen if there is a `flavor` / `main_section` related error ? We won't be able to define `main`, and we'll get another error, right. 

Other alternative: we define another temp model, from fields that are neither flavors nor services (i.e. only notifications and hooks for now). Something like:

```py
   
    try:
        extra_options = ExtraConfigOptions.model_validate(config) # where config does not have either flavors nor service keys anymore.
    except pydantic.ValidationError as error:
        error_messages.append(error.errors()) # some extra formatting required here 

    if error_messages:
        raise_validation_error(
            "".join(error_messages), config_path, error_count=len(error_messages)
        )

    main = flavors[main_section]
    filtered_service_configs = [service_configs[target] for target in main.targets]

    return Config(
        service_configs=filtered_service_configs, main=main, extra_options=extra_options
    )

class ExtraConfigOptions(BaseConfig):
    hooks: Hooks = Hooks()
    notifications: Notifications = Notifications()


class Config(BaseConfig):
    service_configs: list[ServiceConfig]
    main: MainSectionConfig
    extra_options: ExtraConfigOptions

```

We could also do the same as this, but with `Config` inheriting from `ExtraConfigOptions`, but I think it makes more sense to actually have these as a field.

bugwarrior/config/load.py
ryneeverett: The only line I'm aware of which might be less readable is `target = services[loc[0]]["target"]`. My thought is that it would make the rest of the module more readable by avoiding three `dict[str,`'s, particularly in the `get_service_config_union_type` signature since that's arguably the most difficult function in the module to understand. However, this is truly not a big deal. Your call.

bugwarrior/config/load.py
```diff
@@ -52,58 +53,80 @@ def get_config_path():
...
+    services = [
+        {**config.pop(section), "target": section}
+        for section in list(config)
+        if section not in {"hooks", "notifications", "flavor"}
+    ]
+
+    return {
+        **config,  # flavor, hooks, notifications (if present)
+        "services": services,
+    }
```
ryneeverett:
```suggestion
    config["services"] = [
        {**config.pop(section), "target": section}
        for section in list(config)
        if section not in {"hooks", "notifications", "flavor"}
    ]

    return config
```

bugwarrior/config/schema.py
```diff
@@ -312,9 +197,12 @@ class ServiceConfig(_ServiceConfig):
     .. _Pydantic: https://docs.pydantic.dev/latest/
     """
 
+    service: str
+    # Added before validation (computed field)
```
ryneeverett, line 200: Shouldn't comment be above `service` since it's applicable to that as well?

bugwarrior/config/validation.py
ryneeverett: > I could pass `service_configs = []` , but `main` still needs to be defined though. What should happen if there is a `flavor` / `main_section` related error ? We won't be able to define `main`, and we'll get another error, right.

Similarly we could create an inherently valid value -- `main = MainSectionConfig(targets=[])`.

ryneeverett (commented):
FYI, I just happened to check in on this PR to see if I dropped the ball. Somehow I got zero notifications of your activity this week.

I'm not very enthusiastic about `ExtraOptions`. The extra namespace and structure doesn't have any meaning beyond the validation implementation. Moreover, I still think that using the errors from the `Config` construction has virtues beyond eliminating repetition:

- It's the most typical way to use pydantic.
- More future-proofed in that we'll never have an issue where we somehow accidentally pass invalid data into the Config constructor and then end up with a second phase of validation errors that aren't properly formatted.

ryneeverett:
@Lotram No pressure if you've been busy but I wanted to ping you in case github notifications on this PR are also broken for you.

Lotram:
Sorry I was quite busy the past few weeks, I'll definitely try to finish this before the end of the week (probably during the week end)

Lotram:
@ryneeverett I removed `ExtraOptions`. I'm not a big fan of having to define an "empty" config to validate hooks and notifications, but I have nothing better to suggest, so let's keep that for now, and we can create a new PR if we find a better solution

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
+import logging
+import sys
+import typing
+from typing import TYPE_CHECKING, Annotated, Any, Union
```
ryneeverett:
```suggestion
from typing import TYPE_CHECKING, Annotated, Any, Union, NoReturn
```

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+
+import pydantic
+from pydantic import Field, TypeAdapter
```
ryneeverett:
```suggestion
from pydantic import Field, TypeAdapter, ValidationError
```

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def _format_section_error(section: str | int, msg: str) -> str:
+    return f"[{section}]  <- {msg}\n"
+
+
+def _format_field_error(
+    section: str | int, field: str | int, msg: str, error: ErrorDetails | dict
```
ryneeverett: The section and field can be ints?

```py
>>> tomllib.loads('[1]\n2 = 3')
{'1': {'2': 3}}
```

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+from .schema import BaseConfig, Hooks, MainSectionConfig, Notifications, ServiceConfig
+
+if TYPE_CHECKING:
+    ServiceConfigType = ServiceConfig
+
+
+log = logging.getLogger(__name__)
+
+
+def _format_section_error(section: str | int, msg: str) -> str:
```
ryneeverett: The section can be int?

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+    - (index, service_type, '__root__') - model validator
+    - (index, service_type, field) - field error
+    """
+    loc = error["loc"]
+    msg = error["msg"]
+    if error["type"] == "extra_forbidden":
+        msg = "unrecognized option"
+
+    index = int(loc[0])
+    target = services[index].get("target", f"service[{index}]")
```
ryneeverett: Under what circumstances would the `f"service[{index}]"` fallback be used?

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+    formatted = f"[{section}]\n{field}"
+    if error["type"] != "missing":
+        formatted = f"{formatted} = {error['input']}"
+    return f"{formatted}  <- {msg}\n"
+
+
+def _format_service_error(
+    error: ErrorDetails | dict, services: list[dict[str, Any]]
+) -> str:
+    """Format validation error for service configs (discriminated union).
```
ryneeverett: I don't understand what "(discriminated union)" is communicating here.

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+
+    loc structure:
+    - (index,) - discriminator error (missing 'service')
```
ryneeverett: I don't understand what "discriminator error" means here.

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+    loc structure:
+    - (flavor_name, '__root__') - model validator
+    - (flavor_name, field) - field error
+    """
+    loc = error["loc"]
+    msg = error["msg"]
+
+    flavor_name = loc[0]
+
+    if len(loc) == 1 or loc[-1] == "__root__":
```
ryneeverett: According to the `loc structure`, `len(loc) == 1` seems impossible.

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+    - (section_name, '__root__') - model validator
+    """
+    loc = error["loc"]
+    msg = error["msg"]
+    if error["type"] == "extra_forbidden":
+        msg = "unrecognized option"
+
+    section_name = str(loc[0])
+
+    if len(loc) == 1 or loc[-1] == "__root__":
```
ryneeverett: According to the `loc structure`, `len(loc) == 1` seems impossible.

bugwarrior/config/validation.py
```diff
@@ -0,0 +1,207 @@
...
+    raw_flavors = config.pop("flavor", {})
+    # Validate flavors
+    try:
+        flavors = TypeAdapter(dict[str, MainSectionConfig]).validate_python(raw_flavors)
+    except pydantic.ValidationError as error:
+        flavors = {}
+        error_messages.extend(_format_flavor_error(err) for err in error.errors())
+
+    # Check for misquoted flavor sections (e.g., ["flavor.myflavor"] instead of [flavor.myflavor])
+    services_list = []
```
ryneeverett: Rename `services_list -> raw_service_configs` for consistency and clarity?

bugwarrior/db.py
```diff
@@ -240,13 +246,9 @@ def run_hooks(pre_import):
...
-def synchronize(issue_generator, conf, main_section, dry_run=False):
-    main_config = conf[main_section]
-
-    targets = main_config.targets.copy()
-    services = set([conf[target].service for target in targets])
-    key_list = build_key_list(services)
-    uda_list = build_uda_config_overrides(services)
+def synchronize(issue_generator, conf: "Config", dry_run: bool = False):
+    key_list = build_key_list(conf.service_configs)
+    uda_list = build_uda_config_overrides(conf.service_configs)
```
ryneeverett: I'm not seeing the benefit of changing the input parameters on these functions from a list of strings to a list of ServiceConfig. This change motivated the transition from `targets` (a list of strings) to `successful_config_map` (a dict) just so we could pass the entire ServiceConfig into build_key_list, which then only accesses the service name.

bugwarrior/db.py
```diff
@@ -436,26 +442,25 @@ def synchronize(issue_generator, conf, main_section, dry_run=False):
...
-        keys[target] = get_service(target).ISSUE_CLASS.UNIQUE_KEY
-    return keys
-
+def build_key_list(service_configs: "Iterable[ServiceConfig]"):
+    return {
+        service_config.service: get_service(
+            service_config.service
+        ).ISSUE_CLASS.UNIQUE_KEY
+        for service_config in service_configs
+    }
```
ryneeverett, line 455: Note: There were a couple issues with this function before this PR:

- The keys of the return value are never used so this should probably return a list or set, which also realigns with the naming and `find_taskwarrior_uid` docstring. I'd prefer this **not** be done in this PR because db.synchronize is fragile and requires very close review -- I've had to revert an unusually high percentage of changes to it for causing catastrophic bugs.
- The `targets` parameter should have been named `services`.

tests/config/test_load.py
```diff
@@ -15,23 +14,6 @@
...
-            with self.subTest(rcfile=rcfile):
-                os.environ['BUGWARRIORRC'] = str(self.basedir / rcfile)
-                config = load.load_config('general', False, False)
-
-                self.assertEqual(
-                    config['flavor.myflavor'].targets,
-                    ["gitlab_config", "jira_project", "my_github"],
-                )
-
-
```
ryneeverett, line 34: i see that toml is now tested in test_validation.py but I don't see ini tested anywhere. Is there some reason this test doesn't work any more? It isn't a great place to test flavors but it seems like basically the only test of `load.load_config` and I consider it a high-level integration test.

tests/config/test_load.py
```diff
@@ -185,14 +169,14 @@ def test_ini_invalid(self):
...
-                with open(config_path, 'w') as fout:
-                    fout.write(f'{section}\ntargets = ["my_gitlab"]')
-                config = load.parse_file(config_path)
-                self.assertEqual(
-                    config, {'flavor.myflavor': {'targets': ['my_gitlab']}}
-                )
+
+        section = '[flavor.myflavor]'
+        with open(config_path, 'w') as fout:
+            fout.write(f'{section}\ntargets = ["my_gitlab"]')
```
ryneeverett:
```suggestion
        with open(config_path, 'w') as fout:
            fout.write('[flavor.myflavor]\ntargets = ["my_gitlab"]')
```

tests/config/test_validation.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def test_main_section_missing_targets_option(self):
+        del self.config['general']['targets']
+
+        self.assertValidationError("[general]\ntargets  <- Field required")
+
+    def test_target_section_missing(self):
+        del self.config['my_service']
+
+        self.assertValidationError("[general] missing targets: my_service")
```
ryneeverett: I think this is an improvement because unlike the previous message, it indicates which main section is asking for the missing target. However, I think it would be a bit more clear and consistent with the other error messages if it were more like `"[general]\ntargets = [ "my_service" ] <- No [my_service] section found"`.

tests/config/test_validation.py
```diff
@@ -0,0 +1,206 @@
...
+            '[my_service]  <- Value error, section requires one of:\n    username\n    query'
+        )
+
+    def test_no_scheme_url_validator_default(self):
+        conf = self.validate()
+        service_config = next(
+            service_config
+            for service_config in conf.service_configs
+            if service_config.target == "my_service"
+        )
```
ryneeverett: Seeing this pattern repeated 5 times -- consider adding a helper method to this class? Like

```py
def get_service(service: str):
    conf = self.validate()
    return next(service_config for service_config in conf.service_configs if service_config.target = service)
```

tests/config/test_validation.py
```diff
@@ -0,0 +1,206 @@
...
+        """Using ["flavor.myflavor"] instead of [flavor.myflavor] raises an error."""
+        self.config['flavor.myflavor'] = {'targets': ['my_service']}
+        self.assertValidationError(
+            '["flavor.myflavor"]  <- Did you mean [flavor.myflavor]?'
+        )
+
+    def test_load_and_validate_example_toml(self):
+        config_path = Path(__file__).parent / 'example-bugwarrior.toml'
+        raw_config = parse_file(str(config_path))
+        config = validation.validate_config(raw_config, 'general', str(config_path))
```
ryneeverett:
```suggestion
        formatted_config = parse_file(str(config_path))
        config = validation.validate_config(formatted_config, 'general', str(config_path))
```

tests/config/test_data.py
```diff
@@ -49,12 +50,11 @@ def setUp(self):
                 'username': 'ralphbean',
             },
         }
-        self.config = schema.validate_config(rawconfig, 'general', 'configpath')
+        formatted = format_config(rawconfig)
+        self.config = validation.validate_config(formatted, 'general', 'configpath')
```
ryneeverett: I'm just noticing that `self.config` seems to be raw config in `ConfigTest.validate` and in this class it is formatted/validated. Maybe the easiest solution is to rename this `self.validated_config`?

tests/base.py
```diff
@@ -63,10 +100,12 @@ def tearDown(self):
...
-    def validate(self):
-        self.config['general'] = self.config.get('general', {})
-        self.config['general']['interactive'] = False
-        return schema.validate_config(self.config, 'general', 'configpath')
+    def validate(self) -> validation.Config:
+        config = self.config.copy()
+        config['general'] = config.get('general', {})
+        config['general']['interactive'] = False
+        formatted_config = format_config(config)
+        return validation.validate_config(formatted_config, 'general', 'configpath')
```
ryneeverett, line 108: I started looking into this repeated pattern in the tests:

```py
    formatted = format_config(rawconfig)
    validation.validate_config(formatted, 'general', 'configpath')
```

Maybe this method should take an optional rawconfig parameter to avoid the repetition?

tests/test_clickup.py
```diff
@@ -166,11 +166,11 @@ def setUp(self):
     @property
     def service(self):
         conf = self.validate()
-        service = ClickupService(conf['myservice'], conf['general'])
+        service = get_service_instances(conf)[0]
```
ryneeverett, line 169: What would you think about adding a helper to ConfigTest to make this repeated pattern a little more readable and robust? Like:

```py
def get_service(self, target: str) -> ServiceConfig:
    service_configs = get_service_instances(conf)
    return next(service_config for service_config in service_configs if service_config.target == target)
```

Or:

```py
def get_service(self) -> ServiceConfig:
    service_configs = get_service_instances(conf)
    self.assertEqual(len(service_configs), 1)
    return service_configs[0]
```

bugwarrior/config/validation.py
Lotram: You're right, removed

bugwarrior/config/validation.py
Lotram: ~~removed~~
See other comment about typing

bugwarrior/config/validation.py
Lotram:
```
    flavor_name = loc[0]

    if len(loc) == 1 or loc[-1] == "__root__":
        return _format_section_error(flavor_name, msg)
```
Here, `flavor_name` comes from the `ErrorDetails.loc` object, which is `tuple[str | int]`. So we needs to either handle ints, or narrow its type to a string (either using an assert, or cast)

I've another branch fixing all type issues and enforcing a type-checker, so I'd like to avoid introducing any type error

bugwarrior/db.py
Lotram: Right. it used to be better when the `conf` object had a `services` property, but we decided to remove that at some point

Lotram: Regarding `successful_config_map` that's not true though, we do `service_config = successful_config_map[issue.pop('target')]` , then use `service_config` for `service_config.static_fields`

So the dict value should at least contain the `service` and `static_fields` attributes from  the `ServiceConfig` object, so we're better off keeping the whole object IMO

tests/config/test_load.py
Lotram: This test check that all the flavors are returned by `validate_config`, but that's not the case anymore, it returns only the flavor pointed by `main_section`. So I can add another test for ini, but I need to change the assertions anyway

Lotram: `test_load_and_validate_example_files` now tests both ini and toml, for multiple values of `main_section`

tests/base.py
Lotram: To be honest, I think the whole test suite could benefit from some refactoring, with a more explicit behavior, and better consistency between each service (+ using mocks, as you already said). I plan to do that at some point, but I think it should be in another MR

tests/test_clickup.py
Lotram: Same answer as above, I can definitely do that, but I think we should also have a better / more generic template for tests on each service, as we always test the same functions (+ some specific tests if needed). so it could be done in another MR

bugwarrior/config/validation.py
ryneeverett: Any reason not to narrow the type to string?

tests/config/test_load.py
ryneeverett: You could just delete the assertion -- that's how this test was originally. I just tacked on the assertion when fixing the flavors feature a couple months ago because I was too lazy to find a better way to test it. The original test was just that `load_config` doesn't throw an error. Seems to me like that 's still valuable and covers a bit more code than the validation tests.

tests/base.py
ryneeverett: Fair enough. Another idea which might interest you (or be superseded by your ideas) is #955.

tests/test_gitlab.py
```diff
@@ -522,24 +522,27 @@ def setUp(self):
             },
         }
 
+    @property
+    def conf(self):
+        return self.validate()
```
ryneeverett: Why are we introducing a conf property for gitlab's tests?

ryneeverett (approved):

Issue #905, opened 2022-04-19
gitlab: Take advantage of GraphQL api.


ryneeverett:
Idea [contributed in #624](https://github.com/ralphbean/bugwarrior/issues/624#issuecomment-463677702).

We should test it out and explore:

- Can it improve performance?
- Do all gitlab instances necessarily support this API, or do we need to maintain backwards compatibility with the REST API?
- Are hosted gitlab instances likely to exhibit similar performance differences between the GraphQL and REST API's as gitlab.com?

shymega:
As for:

>  Do all gitlab instances necessarily support this API, or do we need to maintain backwards compatibility with the REST API?

I was thinking we could do a fail gracefully approach - first try to probe the GitLab instance for GraphQL support, then fall back to the REST API? We could add an explicit option in the configuration, which would skip the probe.

For my usage of Bugwarrior, I find gitlab.com is fast, but the hosted instance I use on a day-to-day basis is very slow to get results from. I'd like to improve this, but I wasn't sure if there's any work on GraphQL support?

ryneeverett:
Nobody has ever worked on this to my knowledge.

I would say we probably don't need to worry about maintaining REST support. I believe the graphql API was relatively new when I made that comment. Looking at both the [REST](https://docs.gitlab.com/api/rest/) and [graphql](https://docs.gitlab.com/api/graphql/) api docs, they both seem to have the exact same support:

>     Tier: Free, Premium, Ultimate
>     Offering: GitLab.com, GitLab Self-Managed, GitLab Dedicated

I would also say that we don't necessarily need to switch *entirely* to the graphql api -- it's ok if we just target the slowest queries.

shymega:
Thanks. Do we have any stats or a list of slowest queries? There's a broad set of issues relating to this, and it'd be nice to have a breakdown of the slow queries.

ryneeverett:
No, we do not have any data like that. The question highlights the fact that we don't really have a method to compare performance. We've basically only gone after such low-hanging fruit that the question has never been asked. There haven't been very many performance related issues. I actually think gitlab might be the only service anyone has ever complained about and it actually used to be much slower! My recollection is that we addressed it with a combination of observation and common sense. For example, it stands to reason that decreasing the number of queries we make will improve performance.

I'm not sure there even is a truly objective way to measure performance. The bottleneck might be data transmission or server processing depending on your network and server speed.

[...]

PR #1184, opened 2026-02-11
Gitlab worker fails because index out of range

dhufe:
In my case I tried to fetch issues which are assigned to my user and I used 

```
[gitlab_instance]
service = gitlab
gitlab.only_if_assigned = USER
gitlab.owned = True
gitlab.login = USER
gitlab.token = TOKEN
gitlab.host = HOST
gitlab.import_labels_as_tags = True
gitlab.label_template = GL {{label}}
gitlab.description_template = Issue {{gitlabnumber}}: {{gitlabtitle}}
```

At the moment there aren't any assigned issues and the response contains an empty list. I added some code and updated the tests the for `only_if_assigned` behavior in GitLab service. They handle some additional edge cases for user lookup, including missing, multiple matches, and `also_unassigned`.

With theses changes I got no error anymore and if I created some issues in the production system, it runs just fine. 

If you have any further questions, feel free to contact me. 

Bests 
Daniel

### DIFF
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -150,14 +150,14 @@ class GitlabClient(Client):
             users = self._fetch(f'users?username={only_if_assigned}')
 
             if len(users) != 1:
-                log.warning(
+                log.critical(
                     "Expected exactly one user for '%s', got %d. "
-                    "Assignee filter will be ignored.",
+                    "Is only_if_assigned configured to a valid username?",
                     only_if_assigned,
                     len(users),
                 )
-            else:
-                assignee_id = users[0]["id"]
+                sys.exit(1)
+            assignee_id = users[0]["id"]
 
         self.assignee_query = f'assignee_id={assignee_id}' if assignee_id else ''
 
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -141,13 +141,17 @@ class GitlabClient(Client):
 
         self.repo_cache = {}
 
-        # If we're only fetching assigned issues we can reduce requests by
-        # filtering in the query.
-        assignee_id = (
-            self._fetch(f'users?username={only_if_assigned}')[-1]['id']
-            if only_if_assigned and not also_unassigned
-            else None
-        )
+        assignee_id = None
+        if only_if_assigned and not also_unassigned:
+            users = self._fetch(f'users?username={only_if_assigned}')
+            if not users:
+                log.warning(f"User '{only_if_assigned}' not found on GitLab instance")
+            elif len(users) > 1:
+                log.warning(f"Multiple users found for '{only_if_assigned}', using first match")
+                assignee_id = users[0]['id']
+            else:
+                assignee_id = users[0]['id']
+
         self.assignee_query = f'assignee_id={assignee_id}' if assignee_id else ''
 
     def _base_url(self):
--- a/bugwarrior/services/gitlab.py
+++ b/bugwarrior/services/gitlab.py
@@ -142,15 +142,22 @@ class GitlabClient(Client):
         self.repo_cache = {}
 
         assignee_id = None
+
         if only_if_assigned and not also_unassigned:
+            # GitLab API returns:
+            #   - a single-element list if the username exists
+            #   - an empty list if it does not
             users = self._fetch(f'users?username={only_if_assigned}')
-            if not users:
-                log.warning(f"User '{only_if_assigned}' not found on GitLab instance")
-            elif len(users) > 1:
-                log.warning(f"Multiple users found for '{only_if_assigned}', using first match")
-                assignee_id = users[0]['id']
+
+            if len(users) != 1:
+                log.warning(
+                    "Expected exactly one user for '%s', got %d. "
+                    "Assignee filter will be ignored.",
+                    only_if_assigned,
+                    len(users),
+                )
             else:
-                assignee_id = users[0]['id']
+                assignee_id = users[0]["id"]
 
         self.assignee_query = f'assignee_id={assignee_id}' if assignee_id else ''
 
--- a/tests/test_gitlab.py
+++ b/tests/test_gitlab.py
@@ -1167,16 +1167,3 @@ class TestGitlabIssue(AbstractServiceTest, ServiceTest):
             self.get_mock_service(GitlabService, config_overrides=overrides)
         self.assertEqual(cm.exception.code, 1)
 
-    @responses.activate
-    def test_only_if_assigned_with_also_unassigned(self):
-        """Test that assignee_id is None when also_unassigned is True"""
-        overrides = {'only_if_assigned': 'jack_smith', 'also_unassigned': 'true'}
-
-        # User lookup should NOT be called when also_unassigned is True
-        # So we don't add any mock response
-
-        # Should not raise an error and not call the API
-        service = self.get_mock_service(GitlabService, config_overrides=overrides)
-
-        # Verify service was created successfully
-        self.assertIsNotNone(service)
--- a/tests/test_gitlab.py
+++ b/tests/test_gitlab.py
@@ -1166,4 +1166,3 @@ class TestGitlabIssue(AbstractServiceTest, ServiceTest):
         with self.assertRaises(SystemExit) as cm:
             self.get_mock_service(GitlabService, config_overrides=overrides)
         self.assertEqual(cm.exception.code, 1)
-
--- a/tests/test_gitlab.py
+++ b/tests/test_gitlab.py
@@ -1098,3 +1098,107 @@ class TestGitlabIssue(AbstractServiceTest, ServiceTest):
         }
 
         self.assertEqual(TaskConstructor(issue).get_taskwarrior_record(), expected)
+
+    @responses.activate
+    def test_only_if_assigned_user_lookup(self):
+        """Test that only_if_assigned correctly looks up the user and uses first match"""
+        # Mock the user lookup API call - WITH username in query string
+        self.add_response(
+            'https://my-git.org/api/v4/users?username=jack_smith',
+            json=[
+                {
+                    'id': 2,
+                    'username': 'jack_smith',
+                    'name': 'Jack Smith',
+                    'state': 'active'
+                }
+            ],
+        )
+
+        overrides = {
+            'only_if_assigned': 'jack_smith',
+        }
+
+        # Should not raise an error
+        service = self.get_mock_service(GitlabService, config_overrides=overrides)
+
+        # Verify service was created successfully
+        self.assertIsNotNone(service)
+
+    @responses.activate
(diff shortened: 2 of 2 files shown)


dhufe:
I fogort the traceback of the error:

```
  File "/home/dhufe/dev/bugwarrior/.venv/lib64/python3.12/site-packages/bugwarrior/collect.py", line 45, in _aggregate_issues
    service = get_service(conf[target].service)(conf[target], conf[main_section])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dhufe/dev/bugwarrior/.venv/lib64/python3.12/site-packages/bugwarrior/services/gitlab.py", line 499, in __init__
    self.gitlab_client = GitlabClient(
                         ^^^^^^^^^^^^^
  File "/home/dhufe/dev/bugwarrior/.venv/lib64/python3.12/site-packages/bugwarrior/services/gitlab.py", line 147, in __init__
    self._fetch(f'users?username={only_if_assigned}')[-1]['id']
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
IndexError: list index out of range
```

ryneeverett:
> At the moment there aren't any assigned issues and the response contains an empty list.

My understanding is that `self._fetch(f'users?username={only_if_assigned}')` is just looking up the username with the user api. Shouldn't the user be returned irrespective of whether any issues are assigned?

I'm sure there's some room for improvement here but I want to make sure we understand the issue first.

dhufe:
Thanks for your reply. For sure, it only tries to query the user. 

The problem is generated by [-1]['id'] and the case, that the user does not exist (or is incorrect). As far as I understand that would take the last element of a list, no matter if it's an empty one.

ryneeverett:
IMO, if the only_if_assigned username does not exist, it ought to throw an error as it currently does rather than log a warning and continue. An improvement would be to catch the IndexErrror in that case and raise an exception with a more helpful error message.

I don't think we need to handle `len(users) > 1` because [according to the docs](https://docs.gitlab.com/api/users/) that shouldn't be possible:

> Get a single user with a specific username.

And I think it is implied elsewhere in that doc that the username query should only return an "exact match".

dhufe:
> I don't think we need to handle len(users) > 1 because [according to the docs](https://docs.gitlab.com/api/users/) that shouldn't be possible:

That’s a good point, but then -1 is also wrong. It implies directly, that the response might contain multiples users. And that’s ok since the Gitlab docs say:

> Name or username, you do not have to get an exact match because this is a fuzzy search.


and therefore a response may contain multiple users.

> And I think it is implied elsewhere in that doc that the username query should only return an "exact match".

From the [bugwarrior docs](https://bugwarrior.readthedocs.io/en/latest/common_configuration.html#main-section):

> only_if_assigned: If set to a username, only import issues assigned to the specified user.

I don’t think so, but may be I rushed over the text. My first thought was: why should this an user and why Bugwarrior do not use my gitlab.login, which was already defined (by the way if you use a Token the user does not matter anymore, the same for GitHub). I understand the use case to query stuff for a different user. May be it would be handy to allow strings and boolean.

ryneeverett:
> Name or username, you do not have to get an exact match because this is a fuzzy search.

This refers to the `search` parameter, not the `username` parameter.

> My first thought was: why should this an user and why Bugwarrior do not use my gitlab.login, which was already defined (by the way if you use a Token the user does not matter anymore, the same for GitHub). I understand the use case to query stuff for a different user. May be it would be handy to allow strings and boolean.

The answer to a lot of "why does bugwarrior do x?" questions is historical. My recollection is that this feature existed before services started aggressively rate banning or disallowing unauthenticated api queries and authentication options were generally optional. And now basic user/password authentication is largely being phased out for api use so the `login` parameters are probably going to disappear eventually.

Good point about login not being needed with token authentication. Unfortunately, we're using the login to automatically namespace the 'include_repos' and 'exclude_repos' options and there doesn't seem like a great way to avoid that. It does look like it could be easily deleted from the github service though.

dhufe:
> The answer to a lot of "why does bugwarrior do x?" questions is historical. My recollection is that this feature existed before services started aggressively rate banning or disallowing unauthenticated api queries and authentication options were generally optional. And now basic user/password authentication is largely being phased out for api use so the login parameters are probably going to disappear eventually.

That wasn't my intention and this is normal and totally ok for grown projects. 

> Good point about login not being needed with token authentication. Unfortunately, we're using the login to automatically namespace the 'include_repos' and 'exclude_repos' options and there doesn't seem like a great way to avoid that. It does look like it could be easily deleted from the github service though.

I suggest keeping the current logic unchanged in this PR and only improving error handling. As a next step, I would:

add proper exception handling with meaningful log output  
explicitly check the response length (0 → error, 1 → success)  
document that GitLab returns at most one user for this query (verified against multiple instances)

Bests
Daniel

ryneeverett:
> add proper exception handling with meaningful log output

:+1: 

> explicitly check the response length (0 → error, 1 → success)

What advantage does this have over catching the IndexError?

> document that GitLab returns at most one user for this query (verified against multiple instances)

I don't understand what this means.

dhufe:
> > add proper exception handling with meaningful log output
> 
> 👍
> 
> > explicitly check the response length (0 → error, 1 → success)
> 
> What advantage does this have over catching the IndexError?

Using IndexError here would technically handle the empty case, but it hides the actual assumption we are making — namely, that the API must return exactly one user.  
An explicit length check makes this expectation visible in the code and avoids using an exception as part of normal control flow. I believe this improves readability and long-term maintainability.

> 
> > document that GitLab returns at most one user for this query (verified against multiple instances)
> 
> I don't understand what this means.

What I meant is simply documenting in the code that this GitLab endpoint returns a list that either:

contains exactly one element, or  
is empty.

This would clarify the expected behavior and hopefully avoid further discussions like this in the future.

Best regards,
Daniel

dhufe:
Ok, I updated the code: 

Handle unresolved usernames gracefully by logging a warning instead of raising an error.  
When the GitLab API does not return exactly one user, the assignee filter is intentionally ignored and the import continues unchanged.

Let me know if this addresses your feedback.

bugwarrior/services/gitlab.py
```diff
@@ -141,13 +141,24 @@ def __init__(
...
+            #   - an empty list if it does not
+            users = self._fetch(f'users?username={only_if_assigned}')
+
+            if len(users) != 1:
+                log.warning(
+                    "Expected exactly one user for '%s', got %d. "
+                    "Assignee filter will be ignored.",
+                    only_if_assigned,
+                    len(users),
+                )
```
ryneeverett: We can't allow the service to succeed because that would add tasks to the database which are not assigned which is not what the user wants.

```suggestion
                log.critical(
                    "Expected exactly one user for '%s', got %d. "
                    "Assignee filter will be ignored.",
                    only_if_assigned,
                    len(users),
                )
                sys.exit(1)
```

bugwarrior/services/gitlab.py
```diff
@@ -141,13 +141,24 @@ def __init__(
...
+            else:
+                assignee_id = users[0]["id"]
```
ryneeverett: - We can't allow the service to succeed because that would add tasks to the database which are not assigned which is not what the user wants.
- We should give a hint as to the solution to the most likely scenario -- user misconfigured an invalid username resulting in zero elements.

```suggestion
                log.critical(
                    "Expected exactly one user for '%s', got %d. "
                    "Is only_if_assigned configured to a valid username?",
                    only_if_assigned,
                    len(users),
                )
                sys.exit(1)
            assignee_id = users[0]["id"]
```

ryneeverett (changes requested):

bugwarrior/services/gitlab.py
dhufe: I’m a bit hesitant about using sys.exit(1) here. Would it make sense to raise an exception instead and handle it at the point where the client is created?  
My concern is that a hard exit would terminate the entire process, potentially affecting other services as well, whereas an exception would allow more fine‑grained error handling.

ryneeverett: Exceptions, including the `SystemExit` that `sys.exit` raises, *are* handled at the point where the client is created. The result is that the given service is aborted but other services (which are separate processes) can complete. From bugwarrior/collect.py:

```python
def _aggregate_issues(conf, main_section, target, queue):
    """This worker function is separated out from the main
    :func:`aggregate_issues` func only so that we can use multiprocessing
    on it for speed reasons.
    """

    start = time.time()

    try:
        service = get_service(conf[target].service)(conf[target], conf[main_section])
        issue_count = 0
            queue.put(issue)
            issue_count += 1
    except SystemExit as e:
        log.critical(f"Worker for [{target}] exited: {e}")
        queue.put((SERVICE_FINISHED_ERROR, target))
    except BaseException as e:
        if hasattr(e, 'request') and e.request:
            # Exceptions raised by requests library have the HTTP request
            # object stored as attribute. The request can have hooks attached
            # to it, and we need to remove them, as there can be unpickleable
            # methods. There is no one left to call these hooks anyway.
            e.request.hooks = {}
        log.exception(f"Worker for [{target}] failed: {e}")
        queue.put((SERVICE_FINISHED_ERROR, target))
    else:
        log.debug(f"Worker for [{target}] finished ok.")
        queue.put((SERVICE_FINISHED_OK, target))
    finally:
        duration = time.time() - start
        log.info(f"Done with [{target}] in {duration}.")
```

dhufe:
Applied your suggested change and updated the tests accordingly.

tests/test_gitlab.py
```diff
@@ -1098,3 +1098,85 @@ def test_issues(self):
...
+        overrides = {'only_if_assigned': 'jack_smith', 'also_unassigned': 'true'}
+
+        # User lookup should NOT be called when also_unassigned is True
+        # So we don't add any mock response
+
+        # Should not raise an error and not call the API
+        service = self.get_mock_service(GitlabService, config_overrides=overrides)
+
+        # Verify service was created successfully
+        self.assertIsNotNone(service)
```
ryneeverett: The doc-comment doesn't seem aligned with what this test actually does, but it also isn't clear to me that this final test has any value. Seems like it just tests that using both only_if_assigned and also_unassigned doesn't cause an error but that doesn't seem like a real concern.

I don't see any reason to test that it doesn't call the API -- that's an implementation detail and we don't really want to test implementation details because those change. For example, this seems like the kind of situation where the graphql api (#905) might allow us to fetch just the assigned and unassigned issues rather than fetching them all and filtering later. In that case we'd probably still need to fetch the user's id here.

ryneeverett (approved):

ryneeverett:
I'm guessing CI failed because of the extra newline left at the end of test_gilab.py. Could you remove it or run `ruff format`?

ryneeverett (approved):

PR #1150, opened 2025-11-24
fix: require setuptools for pipx/uvx

jochumdev:
Needed to make an install with uvx as seamless as possible.

This command installs bugwarrior without this PR on any platform uv supports:

```sh
uvx -q --with setuptools git+https://github.com/GothenburgBitFactory/bugwarrior
```

After a merge it should be possible to do:

```sh
uvx git+https://github.com/GothenburgBitFactory/bugwarrior
```

P.s.: sorry for the noise in #1149, changed my tooling.

PR #1189, opened 2026-03-19
Add ty as type checker

Lotram:
Following https://github.com/GothenburgBitFactory/bugwarrior/issues/1151, this add `ty` as a type checker, and fix all related errors.

Some notes:
* dev dependencies are now defined in a [dependency group](https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups), to avoid including them in the project when it's built
* I added a CI job for that, as well as for ruff (linter + formatter). `ruff` used to be checked through a test (removed to so tests don't require having ruff anymore), which is not really standard IMO, but I can revert this if needed. 
* I migrated all CI jobs to use uv instead of pip, for better consistency. 
* tests and docs are excluded from type checking. This can be changed in the future, but I don't see a good reason to enforce it for now
* I'll introduce new ruff rules (some [ANN](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann) rules) to ensure all functions are properly typed, in another MR


bugwarrior/services/teamwork_projects.py
```diff
@@ -54,12 +54,6 @@ class TeamworkIssue(Issue):
     UNIQUE_KEY = (URL,)
     PRIORITY_MAP = {"low": "L", "medium": "M", "high": "H"}
 
-    def get_owner(self):
```
Lotram, line 57: This method seemed broken, and unused (except in the tests), so I removed it

bugwarrior/services/trac.py
```diff
@@ -105,7 +107,9 @@ def annotations(self, issue):
         # without offtrac, we can't get issue comments
         if self.trac is None:
             return annotations
-        changelog = self.trac.server.ticket.changeLog(issue['number'])
+        changelog = typing.cast(
```
Lotram, line 108: I think we could either use `typing.cast` or an `assert`, I don't have any strong opinion about this

.github/workflows/bugwarrior.yml
```diff
@@ -64,32 +56,26 @@ jobs:
...
             source .venv/bin/activate
-            pip install --upgrade pip setuptools
+            pip install --upgrade "pip>=25.1" setuptools
 
             pip install -e .[all]
+            pip install --group test
 
             bugwarrior --version
             pytest
-  bugwarrior-development-uv-test:
```
Lotram, line 73: removed this since we now use uv in the main job

.readthedocs.yaml
```diff
@@ -10,11 +11,12 @@ build:
...
     python: "3.10"
-
-# Optionally declare the Python requirements required to build your docs
-python:
-   install:
-      - method: pip
-        path: .
-        extra_requirements:
-           - all
+  jobs:
```
Lotram, line 14: this file needed to be updated to use dependency groups, though if we want to keep pip, [that's also possible](https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-from-dependency-groups)

.github/workflows/bugwarrior.yml
```diff
@@ -64,32 +56,26 @@ jobs:
             mkdir .venv
             python3 -m venv .venv
             source .venv/bin/activate
-            pip install --upgrade pip setuptools
+            pip install --upgrade "pip>=25.1" setuptools
```
Lotram, line 83: ensure pip >= 25.1 to support the --group flag

bugwarrior/config/ini2toml_plugin.py
```diff
@@ -131,6 +131,6 @@ def unquote_flavors(file_contents: str) -> str:
 
 def activate(translator: Translator):
     profile = translator["bugwarriorrc"]
-    profile.description = "Convert 'bugwarriorrc' files to 'bugwarrior.toml'"
+    profile.help_text = "Convert 'bugwarriorrc' files to 'bugwarrior.toml'"
```
ryneeverett: Do you know what the difference is between these attributes? Ini2toml docs seem unclear.

bugwarrior/services/__init__.py
```diff
@@ -358,7 +359,7 @@ def issues(self):
 
     @staticmethod
     @abc.abstractmethod
-    def get_keyring_service(service_config) -> str:
+    def get_keyring_service(config) -> str:
```
ryneeverett:
```suggestion
    def get_keyring_service(config: schema.ServiceConfig) -> str:
```

bugwarrior/services/bts.py
```diff
@@ -94,9 +94,9 @@ def get_default_description(self):
         )
 
     def get_priority(self):
-        return self.PRIORITY_MAP.get(
-            self.record.get('severity'), self.config.default_priority
-        )
+        if "severity" not in self.record:
+            return self.config.default_priority
+        return self.PRIORITY_MAP.get(self.record['severity'])
```
ryneeverett: This seems like a change in behavior. If `self.record["severity"] not in self.PRIORITY_MAP` before it would return `self.config.default_priority` but now it returns `None`.

bugwarrior/services/gitbug.py
```diff
@@ -32,19 +32,16 @@ def __init__(self, path, port):
...
-        try:  # Windows
-            kwargs = {'creationflags': (subprocess.CREATE_NEW_PROCESS_GROUP,)}
-        except AttributeError:  # Posix
-            self.windows = False
-            kwargs = {'start_new_session': True}
-
+        if sys.platform == "win32":
+            popen_kwargs = {"creationflags": subprocess.CREATE_NEW_PROCESS_GROUP}
+        else:
+            popen_kwargs: dict[str, typing.Any] = {"start_new_session": True}
```
ryneeverett: Seems weird (and ugly) that this typehint is only needed conditionally.

bugwarrior/services/gitlab.py
```diff
@@ -610,8 +611,8 @@ def _get_todo_objs(self, todos):
             todo_obj = self.get_issue_for_record(todo)
             todo_url = todo['target_url']
             project_name = repo['path']
-            if self.config.project_owner_prefix:
-                project_name = repo['namespace']['path'] + "." + project_name
+            if self.config.project_owner_prefix and project is not None:
+                project_name = project['namespace']['path'] + "." + project_name
```
ryneeverett, line 611: Can we just eliminate the `repo` variable?

- `todo['repo'] = project['path'] if project is not None else 'the instance'`
- `project_name = todo['repo']`

bugwarrior/services/phab.py
```diff
@@ -71,10 +71,9 @@ def get_default_description(self):
 
     @property
     def priority(self):
-        return (
-            self.PRIORITY_MAP.get(self.record.get('priority'))
-            or self.config.default_priority
-        )
+        if "severity" not in self.record:
+            return self.config.default_priority
+        return self.PRIORITY_MAP.get(self.record['severity'])
```
ryneeverett: This seems like a change in behavior. If `self.record["severity"] not in self.PRIORITY_MAP` before it would return `self.config.default_priority` but now it returns `None`.

pyproject.toml
```diff
@@ -112,6 +98,37 @@ max-line-length = 100
...
+[dependency-groups]
+docs = [
+    # sphinx-inline-tabs is incompatible with docutils-0.22 https://github.com/pradyunsg/sphinx-inline-tabs/pull/51
+    "docutils<0.22",
+    "sphinx>=1.0,<9.0 ; python_version <= '3.10'",
+    "sphinx>=9.0.3 ; python_version >= '3.11'",
+    "sphinx-click",
+    "sphinx-inline-tabs",
+    # To make this installable with uvx/pipx and no additional options needed.
+    "setuptools",
```
ryneeverett: I suspect setuptools needs to stay in optional-dependencies -- it doesn't make much sense here -- but I don't really understand why #1150 worked. Do you?

.github/workflows/bugwarrior.yml
ryneeverett: Maybe note this constraint in the pip tab of contributing.rst?

ryneeverett (commented):
>     * dev dependencies are now defined in a [dependency group](https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups), to avoid including them in the project when it's built

Fine, but it seems like we need to update contributing.rst to include the `group --dev` for uv and whatever pip needs to adapt.
      
>     * I added a CI job for that, as well as for ruff (linter + formatter). `ruff` used to be checked through a test (removed to so tests don't require having ruff anymore), which is not really standard IMO, but I can revert this if needed.

What is the problem with tests requiring ruff and ty? I know it's not standard practice, but from my perspective the more we have in tests and the less in CI the better because (a) less distinction between local/CI to debug (b) the tests are portable and extremely stable whereas CI configuration is vendored and less stable.

>     * I migrated all CI jobs to use uv instead of pip, for better consistency.

That's fine, but could we copy the old `bugwarrior-test` to `bugwarrior-development-pip-test` so we're still testing basic pip functionality (since it's documented as an option). This is in the spirit of what `bugwarrior-development-uv-test` was doing before, but for pip.

bugwarrior/services/gitbug.py
Lotram: Agreed. I moved the typing before the if condition.

.github/workflows/bugwarrior.yml
```diff
@@ -64,32 +80,26 @@ jobs:
...
-          uv run bugwarrior --version
-          uv run pytest
+      - name: Install dependencies
+        run: uv sync --all-extras --group test
+      - name: Ruff check
+        run: uv run ruff check
+      - name: Ruff format
+        run: uv run ruff format --check
+      - name: Type check
+        run: uv run ty check
```
ryneeverett: Do you think it makes sense to run linters in a separate job like this when they're already run in the test suite?

bugwarrior/docs/contributing.rst
```diff
@@ -61,10 +65,11 @@ To run the tests, use ``pytest``:
...
 
-We use ruff for linting and formatting. This is checked in the test suite, so
-you don't necessarily need to worry about it separately, but you may find it
-convenient to run ``ruff check`` and ``ruff format`` or `integrate ruff with
-your editor <https://docs.astral.sh/ruff/editors/setup/>`_.
+We use ruff for linting and formatting and ty for type checking. These are
+checked in the test suite, so you don't necessarily need to worry about them
+separately, but you may find it convenient to run ``ruff check``, ``ruff
+format``, and ``ty check``, or `integrate ruff with your editor
+<https://docs.astral.sh/ruff/editors/setup/>`_.
```
ryneeverett: Can we link to ty's editor integration page too? https://docs.astral.sh/ty/editors/

ryneeverett (approved):

Issue #1194, opened 2026-04-01
type checker is failing in CI


Lotram:
So I investigated why the ty test is failing on the CI. 

It turns out I was using the wrong ignore comments, maybe that changed recently (see [the doc](https://docs.astral.sh/ty/suppression/#standard-suppression-comments)):
I should have used either `type: ignore[ty:unresolved-import]`  or `ty: ignore[unresolved-import]`, but `type: ignore[unresolved-import]` is not supposed to work. 

I also have a specific warning with python 3.10:
```
warning[unused-type-ignore-comment]: Unused `type: ignore` directive
  --> bugwarrior/config/load.py:11:30
   |
 9 |     import tomllib
10 | else:
11 |     import tomli as tomllib  # type: ignore[ty:unresolved-import]
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
12 |
13 | from .validation import Config, validate_config
```
Because in python 3.10, that import does exist, so the ignore comment is useless, and we have a setting to prevent this.

We have several options for this:
* Remove the "unused-type-ignore-comment" rule (I'm not in favor of this one)
* ignore the ignore comment: `import tomli as tomllib  # type: ignore[ty:unresolved-import]  # ty: ignore[unused-ignore-comment]` (that [does work](https://docs.astral.sh/ty/suppression/#unused-suppression-comments))
* run ty (as well as ruff) only on a specific python version (e.g. the latest we're supporting). I'm not sure it useful to run these tests on each supported version.

You should be able to reproduce with `uv sync --all-extras --group test --python 3.10 && uv run ty check`

Even though it does not seem directly related to our problem, I can also pin ty to a specific version, to avoid CI randomly starting to fail.

Lotram:
On a related topic (as I saw while working on the unresolved import errors): 
In notifications.py, there are 2 optional imports: `gntp` and `gi`. Shouldn't they be declared as optional deps in pyproject.toml ? Though, as [gntp](https://github.com/kfdm/gntp/) has been archived for the past 5 years (last commit 10 years ago), we could also consider removing it entirely.

ryneeverett:
> I'm not sure it useful to run these tests on each supported version.

I think it is preferable to run on all supported versions because that way we support bugwarrior developers using any of those versions in their development environment, but we shouldn't jump through hoops if it is inconvenient. Is another option to just exclude python-3.10 by decorating `test_ty_check` with a `skipIf`?

[...]

PR #1195, opened 2026-04-07
fix: now use proper ty comments

Lotram:
`ty` use `ty:` comments, and not `type:` comments, so the ignore comments were not used. That's now fixed.

https://github.com/GothenburgBitFactory/bugwarrior/issues/1194


ryneeverett (approved):
Thanks!

PR #1196, opened 2026-04-07
linear: cursor pagination + set entry from createdAt

lxcode:
Two related fixes for the Linear service.

## 1. Pagination

The Linear GraphQL service issues a single query without `first:` or cursor pagination, so Linear's Relay-style `issues` connection returns only its default page size (50 issues) and any remaining matches are silently dropped. Workspaces with more than 50 matching issues miss the overflow on every pull.

The fix requests `first: 250` (Linear's documented maximum) and follows `pageInfo.endCursor` until `hasNextPage` is false. `get_issues` becomes a generator that streams pages, so memory stays flat regardless of total issue count. The `issues()` caller already iterated lazily, so no caller updates are needed.

A new test (`test_issues_paginates`) registers two paginated mock responses and asserts both pages' issues are returned and that the second request carries the cursor returned by the first.

## 2. Set `entry` from Linear `createdAt`

Without this, freshly imported Linear issues have `entry` set to the local pull time, so `entry.age` and the urgency age coefficient see every imported issue as brand-new regardless of its real upstream age. Mirrors the pattern already used by several other services.

The existing `test_to_taskwarrior` and `test_issues` dict-equality assertions are updated to expect the new key.

### DIFF
--- a/bugwarrior/services/linear.py
+++ b/bugwarrior/services/linear.py
@@ -149,8 +149,8 @@ class LinearService(Service, Client):
             )
 
         self.query = """
-            query Issues($filter: IssueFilter!) {
-              issues(filter: $filter) {
+            query Issues($filter: IssueFilter!, $after: String) {
+              issues(filter: $filter, first: 250, after: $after) {
                 nodes {
                   url
                   title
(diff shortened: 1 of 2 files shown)


ryneeverett (approved):

djmitche (approved):
Makes sense to me!

PR #1197, opened 2026-04-07
Add ruff rules to enforce typing annotations

Lotram:
Add ruff rules [ANN](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann) to ensure all functions are annotated, improving typing coverage. 

First commit enforces this everywhere except `services/` ( + `docs/` and `tests/`, already excluded from type checking).

Second commit enforces it in `services/__init__.py` + 2 services, as an example

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -4,13 +4,13 @@ Service API
 """
 
 import abc
-from collections.abc import Iterable
+from collections.abc import Iterable, Iterator
 import datetime
 import logging
 import math
 import os
 import re
-import typing
+from typing import Any, Generic, Optional, TypeVar
 import zoneinfo
 
 from dateutil.parser import parse as parse_date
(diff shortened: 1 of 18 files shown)


ryneeverett:
Does annotating all functions necessarily improve typing coverage (as enforced by ty)? Doesn't ty complain if it can't infer a type?

Lotram:
>  Doesn't ty complain if it can't infer a type?

Not really, all missing annotations are replaced by `Unknown`, an implicit equivalent of `Any` (see [ty doc](https://docs.astral.sh/ty/reference/typing-faq/#why-doesnt-ty-warn-about-missing-type-annotations))

See for example this piece of code:
```
def f(a):  # expects str
    a.splitlines()


def g():
    b = 5
    return f(b)
```
without type annotation, `ty` (as well as `pyright` and `mypy`) won't find any issue. 

whereas
```
def f(a: str) -> list[str]:  # expects str
    return a.splitlines()


def g():
    b = 5
    return f(b)
```
raises an error with both `ty` and `pyright` (strangely, `mypy` does not raise anything with the default config, I haven't investigated)


All type checkers require type annotations to be fully working. Both `mypy` and `pyright` have settings to enforce this, Astral uses `ruff` rules for that

bugwarrior/services/__init__.py
```diff
@@ -373,20 +378,14 @@ class Client:
...
-                "Non-200 status code %r; %r; %r"
-                % (response.status_code, response.url, response.text)
+                f"Non-200 status code {response.status_code}; {response.url}; {response.text}"
             )
-        if callable(response.json):
-            # Newer python-requests
-            return response.json()
-        else:
-            # Older python-requests
-            return response.json
```
ryneeverett, line 389: Can we constrain `requests>=1` in pyproject.toml? (based on first item: https://requests.readthedocs.io/en/stable/api/#migrating-to-1-x)

ryneeverett (changes requested):

bugwarrior/services/__init__.py
Lotram: requests 1.0.0 was released in 2012, so I doubt anyone is using earlier versions, but sure, I can add the constraint

ryneeverett (approved):

PR #1198, opened 2026-04-08
linear: map issue priority onto taskwarrior priority

lxcode:
Linear's GraphQL exposes a per-issue priority integer (0=No priority, 1=Urgent, 2=High, 3=Medium, 4=Low) that wasn't being queried, so every imported task got the service-wide default_priority. Add the field to the issues query and translate it onto taskwarrior's H/M/L buckets, falling back to default_priority for "No priority".

### DIFF
--- a/bugwarrior/services/linear.py
+++ b/bugwarrior/services/linear.py
@@ -66,18 +66,7 @@ class LinearIssue(Issue):
 
     # Linear exposes issue priority as an integer:
     #   0 = No priority, 1 = Urgent, 2 = High, 3 = Medium, 4 = Low.
-    # Map onto taskwarrior's three priority buckets. ``None`` (used for "No
-    # priority") tells ``get_priority`` to fall back to ``default_priority``.
-    PRIORITY_MAP = {0: None, 1: "H", 2: "H", 3: "M", 4: "L"}
-
-    def get_priority(self):
-        priority = self.record.get("priority")
-        if priority is not None:
-            mapped = self.PRIORITY_MAP.get(priority)
-            if mapped is not None:
-                return mapped
-        return self.config.default_priority
-
+    PRIORITY_MAP = {1: "H", 2: "H", 3: "M", 4: "L"}
     def to_taskwarrior(self):
         description = self.record.get("description")
         created = self.parse_date(self.record.get("createdAt"))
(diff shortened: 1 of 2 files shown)


bugwarrior/services/linear.py
```diff
@@ -64,6 +64,20 @@ class LinearIssue(Issue):
...
+    PRIORITY_MAP = {0: None, 1: "H", 2: "H", 3: "M", 4: "L"}
+
+    def get_priority(self):
+        priority = self.record.get("priority")
+        if priority is not None:
+            mapped = self.PRIORITY_MAP.get(priority)
+            if mapped is not None:
+                return mapped
+        return self.config.default_priority
+
```
ryneeverett: I believe you can get identical behavior using the inherited `get_priority` method and eliminating the `0` entry. Your test passes, anyway.

```suggestion
    PRIORITY_MAP = {1: "H", 2: "H", 3: "M", 4: "L"}
```

ryneeverett (changes requested):

bugwarrior/services/linear.py
lxcode: Ope, thanks!

ryneeverett (approved):

djmitche (commented):
Apologies if I've forgotten some magic here, but I don't see `get_priority` defined.

It seems the `0` priority from linear should map to the BW default priority, rather than always being `H`. The tests suggest that is the case but I can't see where that's implemented.

BTW as a daily user of the linear integration I'm looking forward to this. It's been at the end of my `+rainyday` tasks for a long time.

ryneeverett:
> Apologies if I've forgotten some magic here, but I don't see `get_priority` defined.

It is inherited. https://github.com/GothenburgBitFactory/bugwarrior/blob/ac52c2a80ab9121f20eafb61a10a82a01c886083/bugwarrior/services/__init__.py#L169-L173

> It seems the `0` priority from linear should map to the BW default priority, rather than always being `H`. The tests suggest that is the case but I can't see where that's implemented.

Since `0` isn't in PRIORITY_MAP, `get_priority` will return the default value.

djmitche (approved):
Ah, got it. Thanks!

PR #1202, opened 2026-04-22
services: add missing type annotations

Lotram:
Add return type and argument annotations across all services to satisfy the ANN ruff rules, which are now enforced on the full bugwarrior/ package.

* fixes a latent bug in NextcloudDeckService.get_owner, which treated a NextcloudDeckIssue as a subscriptable dict (TypeError at runtime when only_if_assigned was configured), and adds a regression test.
* Remove `Client` inheritance from any `Service` subclass. Since this `Client` class only has a single staticmethod, and nothing else, we call `Client.json_responses` directly instead. Services that subclass `Client` already use it an attribute rather than a parent class, so this change adds consistency between all services
* use explicit `__init__` parameters for all `Service` subclasses (using `config` and `main_config`), as they are required anyway, for easier typing.
* gitlab and github use type aliases, to improve readability

### DIFF
--- a/bugwarrior/services/teamwork_projects.py
+++ b/bugwarrior/services/teamwork_projects.py
@@ -1,5 +1,7 @@
+from collections.abc import Iterator
 import logging
 import typing
+from typing import Any
 
 import requests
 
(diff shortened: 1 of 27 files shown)


bugwarrior/services/teamwork_projects.py
```diff
@@ -19,18 +21,12 @@ class TeamworkConfig(config.ServiceConfig):
...
 
-    def authenticate(self):
-        response = requests.get(self.host + "/authenticate.json", auth=(self.token, ""))
-        return self.json_response(response)
-
-    def call_api(self, method, endpoint, data=None):
-        response = requests.get(
-            self.host + endpoint, auth=(self.token, ""), params=data
-        )
+    def _get(self, endpoint: str) -> dict[str, Any]:
```
ryneeverett: Seems weird to make this a `_` method when it is used by another class.

pyproject.toml
```diff
@@ -115,6 +114,7 @@ include = ["bugwarrior*"]
 
 [tool.ty.environment]
 python-version = "3.13"
+python = ".venv"
```
ryneeverett: This breaks my nix-shell. It seems like this setting shouldn't be used if development environment diversity is going to be permitted.

```console
$ ty check
ty failed
  Cause: Invalid `environment.python` setting

--> Invalid setting in configuration file `/home/ryne/projects/bugwarrior/pyproject.toml`
    |
115 | [tool.ty.environment]
116 | python-version = "3.13"
117 | python = ".venv"
    |          ^^^^^^^ does not point to a Python executable or a directory on disk
118 |
119 | [tool.ty.src]
    |

  Cause: No such file or directory (os error 2)
```

PR #1204, opened 2026-04-27
Remove interactive flag

### LINKED ISSUE #1199: Move `interactive` to root `Config`
Per @Lotram in #1190.

Alternatively, I wonder if it would make more sense to eliminate this flag? I believe the original problem it solved is that if you run bugwarrior as a timed job rather than in a terminal and it prompts for a password it will hang indefinitely which can be a pain to debug in some deployment environments versus an error. Seems like it would simplify both our configuration data model and the command line interface to introduce a timeout in these circumstances.

Lotram:
Fixes #1199

* `--interactive` is still accepted, but do nothing except raising a deprecation warning
* refactor `get_service_password` to simplify the logic
* add several tests for `get_service_password`

### DIFF
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -5,6 +5,7 @@ import logging
 import os
 import sys
 from typing import TYPE_CHECKING, Any
+import warnings
 
 import click
 from lockfile import LockTimeout
@@ -110,8 +111,6 @@ def pull(
     Relies on configuration file.
     """
     if interactive:
-        import warnings
-
         warnings.warn(
             "The --interactive flag is deprecated and has no effect. "
             "Interactive mode is now detected automatically via sys.stdin.isatty().",
--- a/bugwarrior/command.py
+++ b/bugwarrior/command.py
@@ -29,11 +29,9 @@ def _get_section_name(flavor: str | None) -> str:
     return 'general'
 
 
-def _try_load_config(
-    main_section: str, interactive: bool = False, quiet: bool = False
-) -> "Config":
+def _try_load_config(main_section: str, quiet: bool = False) -> "Config":
     try:
-        return load_config(main_section, interactive, quiet)
+        return load_config(main_section, quiet)
     except OSError:
         # Our standard logging configuration depends on the bugwarrior
         # configuration file which just failed to load.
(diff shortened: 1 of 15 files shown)


bugwarrior/command.py
```diff
@@ -107,10 +109,19 @@ def pull(
 
     Relies on configuration file.
     """
+    if interactive:
+        import warnings
```
ryneeverett: Can we keep imports at the top of the file? it strikes me as premature optimization to import standard library modules conditionally.

tests/config/test_secrets.py
```diff
@@ -1,8 +1,113 @@
...
+    def _call(self, oracle):
+        return secrets.get_service_password(self.SERVICE, self.USERNAME, oracle=oracle)
+
+    def _mock_keyring(self, password=None, locked=False):
+        """Return a mock keyring module.
+
+        If "locked" is True, get_password raises KeyringLocked.
+        Otherwise it returns the given password.
+        """
+        import keyring.errors
```
ryneeverett: Same critique here -- can we move this import to the top of file?

bugwarrior/config/secrets.py
```diff
@@ -18,9 +18,20 @@ def get_keyring() -> ModuleType:
     return keyring
 
 
-def get_service_password(
-    service: str, username: str, oracle: str | None = None, interactive: bool = False
-) -> str:
+def _ask_password(service: str) -> str:
+    import getpass
```
ryneeverett: Same critique here, although you didn't introduce the inline import.

bugwarrior/config/schema.py
```diff
@@ -138,10 +138,6 @@ class MainSectionConfig(BaseConfig):
     # required
     targets: ConfigList
 
-    # added during configuration loading
-    #: Interactive status.
-    interactive: bool
-
```
ryneeverett, line 144: Since this is in the documented stable API, can we actually leave it in with a default value after all? I'm thinking:

```py
#: *DEPRECATED*
interactive: bool = True
```

ryneeverett (approved):

ryneeverett:
FYI, we're seeing the same CI failure in the other PR, though it doesn't seem to be consistent. Don't have time to investigate further at the moment.

Lotram:
https://github.com/GothenburgBitFactory/bugwarrior/pull/1205 might fix the flaky error

Issue #1176, opened 2026-01-06
Consider simplifying ini config parsing


Lotram:
I started working on static config models. While doing that, I added some tests on the `load.parse_file` function. 
I then realized that all the parsing done in `ini2toml_plugin` is not used when parsing a ini file (except for the `ini2toml` command). 

I see 2 problems with that:
* The result of `parse_file` is quite different, depending on the input file (toml vs ini)
* The ini parsing logic is duplicated between `parse_file` and `ini2toml_plugin`, with different results

Can't we simplify this ? We could either
* add ini2toml to required dependencies and always use it when parsing an ini file (we could even write the toml file ourselves ? )
* keep it as optional, but force the user to install it if they want to keep an ini file (or suggest to run it once to use a toml file instead)
* same, but always raising an error when trying to parse an ini file, and ask the user to migrate to toml file, using the ini2toml command

I know those solutions are drastic, and we probably want to deprecate it first, but I think we should avoid having multiple ways to parse a config file, and multiple results.

ryneeverett:
I'd like to avoid adding ini2toml as a required dependency if possible because that would be a burden on distributions since I don't believe ini2toml is widely packaged.

I gather from your "drastic" solutions that the problem is severe? Could you give some examples or maybe a test that demonstrates the issue?

related: #1042

Lotram:
"drastic" might not be the right word, and the problem is not actually severe. Basically, when parsing the ini file, all the int / boolean values are parsed as strings, and the comma-separated lists are strings as well, instead of string. 

So everything is handled by pydantic, but having to support two such files (+ the ini2toml parser) can lead to more errors (as we saw with flavors). So I think we should aim at having only one file (toml) in the future.

So I guess the actual solutions depends on how fast / slow we want to implement breaking changes

ryneeverett:
I agree that we should probably converge on only supporting toml at some point, but I'm not inclined to drop ini support just because there are probably more bugs which we haven't discovered. (And I agree that those bugs are probably waiting to be found!)

My thought is that at some point we will either discover severe bugs which are impractical to fix or impose a maintenance cost on the rest of the code base we don't want to pay. But at this point it seems like continued ini support is zero cost.

One step I do think would make sense to take immediately would be for the documentation to say that toml is "recommended" and to flip the tabs so that toml is first and default.

Lotram:
What do you think about also logging a deprecation warning when parsing ini ?

ryneeverett:
[...]

PR #1206, opened 2026-04-28
explictly state in documentation that toml is the prefered config format

Lotram:
#1176 

toml is now the first format in the docs, and we clearly indicate it's the preferred one.

### DIFF
--- a/bugwarrior/docs/common_configuration.rst
+++ b/bugwarrior/docs/common_configuration.rst
@@ -1,7 +1,7 @@
 How to Configure
 ================
 
-Bugwarrior's configuration file can be written either in `ini <https://en.wikipedia.org/wiki/INI_file#Format>`_ or `toml <https://toml.io>`_ format. See :ref:`the man page <configuration-files>` to determine the name and location of this file.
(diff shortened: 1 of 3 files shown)


bugwarrior/docs/conf.py
```diff
@@ -118,7 +118,7 @@
 # show_authors = False
 
 # The default language to highlight source code in.
-highlight_language = 'ini'
```
ryneeverett, line 121: Please delete the comment too.

bugwarrior/docs/common_configuration.rst
```diff
@@ -1,7 +1,7 @@
 How to Configure
 ================
 
-Bugwarrior's configuration file can be written either in `ini <https://en.wikipedia.org/wiki/INI_file#Format>`_ or `toml <https://toml.io>`_ format. See :ref:`the man page <configuration-files>` to determine the name and location of this file.
+Bugwarrior's configuration file can be written in `toml <https://toml.io>`_ (recommended) or `ini <https://en.wikipedia.org/wiki/INI_file#Format>`_ (deprecated) format. See :ref:`the man page <configuration-files>` to determine the name and location of this file.
```
ryneeverett: I don't think I'm ready to say "deprecated". I don't foresee a need to drop or lessen support for ini in the near future. The only reason it isn't recommended is that there aren't currently any bugwarrior core developers (just you and I) using it personally.

ryneeverett (changes requested):

ryneeverett (approved):

PR #1207, opened 2026-04-29
Add prek pre-commit hooks for ruff and ty

Lotram:
Final step for #1151 

So there are multiple ways to define the pre-commits, we can either:

* use the python env from the project. it uses a small script in `run.sh` to determine what env to use (either using `uv` or from `pip`), but I'm not sure how robust this is.
* let `prek` define its own environment, so it does not depend on how the python env is installed, the major downside is having to keep the versions from `pyproject.toml` and `.pre-commit-config.yaml` in sync.

the `run.sh` alternative also allows us to add some help messages to disable the pre commits.
As discussed, neither `ruff-format` nor `ruff-lint` modify the code, it simply fails even something is wrong.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,2 +1,22 @@
 ci: {}
-repos: []
+repos:
+  - repo: local
+    hooks:
+      - id: ruff-format
+        name: ruff format
+        entry: scripts/run.sh ruff format --force-exclude --check
+        language: system
+        types_or: [python, pyi]
+
+      - id: ruff-lint
+        name: ruff lint
+        entry: scripts/run.sh ruff check --force-exclude
+        language: system
+        types_or: [python, pyi]
+
+      - id: ty
+        name: ty check
+        entry: scripts/run.sh ty check
+        language: system
+        types_or: [python, pyi]
+        pass_filenames: false
--- a/bugwarrior/docs/contributing.rst
+++ b/bugwarrior/docs/contributing.rst
@@ -71,6 +71,28 @@ separately, but you may find it convenient to run ``ruff check``, ``ruff
 format``, and ``ty check``, or integrate `ruff <https://docs.astral.sh/ruff/editors/setup/>`_
 and `ty <https://docs.astral.sh/ty/editors/>`_ with your editor.
 
+Optionally, you can install the provided hooks to have ruff and ty run
+automatically before each commit. The hooks use `prek <https://github.com/j178/prek>`_
+(or `pre-commit <https://pre-commit.com/>`_ as an alternative):
+
+.. tab:: pip
+
+   .. code-block:: bash
+
+    $ pip install prek
+    $ prek install
+
+.. tab:: uv
+
+   .. code-block:: bash
+
+    $ uv tool install prek
+    $ prek install
+
+The hooks use ``scripts/run.sh`` to invoke ruff and ty from the project's
+virtualenv (via ``uv run`` for uv users, or ``.venv`` for pip users), so they
+always run the same versions as the test suite.
+
 Making a pull request
 ---------------------
 
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+# Run a command via `uv run` if uv is available, otherwise run it directly.
+# If a .venv directory exists, prepend it to PATH so the venv's tools are
+# used without requiring the user to have manually activated it.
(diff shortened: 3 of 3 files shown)


.pre-commit-config.yaml
ryneeverett, line 1: Why do we need `--force-exclude` on the ruff invocations?

ryneeverett (commented):
What would you think about automated testing of the pre-commit hooks? I'm thinking:

- Add `prek` to "test" dependency group.
- Simplify contributing docs and eliminate tab because all you have to do is `prek install`.
- Add a `prek run` test to test_general.py

This would test the hook script with both uv and pip in CI.

.pre-commit-config.yaml
Lotram:
```
      --force-exclude
          Enforce exclusions, even for paths passed to Ruff directly on the command-line.
```
all staged files are passed as argument to ruff. Without this option, staged files we'd normally exclude (from the config in `pyproject.toml`) would be formatted / linted, potentially raising errors even though we want to ignore them.

Lotram:
> * Add prek to "test" dependency group.
> * Simplify contributing docs and eliminate tab because all you have to do is prek install.

Sure

> * Add a prek run test to test_general.py

Would it be in addition to existing tests, or to replace them ? `prek` only runs ruff on staged files, so we can't simply replace them, but just adding a new test would run `ty` twice  (as `ty` always run on all files, thanks to `pass_filenames: false`). 

As I'm writing that, I'm also thinking we could always run `ruff` on the whole codebase, instead of only the staged files, as it's much faster than ty anyway:
```
❯ time uv run ruff format --check
81 files already formatted
uv run ruff format --check  0.03s user 0.05s system 127% cpu 0.063 total

❯ time uv run ruff check
All checks passed!
uv run ruff check  0.03s user 0.06s system 177% cpu 0.055 total

❯ time uv run ty check
All checks passed!
uv run ty check  0.88s user 0.24s system 562% cpu 0.200 total
```

.pre-commit-config.yaml
ryneeverett: Ah, I'd read the docs but your explanation makes more sense to me. Shouldn't we add this flag to the tests in test_general.py too? Couldn't this be an issue when running the tests locally in development?

ryneeverett: Rereading that helptext again, I think it is totally unclear that configured exclusions are ignored without this flag. Are you sure about that?

ryneeverett:
I'd like it to be an additional test which ideally only fails if prek is misconfigured or the script is broken. It seems like just running a single hook is sufficient so let's go with the fastest one -- `prek run ruff-format`.

However, it isn't clear to me that there is any way to distinguish between a non-zero exit code caused by script failure or misconfiguration versus one caused by hook failure. I haven't played with it though. Maybe this test could run conditionally only if test_ruff_format passes? It might be annoying if this test always failed when `ruff format --check` fails.

.pre-commit-config.yaml
Lotram: `test_general.py` is ok, because it runs a plain `ruff check` / `ruff format`, no files as arguments. 

but staged files are passed as arguments to ruff, inside a pre-commit: `ruff check file_1.py file_2.py`. If `file_1.py` is excluded from ruff, it would still be formatted / linted because it has been passed as an argument. 

the [doc](https://docs.astral.sh/ruff/settings/#force-exclude) says:
> This is useful for [pre-commit](https://pre-commit.com/), which explicitly passes all changed files to the [ruff-pre-commit](https://github.com/astral-sh/ruff-pre-commit) plugin, regardless of whether they're marked as excluded by Ruff's own settings

Lotram:
If you think it's easier, we can also let prek define its own environment, ruff and ty don't even have to be installed locally, see the [doc](https://github.com/astral-sh/ruff-pre-commit) here, and simply have a test to ensure versions in `pyproject.toml` and `.pre-comit-config.yaml` are the same

Edit: apparently there is [no equivalent](https://github.com/astral-sh/ty/issues/269) for ty at the moment

Lotram:
> However, it isn't clear to me that there is any way to distinguish between a non-zero exit code caused by script failure or misconfiguration versus one caused by hook failure. I haven't played with it though. Maybe this test could run conditionally only if test_ruff_format passes? It might be annoying if this test always failed when ruff format --check fails.

We could add a hook that simply does something like `ruff --version`, and run this hook only from the tests. I think it's a bit weird, but that would work

ryneeverett:
Testing isn't as clean and easy as I'd hoped it would be so lets skip it for now.

PR #1209, opened 2026-05-06
Replace get_keyring_service by KEYRING_SERVICE class variable

Lotram:
## First commit
Fix a test in `test_service.py` : since [this commit](https://github.com/GothenburgBitFactory/bugwarrior/commit/a226abd6591093ded6b5cbf2dcf60d5f0deb84bb#diff-d7fb70d71c01e6c6a932249a22ad45e9e8e5dd800f8d9260922b7726ada72402) the test check each abstract method is _defined_ only once, which does not really make sense. I changed it back to check the number of calls, while adding an allowlist for some abstract methods we do want to call from concrete methods. 

An alternative could be to simply remove this test, since we already have an exception

## Second commit
Replace `Service.get_keyring_service` by a `ServiceConfig.KEYRING_SERVICE` string

### DIFF
--- a/bugwarrior/config/schema.py
+++ b/bugwarrior/config/schema.py
@@ -1,9 +1,11 @@
+from functools import cache
+from importlib.metadata import entry_points
 import logging
 import os
 from pathlib import Path
 import re
 import typing
-from typing import Annotated, Any, Literal
+from typing import TYPE_CHECKING, Annotated, Any, Literal
 
 import pydantic
 from pydantic import (
@@ -22,6 +24,9 @@ import taskw
 import taskw.task
 
 from .data import BugwarriorData, get_data_path
+
+if TYPE_CHECKING:
+    from bugwarrior.services import Service
 
 log = logging.getLogger(__name__)
 
@@ -215,8 +220,6 @@ class ServiceConfig(_ServiceConfig):
 
     @property
     def keyring_service(self) -> str:
-        # FIXME change this import and move it outside method after rebase
-        from bugwarrior.collect import get_service
 
         service = get_service(self.service)
         if service.API_VERSION < 2:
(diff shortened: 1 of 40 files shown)


ryneeverett (commented):
Regarding the first commit -- oops! Thanks for catching this.

It isn't clear to me that an exception is justified for `get_keyring_service` though. Perhaps this would be a good time to eliminate the method in favor of a class property string on which `Service.get_secret` calls `.format(config)`?

Lotram:
I'm ok with removing `get_keyring_service`, but it wouldn't work right away with `KanboardService` and `PhabricatorService`:

```
    def get_keyring_service(config: KanboardConfig) -> str:
        parsed = urlparse(config.url)
        return f"kanboard://{config.username}@{parsed.netloc}"
```

One solution would be to use a `computed_field for the host, in both `KanboardConfig` and `PhabricatorConfig`

 Another solution would be to make `get_keyring_service` a method of the `ServiceConfig` subclasses, instead of the `Service` ones ? IMO it makes sense to attach this to the config. Users could want to customize this in their config, if we ever want to support this.

ryneeverett:
The computed_field solution seems good. Making this a string rather than a method is the main thing I'd like to see.

I'd be ok with that string moving to ServiceConfig, but I wouldn't want to see the `get_keyring_service` method moved to ServiceConfig. As you probably realize, it shouldn't be configurable at this point since there is no apparent demand for that.

bugwarrior/config/schema.py
```diff
@@ -197,10 +197,16 @@ class ServiceConfig(_ServiceConfig):
...
+    KEYRING_SERVICE: typing.ClassVar[str]
+
     # Added before validation (computed field)
     service: str
     target: str
 
+    @property
+    def keyring_service(self) -> str:
+        return self.KEYRING_SERVICE.format(**self.model_dump())
+
```
ryneeverett: Can we move this method down below all the class variables with the rest of the methods?

bugwarrior/services/pagure.py
```diff
@@ -16,6 +16,7 @@
 class PagureConfig(config.ServiceConfig):
     # strictly required
```
ryneeverett: Let's remove this unnecessary comment which isn't quite accurate any more after this change.
```suggestion
```

ryneeverett (commented):
This is a breaking change to the service API so we need to bump the version to 2.0 (LATEST_API_VERSION, API_VERSION of each service, header of api.rst). I'd also like to see backwards compatibility:

```py
    @property
    def keyring_service(self) -> str:
        service = get_service(self.service).API_VERSION
        if service.API_VERSION < 2:
            return service.get_keyring_service(self)
        return self.KEYRING_SERVICE.format(**self.model_dump())
```

ryneeverett:
One more thing I forgot in addition to the rebase -- please add a Changelog section to the bottom of api.rst and mention the removal of get_keyring_service and addition of KEYRING_SERVICE.

ryneeverett (approved):

Issue #1077, opened 2024-11-13
taskchampion integration


ryneeverett:
In [the last community meeting](https://github.com/GothenburgBitFactory/taskwarrior/discussions/3675) @djmitche mentioned that he was interested in seeing an integration of taskchampion into bugwarrior and I'd be curious for his thoughts on that. There is an effort to create python bindings to taskchampion -- https://github.com/GothenburgBitFactory/taskchampion-py -- and I'm curious if we should be depending on an upstream like that or creating our own bindings for the presumably minimal subset of the API we actually need to use.

My primary concern I hope Dustin can address is about taskchampion's database schema version interoperability. As I understand it, bugwarrior would have taskchampion as a dependency, so breaking changes in taskchampion wouldn't be an issue because we'd be linking against a version of our choosing. However, the user's taskwarrior installation might be linked to a different version of taskchampion and the database might have a different schema. How much interoperability should we expect between different taskchampion versions going forward? I suspect it would be very disruptive to our users if bugwarrior didn't have several years of interoperability with taskwarrior.

djmitche:
My intention is for sync to be the means by which different applications communicate, rather than sharing a database. Having thought about modifying the sync protocol by adding more operations in https://github.com/GothenburgBitFactory/taskchampion/issues/373, I suspect this is nearly impossible, so I think this is likely a _very_ stable way to interoperate.

Sync can be performed locally, so even for a single-machine setup this should be pretty straightforward.

That said, I'm sure folks will want to use a single DB for both bugwarrior and taskwarrior, so we should at least make that option as good as we can. https://github.com/GothenburgBitFactory/taskchampion/issues/477 should help to at least prevent DB corruption if DB versions don't align -- at the cost of making one of the two applications fail with a DB version error. And, it's worth noting that the operations work was a pretty major change to TaskChampion and I don't expect that to happen too often. Breaking task properties out into a `task_property` table for easier querying is the only incompatible schema change I can see on the horizon, and I'm not even sure that's a good idea (virtual indices might be a better idea, and are backward compatible).

Regarding the Python bindings -- I'd prefer to depend on an upstream, just to focus attention on the quality of that upstream.

ryneeverett:
[...]

PR #1210, opened 2026-05-08
improvements in db.py

Lotram:
I've started looking at db.py to prepare for #1077 and #1078 
I've removed some unused code and simplified some logic. 

There are two improvements I have not implemented yet, because I think it's worth discussing it @ryneeverett 
* the smaller one: `merge_left` / `replace_left`, and duplication / ordering
    *  on `tags`, but taskwarrior treats them as a simple (ordered) set, so I don't really understand why we can simply do `local["tags"] |= remote["tags"]` (merge) / `local["tags"] = remote["tags"]` (replace), and order them if needed. 
    *  on `annotations` the tests I added in `tests/test_synchronize_ordering.py` also show that deduplication and ordering are inconsistent, because `taskwarrior` consider them as a normal list, but `taskw` uses a set. Depending on task creation / update, we could end up with duplicates or not, ordered or not. My suggestion would be to always treat them as set (as `taskw` does), I don't see a good use case for annotation duplication, but maybe I'm wrong. 
  
* bigger one: all functions involving `key_lists` rely on finding issues in DB based on the UDAs that identify them (defined by `UNIQUE_KEY`. I see multiple issues with that

   * the queries to retrieve a single issue from taskwarrior (`find_taskwarrior_uuid`), or all of them (`get_managed_task_uuids`) are complex (especially the first one). As we already compute a bugwarrior identifier, why can't we store it as a UDA (hashing it first) for all tasks created by bugwarrior ? Retrieving a single issue from taskwarrior (or all of them) would be much simpler and straightforward. 
   
   * I'm not sure about this one, but if I understand correctly, the current behavior also explains why each service needs to have specific UDAs (`githuburl`, `gitlaburl`, `linearurl`...). Wouldn't it be cleaner to have more generic UDAs (`url` in that case) ? The `UNIQUE_KEY` tuple could simply contain the `target` / `service` (depending on whether we want to consider  duplicated issues across targets from the same service as actual duplicates or not) to ensure the identifier is indeed unique. We probably can't change this overnight, but I still think we should consider the idea
  
   *  This behavior in case of multiple targets using the same service (e.g. in my own config, multiple Gmail targets) is not explicit: should we consider duplicated issues from multiple targets as the same one (currently the case, based on `make_identifier`) or not ?

### DIFF
--- a/bugwarrior/collect.py
+++ b/bugwarrior/collect.py
@@ -1,7 +1,8 @@
-from collections.abc import Iterator
+from collections.abc import Iterable, Iterator
 import copy
 from functools import cache
 from importlib.metadata import entry_points
+import json
 import logging
 import multiprocessing
 import time
@@ -9,6 +10,8 @@ from typing import TYPE_CHECKING, Any
 
 from jinja2 import Template
 from taskw.task import Task
+
+from bugwarrior.types import CollectedIssue, CollectionErrorData, TaskwarriorData
 
 if TYPE_CHECKING:
     from bugwarrior.config.validation import Config
@@ -81,7 +84,9 @@ def _aggregate_issues(service: "Service", queue: multiprocessing.Queue) -> None:
         log.info(f"Done with [{target}] in {duration}.")
 
 
-def aggregate_issues(conf: "Config", debug: bool) -> Iterator[dict | tuple[str, str]]:
+def aggregate_issues(
+    conf: "Config", debug: bool
+) -> Iterator[CollectedIssue | CollectionErrorData]:
     """Return all issues from every target."""
     log.info("Starting to aggregate remote issues.")
 
@@ -111,8 +116,7 @@ def aggregate_issues(conf: "Config", debug: bool) -> Iterator[dict | tuple[str, 
     while currently_running > 0:
         issue = queue.get(True)
         try:
-            record = TaskConstructor(issue).get_taskwarrior_record()
-            record['target'] = issue.config.target
+            record = TaskConstructor(issue).get_data_to_sync()
             yield record
         except AttributeError:
             if isinstance(issue, tuple):
@@ -120,11 +124,24 @@ def aggregate_issues(conf: "Config", debug: bool) -> Iterator[dict | tuple[str, 
                 completion_type, target = issue
                 if completion_type == SERVICE_FINISHED_ERROR:
                     log.error(f"Aborted [{target}] due to critical error.")
-                    yield ('SERVICE FAILED', target)
+                    yield CollectionErrorData('SERVICE FAILED', target)
                 continue
             raise
 
     log.info("Done aggregating remote issues.")
+
+
+# NOTE: should this be a method of Issue instead ?
+def make_unique_identifier(
+    key_list: Iterable[str], taskwarrior_data: TaskwarriorData
+) -> str:
+    """For a given issue, make an identifier from its unique keys.
+
+    This is not the same as the taskwarrior uuid, which is assigned
+    only once the task is created.
+    """
+    subset = {key: taskwarrior_data[key] for key in key_list}
+    return json.dumps(subset, sort_keys=True)
 
 
 class TaskConstructor:
@@ -152,6 +169,10 @@ class TaskConstructor:
             record['tags'] = []
         if refined:
             record['tags'].extend(self.get_added_tags())
+
+        # Blank priority should mean *no* priority
+        if record['priority'] == '':
+            record['priority'] = None
         return record
 
     def get_template_context(self) -> dict[str, Any]:
@@ -168,3 +189,11 @@ class TaskConstructor:
             elif field == 'description':
                 record['description'] = self.issue.get_default_description()
         return record
+
+    def get_data_to_sync(self) -> CollectedIssue:
+        taskwarrior_data = self.get_taskwarrior_record()
+        return CollectedIssue(
+            taskwarrior_data=taskwarrior_data,
+            identifier=make_unique_identifier(self.issue.UNIQUE_KEY, taskwarrior_data),
+            target=self.issue.config.target,
+        )
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -1,6 +1,5 @@
 from collections.abc import Iterable, Iterator
 import itertools
-import json
 import logging
 import re
 import subprocess
@@ -11,6 +10,7 @@ from taskw.exceptions import TaskwarriorError
 
 from bugwarrior.collect import get_service
 from bugwarrior.notifications import send_notification
+from bugwarrior.types import CollectedIssue, CollectionErrorData
 
 if TYPE_CHECKING:
     from bugwarrior.config.validation import Config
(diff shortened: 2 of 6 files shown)


ryneeverett (commented):
Only looked at the first commit so far (**removed unused hamming distance functions**) -- seems like `hamdist` could be replaced by an equality check without changing the behavior, but no longer shortening the longer of the two strings is a behavior change. Are we sure the hamming behavior isn't desirable? (I must admit, I can't think of any good reason for it myself but somebody did bother to implement it...)

If we're going to remove the hamming behavior we should at the very least rename the `hamming` parameter to `annotations` and update the docstring accordingly.

You might consider making a separate PR for this commit for a faster review/merge cycle.

ryneeverett:
- merge_left / replace_left, and duplication / ordering
  - tags: :+1: 
  - annotations: Doesn't "always treat them as set" mean we wouldn't preserve order any more? That seems wrong but perhaps I misunderstand.
- UNIQUE_KEY
  - bugwarrior identifier: I suspect the reason we didn't go with our own UDA is to avoid adding visual junk to the default task report. I don't think the complexity of `find_taskwarrior_uuid` outweighs that virtue.
  - url prefixes: My recollection is that the UDA prefixes predate the use of UDA's to identify tasks and we used to use description pattern matching (which I think is why the default descriptions have "(bw)" in them). Regardless, I agree that removing the prefixes would be an improvement and since that is a "visual junk" reduction I think this helps justify a bugwarrior id UDA. I do think this would trigger a major version though.
  - duplicate issues from multiple targets: I think the current behavior is correct. Why would one want duplicate tasks?

Lotram:
> Are we sure the hamming behavior isn't desirable? 
We already remove any special char from the string, I'm not sure hamming distance is necessary here. I don't see why annotations could differ slightly between calls. If that happens, I think it should be fixed per service, not here. 

Good point about string truncation. Even though I don't really see why we should handle this case (why would an annotation's length should change between calls ?), that's still a behavior change, I'll fix that. 

> annotations: Doesn't "always treat them as set" mean we wouldn't preserve order any more? That seems wrong but perhaps I misunderstand

My point was: `taskw` does not preserver the order when updating annotations, but you're right, we should try to preserve it in our side. Are you ok with deduplicating annotations though ? 


About `UNIQUE_KEY`:
* One other argument in favor of a bugwarrior ID: for now, if we ever change the naming or the number of fields in `UNIQUE_KEY`, all behaviors related to `key_list` are broken, whereas with an identifier, the identifier computation becomes an implementation detail inside bugwarrior. If we later change the way we compute this identifier (which can be customized per service), the only broken feature is task deduplication (and we can find solutions to mitigate this), we'll still be able to retrieve all bugwarrior tasks.  I'll write an issue for this
* duplicate issues from multiple targets: I'm fine with this, but since the config is "per-target" rather than "per-service", I think it would make sense to explicit this in the documentation (and the code). It's quite hidden inside the `key_list` behavior right now

Also note that the two other commits ([simplify key_list](https://github.com/GothenburgBitFactory/bugwarrior/pull/1210/changes/f0ea52e0c4f0e6db20080791e5f0ac946d9051f8) and [Move make_identifier function to collect.py](https://github.com/GothenburgBitFactory/bugwarrior/pull/1210/changes/e428f9c2e11bef24f6142dcf99f9770b7dce8093)) are pure code refactoring, and don't change the current behavior.

ryneeverett:
> Are you ok with deduplicating annotations though ?

All else equal, I don't think we should deduplicate annotations. If somebody accidentally posts the same comment twice on Github, ideally it would show up twice in the annotations.

However, since taskw already uses sets I think it's probably better to do so more consistently and deduplicate on task creation. I also think it would be fine for the taskchampion backend to have the same behavior if it simplifies to the implementation.

> One other argument in favor of a bugwarrior ID: for now, if we ever change the naming or the number of fields in UNIQUE_KEY, all behaviors related to key_list are broken...

I see your point but the only consequence of this is that all issues will be closed and re-opened and any data the user manually added to the task is lost. Given that we've been using this mechanism for quite a while and never needed to do this, it seems like both a low-likelihood and low-impact concern.

bugwarrior/db.py
```diff
@@ -312,12 +296,8 @@ def synchronize(
                 except UnicodeDecodeError:
                     log.warning("Failed to interpret %r as utf-8" % key)
 
-        # Blank priority should mean *no* priority
-        if issue['priority'] == '':
-            issue['priority'] = None
-
         # Target was only tacked on to pass configuration to this function.
```
ryneeverett, line 299: The above comment should be removed as it was only explaining the `pop()`.

bugwarrior/collect.py
```diff
@@ -111,22 +116,34 @@ def aggregate_issues(conf: "Config", debug: bool) -> Iterator[dict | tuple[str,
...
                     log.error(f"Aborted [{target}] due to critical error.")
-                    yield ('SERVICE FAILED', target)
+                    yield CollectionErrorData('SERVICE FAILED', target)
                 continue
             raise
 
     log.info("Done aggregating remote issues.")
 
 
+# NOTE: should this be a method of Issue instead ?
```
ryneeverett, line 134: Did you mean a method of TaskConstructor? That would make sense to me. I think this is the kind of higher-level concern that I was trying to extract out of the Service/Issue base classes to minimize their scope.

bugwarrior/db.py
ryneeverett, line 1: Maybe `key_lists` should be renamed `unique_keys` / `build_unique_keys`? It's been a while since any `list`s were involved and the name was already quite vague.

ryneeverett (commented):
I didn't do a thorough review, but the basic ideas of the other two commits seems sound: 
- **simplify key_list**: The <service> keys of the former `key_list` were unused so we may as well just have lists (well actually tuples -- seems like the `list[str]` type annotations were wrong before?). And since the UNIQUE_KEY is the same for each given service even if the user has multiple targets, they may as well be deduplicated in a `set()` to simplify the query. 👍 
- **Move make_identifier function to collect.py**: We were formerly making the unique identifier after we'd already "forgotten" which service the issue belongs to which means we had to check each service's UNIQUE_KEY to figure out which one we were dealing with. So instead we calculate it at an earlier stage when we still have access to that data. This is a good opportunity to make a proper data structure `CollectedIssue` rather than temporarily tack more data onto the task like we formerly were with `target`. 👍

bugwarrior/collect.py
Lotram: No I was really suggesting `Issue`, as `UNIQUE_KEY` is an attribute of `Issue`, and the `make_unique_identifier` method depends only on this attribute, and `Issue.get_taskwarrior_record`

we could have something like 
```py
from functools import cached_property


class Issue(abc.ABC):
    UNIQUE_KEY: tuple[str, ...]

    @abc.abstractmethod
    @cached_property
    def to_taskwarrior(self) -> TaskwarriorData:
        """Transform a foreign record into a taskwarrior dictionary."""
        raise NotImplementedError()

    @property
    def unique_identifier(self) -> str:
        subset = {key: self.to_taskwarrior[key] for key in self.UNIQUE_KEY}
        return json.dumps(subset, sort_keys=True)
```
Issue could also be a pydantic model, where `to_taskwarrior` is simply the `model_dump` method, and `unique_identifier` could be a computed_field, as I think it makes sense to attach the identifier to the `Issue` object. 

But if you wanted to extract some logic away from this, I'm ok with what we have now.

ryneeverett: Yeah, I'd rather keep it out of Issue at this point since that would make it accessible to services as part of the stabilized API.

> Issue could also be a pydantic model, where `to_taskwarrior` is simply the `model_dump` method, and `unique_identifier` could be a computed_field, as I think it makes sense to attach the identifier to the `Issue` object.

This is very interesting. My original motivation for using pydantic for configuration is that I think the more we can make services declarative data rather than arbitrary code the better off we are and this seems like another great step in that direction.

Lotram:
I created 1 MR for each commit in this one, for faster iterations.

PR #1211, opened 2026-05-18
taiga: import due dates

Fubukimaru:
Taiga exposes a due_date field for user stories/tasks, but the Taiga service did not map it to Taskwarrior's native due field.

This patch imports due_date using the existing Issue.parse_date() helper, inspired in the pattern used by other services such as Trello. Includes tests.

### DIFF
--- a/bugwarrior/services/taiga.py
+++ b/bugwarrior/services/taiga.py
@@ -36,7 +36,7 @@ class TaigaIssue(Issue):
     UNIQUE_KEY = (URL,)
 
     def to_taskwarrior(self) -> dict[str, Any]:
-        task = {
+        return {
             'project': self.extra['project'],
             'annotations': self.extra['annotations'],
             self.URL: self.extra['url'],
@@ -44,12 +44,8 @@ class TaigaIssue(Issue):
             'tags': self.get_tags(),
             self.FOREIGN_ID: self.record['ref'],
             self.SUMMARY: self.record['subject'],
+            'due': self.parse_date(self.record.get('due_date'))
         }
-
-        if self.record.get('due_date'):
-            task['due'] = self.parse_date(self.record['due_date'])
-
-        return task
 
     def get_tags(self) -> list[str]:
         return [x if isinstance(x, str) else x[0] for x in self.record['tags']]
--- a/bugwarrior/services/taiga.py
+++ b/bugwarrior/services/taiga.py
@@ -44,7 +44,7 @@ class TaigaIssue(Issue):
             'tags': self.get_tags(),
             self.FOREIGN_ID: self.record['ref'],
             self.SUMMARY: self.record['subject'],
-            'due': self.parse_date(self.record.get('due_date'))
+            'due': self.parse_date(self.record.get('due_date')),
         }
 
     def get_tags(self) -> list[str]:
--- a/bugwarrior/services/taiga.py
+++ b/bugwarrior/services/taiga.py
@@ -35,19 +35,19 @@ class TaigaIssue(Issue):
     }
     UNIQUE_KEY = (URL,)
 
-    def to_taskwarrior(self):
+    def to_taskwarrior(self) -> dict[str, Any]:
         task = {
-            "project": self.extra["project"],
-            "annotations": self.extra["annotations"],
-            self.URL: self.extra["url"],
-            "priority": self.config.default_priority,
-            "tags": self.get_tags(),
-            self.FOREIGN_ID: self.record["ref"],
-            self.SUMMARY: self.record["subject"],
+            'project': self.extra['project'],
+            'annotations': self.extra['annotations'],
+            self.URL: self.extra['url'],
+            'priority': self.config.default_priority,
+            'tags': self.get_tags(),
+            self.FOREIGN_ID: self.record['ref'],
+            self.SUMMARY: self.record['subject'],
         }
 
-        if self.record.get("due_date"):
-            task["due"] = self.parse_date(self.record["due_date"])
+        if self.record.get('due_date'):
+            task['due'] = self.parse_date(self.record['due_date'])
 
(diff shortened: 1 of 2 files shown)


Fubukimaru:
It looks like the tests are failing here:

```
=========================== short test summary info ============================
FAILED tests/test_general.py::TestGeneral::test_ty_check - subprocess.CalledProcessError: Command '['ty', 'check']' returned non-zero exit status 1.
=========== 1 failed, 293 passed, 1 skipped, 175 warnings in 12.90s ============
```

But I'm not sure that this comes from Taiga.

```
=================================== FAILURES ===================================
__________________________ TestGeneral.test_ty_check ___________________________

self = <tests.test_general.TestGeneral testMethod=test_ty_check>

    def test_ty_check(self):
>       subprocess.run(['ty', 'check'], check=True)

tests/test_general.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = None, capture_output = False, timeout = None, check = True
popenargs = (['ty', 'check'],), kwargs = {}
process = <Popen: returncode: 1 args: ['ty', 'check']>, stdout = None
stderr = None, retcode = 1

    def run(*popenargs,
            input=None, capture_output=False, timeout=None, check=False, **kwargs):
        """Run command with arguments and return a CompletedProcess instance.
    
        The returned instance will have attributes args, returncode, stdout and
        stderr. By default, stdout and stderr are not captured, and those attributes
        will be None. Pass stdout=PIPE and/or stderr=PIPE in order to capture them,
        or pass capture_output=True to capture both.
    
        If check is True and the exit code was non-zero, it raises a
        CalledProcessError. The CalledProcessError object will have the return code
        in the returncode attribute, and output & stderr attributes if those streams
        were captured.
    
        If timeout is given, and the process takes too long, a TimeoutExpired
        exception will be raised.
    
        There is an optional argument "input", allowing you to
        pass bytes or a string to the subprocess's stdin.  If you use this argument
        you may not also use the Popen constructor's "stdin" argument, as
        it will be used internally.
    
        By default, all communication is in bytes, and therefore any "input" should
        be bytes, and the stdout and stderr will be bytes. If in text mode, any
        "input" should be a string, and stdout and stderr will be strings decoded
        according to locale encoding, or by "encoding" if set. Text mode is
        triggered by setting any of text, encoding, errors or universal_newlines.
    
        The other arguments are the same as for the Popen constructor.
        """
        if input is not None:
            if kwargs.get('stdin') is not None:
                raise ValueError('stdin and input arguments may not both be used.')
            kwargs['stdin'] = PIPE
    
        if capture_output:
            if kwargs.get('stdout') is not None or kwargs.get('stderr') is not None:
                raise ValueError('stdout and stderr arguments may not be used '
                                 'with capture_output.')
            kwargs['stdout'] = PIPE
            kwargs['stderr'] = PIPE
    
        with Popen(*popenargs, **kwargs) as process:
            try:
                stdout, stderr = process.communicate(input, timeout=timeout)
            except TimeoutExpired as exc:
                process.kill()
                if _mswindows:
                    # Windows accumulates the output in a single blocking
                    # read() call run on child threads, with the timeout
                    # being done in a join() on those threads.  communicate()
                    # _after_ kill() is required to collect that and add it
                    # to the exception.
                    exc.stdout, exc.stderr = process.communicate()
                else:
                    # POSIX _communicate already populated the output so
                    # far into the TimeoutExpired exception.
                    process.wait()
                raise
            except:  # Including KeyboardInterrupt, communicate handled that.
                process.kill()
                # We don't call process.wait() as .__exit__ does that for us.
                raise
            retcode = process.poll()
            if check and retcode:
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['ty', 'check']' returned non-zero exit status 1.

/usr/lib/python3.12/subprocess.py:571: CalledProcessError
----------------------------- Captured stdout call -----------------------------
error[invalid-assignment]: Object of type `dict[str, str]` is not assignable to attribute `headers` of type `CaseInsensitiveDict[str]`
  --> bugwarrior/services/azuredevops.py:53:9
   |
53 |         self.session.headers = {
   |         ^^^^^^^^^^^^^^^^^^^^
   |

error[invalid-argument-type]: Argument to `StringIO.__init__` is incorrect
   --> bugwarrior/services/trac.py:161:61
    |
161 |             tickets = list(csv.DictReader(StringIO.StringIO(text.encode('utf-8'))))
    |                                                             ^^^^^^^^^^^^^^^^^^^^ Expected `str | None`, found `bytes`
    |
info: Method defined here
   --> stdlib/_io.pyi:824:9
    |
824 |     def __init__(self, initial_value: str | None = "", newline: str | None = "\n") -> None: ...
    |         ^^^^^^^^       ------------------------------ Parameter declared here
    |
 ```

File: bugwarrior/services/taiga.py
Comment (ryneeverett) [inline bugwarrior/services/taiga.py:1]: This ought to be a one-line change by adding the field to the dict construction rather than mutating it:

```py
'due': self.parse_date(self.record.get('due_date'))
```

(`.get()` defaults to `None` which `parse_date` then returns as `None`)

ryneeverett (changes requested):

ryneeverett:
Tests should be fixed if you rebase.

Fubukimaru:
It should be ok now. Thanks for the support @ryneeverett !

Let me know if there's anything missing. Also, sorry for the big amount of commits 😅. I suppose they'll be squashed.

bugwarrior/services/taiga.py
Fubukimaru: Change applied.

ryneeverett (approved):

PR #1212, opened 2026-05-19
fix typing errors coming from requests==2.34.2

Lotram:
Fix typing issue related to https://github.com/GothenburgBitFactory/bugwarrior/pull/1211

The error appeared with recent version of `requests`, which has better typing

One related question: why don't we commit the `uv.lock` file ? bugwarrior is a standalone CLI, not a shared library, so is there a good reason not to pin all dependencies for users ? Right now, if a dependency upgrade breaks something, it's not easily detectable. We could also pin the dependencies in pyproject.toml, but it's not as easy to upgrade as a single `uv lock --upgrade`

### DIFF
--- a/bugwarrior/services/azuredevops.py
+++ b/bugwarrior/services/azuredevops.py
@@ -50,11 +50,13 @@ class AzureDevopsClient(Client):
         self.host = host
         self.base_url = f"https://{host}/{org}/{project}/_apis/wit"
         self.session = requests.Session()
-        self.session.headers = {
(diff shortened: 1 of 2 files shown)


ryneeverett (approved):

ryneeverett:
> why don't we commit the `uv.lock` file ?

Committing the lockfile opens the question of how the lockfile gets updated without constant dependabot noise. Maybe a monthly CI job?

PR #1213, opened 2026-05-19
removed unused hamming distance functions

Lotram:
I simplified the code related to merging tags and annotations. 

In particular, the hamming distance code  was unused. 

While I think it makes sense to keep the "normalize_annotation" function (even though I don't see why the annotation would slightly change between calls to services), I'm not sure the "compare only length of shortest string" behavior brings anything useful (but I  kept it  for now).

### DIFF
--- a/bugwarrior/db.py
+++ b/bugwarrior/db.py
@@ -13,6 +13,7 @@ from bugwarrior.collect import get_service
 from bugwarrior.notifications import send_notification
 
 if TYPE_CHECKING:
+    from bugwarrior.config.schema import MainSectionConfig
     from bugwarrior.config.validation import Config
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


bugwarrior/db.py
```diff
@@ -155,99 +137,36 @@ def find_taskwarrior_uuid(
...
-        truncating to maximum length, and comparing hamming distances.
-        Useful generally only for annotations.
 
-    """
+def merge_tags(
+    main_conf: "MainSectionConfig", local: dict[str, Any], remote: dict[str, Any]
+) -> list[str]:
+    task_tags: set[str] = set(local.get("tags", []))
+    if main_conf.replace_tags:
+        task_tags &= set(main_conf.static_tags)
```
ryneeverett, line 159: Shouldn't this be a replacement rather than the intersection?

```suggestion
        task_tags = set(main_conf.static_tags)
```

ryneeverett:
> While I think it makes sense to keep the "normalize_annotation" function (even though I don't see why the annotation would slightly change between calls to services), I'm not sure the "compare only length of shortest string" behavior brings anything useful (but I kept it for now).

Agreed. I'm supportive of intentionally removing that behavior, I just wanted to make sure we didn't do so on accident.

bugwarrior/db.py
ryneeverett: Did you overlook this comment?

Lotram: Ah yes sorry. Actually, the `&=` is right, here: In case we use the options `replace_tags`, we want to keep the tags given in `static_tags`, **but only the ones that already exist**. If items from `static_tags` aren't in the local task, then they shouldn't be copied (hence the `&=` and not `=`)

The related comment was:
`# Delete all items in local_task, unless they are in keep_items`

ryneeverett: My mistake, I had forgotten how `static_tags` works.

ryneeverett (approved):

PR #1216, opened 2026-06-03
Simplify get_tags_from_labels

Lotram:
A few services were not using the default parameter names for the `get_tags_from_labels` (pagure and youtrack)

I changed the config param names for those service, while allowing multiple aliases, to ensure backward compatibility. 

This allows for a much simpler `get_tags_from_labels`, only parameters being the labels and the config. 
We could even consider moving this function outside the `Service` class, since it does not depend on the Service subclass (or this could be a `ServiceConfig` method instead).

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -134,6 +134,16 @@ class Issue(abc.ABC):
         """
         raise NotImplementedError()
 
+    def render_tags_from_labels(self, labels: list[str]) -> list[str]:
+        """Transform labels into suitable taskwarrior tags using the label template."""
+
+        return [
+            Template(self.config.label_template).render(
+                {**self.record, "label": re.sub(r'[^a-zA-Z0-9]', '_', label)}
+            )
+            for label in labels
+        ]
+
     def get_tags_from_labels(
         self,
         labels: list[str],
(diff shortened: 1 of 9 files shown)


bugwarrior/services/__init__.py
```diff
@@ -134,37 +134,22 @@ def get_default_description(self) -> str:
...
+                {**self.record, "label": re.sub(r'[^a-zA-Z0-9]', '_', label)}
+            )
+            for label in labels
+        ]
 
-        for label in labels:
-            normalized_label = re.sub(r'[^a-zA-Z0-9]', '_', label)
-            context.update({template_variable: normalized_label})
-            tags.append(label_template.render(context))
+    def get_tags_from_labels(self, labels: list[str]) -> list[str]:
```
ryneeverett: Can we retain the parameters from the function signature and log a warning if they're changed from the default, like the timezone parameter in `parse_date`?

ryneeverett (commented):
- I don't see much point in adding deprecation notices to the documentation. If somebody isn't using the deprecated options then they don't need to know about them. If they are using them then they'll get the exact same information from the log warnings.
- We should add an entry to the api.rst changelog mentioning the addition of `render_tags_from_labels`. (I'm inclined to think we shouldn't mention deprecated parameters in that changelog. Since deprecations don't necessarily require a version bump, there wouldn't necessarily be any place to log them so we shouldn't set that precedent.)

ryneeverett (approved):

Issue #1160, opened 2025-12-11
Consider full switch to Pytest


Lotram:
Disclaimer: I'm biased towards pytest, it's used in most python projects, and I think it has been the standard in python for quite some time. 

That said, I truly think the codebase would benefit from migrating to pytest
* we already use pytest to run the tests anway, so it wouldn't add a dependency.
* Simpler assertions: No more `self.assertEqual(...)`, a simple `assert a = b` is ok.
* we already have a few pytest fixtures, it's more powerful and (IMO) readable than setUp / teardown methods. I admit that this is the most biased argument I have, some people may disagree on fixtures being better than factories / setup functions. 
* pytest does not require to use classes, simple functions are fine, which is enough for simple tests (we can still use classes for bigger ones)
* parametrize avoids a lot of redundancy
* a lot of plugins are available, if we ever need any (let's say to run tests in parallel, or to randomize test order for example)

ryneeverett:
I think my fundamental question is, which of these benefits do we have to switch completely to pytest to take advantage of? For example, do we need to switch completely over to use the parallelization/randomization plugins?

Lotram:
Regarding plugins, I guess it depends on their actual implementation, as well as unittest.TestCase. I don't know how well pytest fixtures work with TestCases. 

"Switching completely to Pytest" means 2 things IMO:
1. simplifying assertions. 
2. Removing unitest.TestCase inheritance. That does not mean we have to remove all setUp / tearDown methods immediately, or migrate everything to fixtures, but we can then use test functions instead of always classes, and implement setup as we prefer. 

So IMO the code would end up more readable, and more flexible. 

We can definitely do 1. without 2. at first.

ryneeverett:
Personally, I find the readability of pytest assertions marginally better at best. If we wanted a pytest feature that depended on converting all the assertions and TestCase inheritance, I'd be supportive.

As far as I know, there's nothing stopping you from writing new tests in test functions rather than classes. So I don't see how rewriting the existing tests gives us any more flexibility.

My perspective is that there is a cost to intentionally get further locked into pytest, even if it has dominated the field for the last decade. It's a very small cost, but there has to be some concrete benefit to justify it.

Lotram:
> As far as I know, there's nothing stopping you from writing new tests in test functions rather than classes.

It would mean using regular `assert` instead of `self.assertAnything`. While I do prefer that, I also think convention consistency is important, so I think we should stick to only one or the other. 

[...]

PR #1219, opened 2026-06-09
Test improvements

Lotram:
Several improvements on tests: 
* Better test isolation
    * All non-service tests now rely on fake service, config and issue objects
    * tests go through config validation only when this is actually needed, they instantiate config object directly otherwise
* `TestTemplates` inherits from `ConfigTest` instead of `ServiceTest`, as it does not need anything from `ServiceTest
* Move all service-related tests to dedicated services/ folder, to mirror source code
* remove the `add_response`  static method, it became useless because of parameter deprecation. This allowed for simpler inheritance

This changes will also make #1134 and #1160 easier I think. 

I tried to write several independent commits, I can create multiple PRs if that's easier.


tests/config/test_validation.py
```diff
@@ -62,12 +70,15 @@ def test_target_section_missing(self):
...
-            "  <- No [my_service] section found"
+            "[general]\ntargets = ['my_service']  <- No [my_service] section found"
         )
 
     def test_service_missing(self):
+        # A second valid service is needed so the section with the missing
+        # 'service' option fails discrimination against a non-empty union,
+        # which is what produces the "No option 'service'" message.
+        self.config["general"]["targets"] = ["my_service", "other_service"]
+        self.config["other_service"] = {"service": "test", "username": "ralph"}
```
ryneeverett: I don't really follow the comment, but if there is only one target and it's section is missing the service field, shouldn't that also trigger this message? Are we covering over a bug here?

tests/config/test_validation.py
```diff
@@ -88,60 +99,49 @@ def test_root_validator(self):
...
-            'my_kan',
-            'my_gitlab',
-            'my_redmine',
-        ]
-        self.config['my_redmine'] = {
-            'service': 'redmine',
-            'url': 'https://example.com',
-            'key': 'mykey',
-        }
         self.validate()
```
ryneeverett: Could drop this validation since we're not modifying the config before it and the unmodified fake config is already tested by `test_valid`.

```suggestion
```

tests/config/test_validation.py
```diff
@@ -151,6 +151,39 @@ def test_quoted_flavor_key_error(self):
...
+    def test_notifications_invalid_backend(self):
+        self.config['notifications'] = {'backend': 'invalid_backend'}
+        self.assertValidationError(
+            "[notifications]\nbackend = invalid_backend  <- Input should be "
+            "'gobject', 'growlnotify' or 'applescript'"
+        )
+
+    def test_service_and_hooks_errors_reported_together(self):
+        # See test_service_missing: a second valid service is needed for the
+        # "No option 'service'" message.
```
ryneeverett: Note to Self: I need to understand this per my above comment.

tests/base.py
```diff
@@ -22,48 +22,81 @@ class DumbConfig(config.ServiceConfig):
...
+                f"Configured service '{name}' not found. "
+                "Is it installed? Or misspelled?"
+            )
+
+    with contextlib.ExitStack() as stack:
+        for module in _GET_SERVICE_MODULES:
+            stack.enter_context(
+                unittest.mock.patch(f'{module}.get_service', fake_get_service)
+            )
+        yield mapping
```
ryneeverett: I'm not seeing a with/as statement that uses this anywhere. Couldn't it be a bare `yield`?

```suggestion
        yield
```

tests/base.py
```diff
@@ -99,6 +132,16 @@ def tearDown(self):
     def inject_fixtures(self, caplog):
         self.caplog = caplog
 
+    def enter_context(self, cm):
```
ryneeverett: Why not name this `enterContext` so we can just delete it when we drop 3.10 support?

If you'd prefer, I also wouldn't mind dropping python-3.10 support now. It seems pretty likely that we won't even cut a release by the time it EOL's in October.

I also wonder what the advantage is of `enterContext` over decorators or `with`? Is it that you find it more visible when it's in the function body (unlike decorators) but don't like the indentation of `with`?

tests/test_command.py
```diff
@@ -1,38 +1,78 @@
...
+
+def yields_none(self):
+    return iter([])
+
+
+def raises(self):
+    raise Exception('message')
+
+
+def yields_per_target(self):
```
ryneeverett: I find this name confusing because it implies the function will yield more than once, but I don't have a better suggestion.

Maybe I'm missing something but is a distinction between `yields_one` and `yields_per_target` important? That is, couldn't we just rename this `yields_one`?

tests/test_command.py
```diff
@@ -178,6 +207,9 @@ def test_locked_repository(self, file_lock):
         """
         A locked task repository should abort the pull.
         """
+        # The service is never collected (the lock fails first), but config
+        # loading still resolves it.
```
ryneeverett: I do not understand this comment -- "resolves it". What is "it"?

tests/test_db.py
```diff
@@ -61,20 +63,16 @@ def test_handles_missing_tags(self):
...
-            },
-            'my_service': {
-                'service': 'github',
-                'login': 'ralphbean',
-                'username': 'ralphbean',
-                'token': 'abc123',
-            },
-        }
-        self.bwconfig = self.validate()
+        # synchronize() resolves the service via get_service for UNIQUE_KEY/UDAS.
```
ryneeverett: I'm not following the point of this comment -- as a person reading the test, why do I want to know this right now?

Lotram: This comment triesto explain why we need to mock `get_service` in synchronize. but you're right, we can remove it.

tests/test_command.py
Lotram: "it" is the service: even though the lock fails, `get_service` is called, so we need to mock the service using `register_services`. 

is this clearer ? 
```
        # The service is never collected (the lock fails first), but `get_service` is called before failure, we need to mock it
````

File: tests/base.py (hunk 4 above)
Comment (Lotram) [reply tests/base.py]: Done

File: tests/config/test_validation.py (hunk 2 above)
Comment (Lotram) [reply tests/config/test_validation.py]: Done

File: tests/config/test_validation.py (hunk 1 above)
Comment (Lotram) [reply tests/config/test_validation.py]: I'm adding a new test for the case where we only  have this config. it's related to 
```
    # Default to generic ServiceConfig if no service defined, mostly for tests
    if not service_config_classes:
        return ServiceConfig
```
in the `get_service_config_union_type` function

File: tests/base.py (hunk 5 above)
Comment (Lotram) [reply tests/base.py]: > I also wonder what the advantage is of enterContext over decorators or with? Is it that you find it more visible when it's in the function body (unlike decorators) but don't like the indentation of with?
We can't simply use context managers inside `setUp`, that's the whole point of `enterContext`. the `__exit__` function is automatically called during teardown. 

I kept it in simple tests too, but I can easily replace it there by a with statement if you prefer. 

decorators would be more complex here, as the services are defined dynamically IIRC. 

My plan is to replace these by a fixture when I finally migrate to pytest anyway. 

> If you'd prefer, I also wouldn't mind dropping python-3.10 support now. It seems pretty likely that we won't even cut a release by the time it EOL's in October.

ok I can drop python 3.10, so we can also simplify some imports.

Comment (ryneeverett) [reply tests/base.py]: Ah, I had missed the fact that this was used in `setUp` methods. That makes sense now.

File: tests/config/test_validation.py (hunk 1 above)
Comment (ryneeverett) [reply tests/config/test_validation.py]: The error messages in the output of that new test seems nonsensical and unhelpful. It isn't obvious to me why falling back to the base class causes such bad behavior -- shouldn't there still be a "No option 'service'" message since `service` is a required field in `ServiceConfig`. Regardless, can we agree that this is buggy behavior that could be improved in validation.py? (We can open a new issue for this if you prefer.)

File: tests/test_command.py (hunk 7 above)
Comment (ryneeverett) [reply tests/test_command.py]: I'm not really sure, but now that I've read both versions they both make sense! However, if somebody is reading this test and confused about it, I suspect it is easier to delete it and run the test to see what happens than to read an explanation. I'll leave it to your discretion but pleas move the comment down with the moved `register_services` call if you do decide to keep it.

Comment (Lotram) [reply tests/test_command.py]: I removed the comment

File: tests/config/test_validation.py (hunk 1 above)
Comment (Lotram) [reply tests/config/test_validation.py]: You're right, I modified the function in validation.py so it handles the empty case properly.

File: tests/config/test_validation.py [hunk 9]
```diff
@@ -151,6 +144,37 @@ def test_quoted_flavor_key_error(self):
...
+    def test_notifications_invalid_backend(self):
+        self.config['notifications'] = {'backend': 'invalid_backend'}
+        self.assertValidationError(
+            "[notifications]\nbackend = invalid_backend  <- Input should be "
+            "'gobject', 'growlnotify' or 'applescript'"
+        )
+
+    def test_service_and_hooks_errors_reported_together(self):
+        self.config['general']['targets'] = ['my_service', 'other_service']
+        self.config['other_service'] = {'service': 'test', 'username': 'ralph'}
```
ryneeverett: We can delete these lines now since we don't need to add a second service to get the correct validation message.

```suggestion
```

bugwarrior/config/validation.py
ryneeverett, line 1: This does seem to improve the outcome as shown in the tests, but can you explain it? I take it that the problem had something to do with the early `return` without the descriminator `Field`? But what is the fundamental difference between `_MissingServiceDiscriminator` and `ServiceConfig`?

Lotram: It's related to `_format_service_error` error handling. When going through the `ServiceConfig` block, 
1. We have an error   where  `loc = (0, "service"), type = "missing"` whereas the block we want in `_format_service_error` is `if len(loc) == 1`
2. Even if we change the condition to handle this, we would also get  a `loc = (0, "username")   type = "extra_forbidden"` error, because `ServiceConfig` does not know the service-specific fields. When using a discriminator field, if the discriminator raises an error, pydantic stop there and does not go through other errors. 

This can also be modified in the error handling function, though I thought it was simpler to simply have an empty discriminator class

ryneeverett: I think I basically follow that explanation, but how does `_MissingServiceDiscriminator` avoid those issues?

Lotram: We need to have discriminated union to be as close as possible to the general case (when we have valid configs), for the reasons above. 

so I guess your question is: why don't we fallback on  `service_config_classes = ServiceConfig` instead of `service_config_classes = _MissingServiceDiscriminator`, when there is no valid service config ? 

Simply because a discriminator requires the field (`service` here), to be a `Literal`, not a `str`

To validate models based on that information, you can set a common field on each model of the union (pet_type in the example below), typed as accepting one or multiple literal values. 

[The doc](https://pydantic.dev/docs/validation/latest/concepts/unions/#discriminated-unions-with-string-discriminators) says:
> To validate models based on that information, you can set a common field on each model of the union typed as accepting one or multiple literal values.

ryneeverett: Thanks, I get it now! I opened https://github.com/Lotram/bugwarrior/pull/2 as a proposed alternative which i find more readable and possibly more correct since `service` can't actually be an arbitrary `str`.

Lotram: I can see a problem with your implementation: ` Literal["__bugwarrior_service_placeholder__"]` (type from the base class) is not compatible with `Literal["github"]`, so there is a theoretical typing problem. 

I agree `service` can't be an arbitrary string, it should be the Union of all declared service types, but since this registration is dynamic (and we don't want to register unused services), we have a problem. So IMO, `str` is "truer" than ` Literal["__bugwarrior_service_placeholder__"]`

Here is a minimal example to reproduce:
```
from typing import Literal

from pydantic import BaseModel


class _String(BaseModel):
    service: str


class _Literal(BaseModel):
    service: Literal["_error_"]


class Config(_Literal):
    pass


class GitHubConfig(Config):
    service: Literal["github"]


class GitlabConfig(Config):
    service: Literal["gitlab"]
```

* ty is ok
* both mypy and pyright complain about the type (mypy error is: `Incompatible types in assignment (expression has type "Literal['github']", base class "Config" defined the type as "Literal['_error_']")`)

Note that pyright would still complain with the regular `str` instead of `Literal`, because "Variable is mutable so its type is invariant" (and it still complains when using `frozen=True` and `ClassVar), pyright complains _a lot_


So, as a conclusion, I'm ok with your implementation if you prefer it, for simplicity sake, but I think it's just a tiny bit more wrong

ryneeverett: No, let's stick with the more correct implementation. My only ask would be, please add at least a sentence of explanation to the `_MissingServiceDiscriminator` docstring about how it's essential purpose is to make the `service` field a `Literal` so it will work as a discriminator. I think using the `ServiceConfig` base class as a fallback here is the intuitive implementation -- after all it's what you did initially -- so this is a natural place for anyone to wonder why this class is necessary.

ryneeverett (approved):

PR #1220, opened 2026-06-17
Drop python 310

Lotram:
Also upgrade ty to latest version (0.49)

About `exclude-newer`:  There are more and more attacks targeting package dependencies, it's now consider a good practice to add a cooldown to reduce the risk. it supports friendly timedeltas (e.g. "3 days") since 0.9.17, released 2025-12-09. It wouldn't work with older uv versions. Is that acceptable to you, or should I drop the commit for now ?

### DIFF
--- a/.github/workflows/bugwarrior.yml
+++ b/.github/workflows/bugwarrior.yml
@@ -41,7 +41,9 @@ jobs:
 
           python3 -m venv .venv
           source .venv/bin/activate
-          pip install --upgrade "pip>=25.1" setuptools
+          pip install --upgrade "pip>=26.1"
+          pip config --site set install.uploaded-prior-to P3D
+          pip install --upgrade setuptools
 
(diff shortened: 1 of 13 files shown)


ryneeverett:
I'm all for the cooldown, I only wish it were standardized across tooling so pip could do the same.

Lotram:
pip actually has an option for this too, since the recent [26.1](https://pip.pypa.io/en/stable/news/#v26-1) release: 
> Allow --uploaded-prior-to to accept a duration in days (e.g., P3D for 3 days ago). ([#13674](https://github.com/pypa/pip/issues/13674))

ryneeverett:
Maybe we should add that flag to the pip CI test and documentation for consistency? It seems preferable that dependency breakages hit all workflows at the same time.

ryneeverett (approved):

PR #1225, opened 2026-07-10
Full migration to idiomatic pytest

### LINKED ISSUE #1160: Consider full switch to Pytest

Lotram:
Last PR to close #1160 

This PR mostly does 2 things: 
1. migrate everything to more standard pytest (fixtures, parametrize...)
2. Improve consistency across service test files.

### Standard Pytest
This change is quite opinionated, but here is why I think it improves the code:
* Remove *a lot* of silent mutations. The tests used to rely on manual copies, setting / unsetting environment variable etc to make everything works. Tests are now isolated by construction. We use standard fixture for tmp paths, env variable overrides, and we define our own to create services, issues etc.
* composition over inheritance. Standard principle too, we used to rely a lot on test class inheritance, which is quite fragile, and prevented us to easily modify tests (hence this big PR)
* warnings are now considered as errors, so we can detect and fix them (except for taskw deprecation warnings, as we can't really do anything about that for now). Several bugs were detected and fixed thanks to that. 

### Service tests consistency
Each service test was defining and using its data differently, once again making it harder to add a change for all services at once, I tried to make them a bit more similar. I added a few tests to ensure basic consistency

I tried to create atomic commits to ease the review. If that's still too much, I can create multiple PRs

### DIFF
--- a/tests/config/test_data.py
+++ b/tests/config/test_data.py
@@ -9,8 +9,7 @@ from bugwarrior.config import data, schema
 
 def assert_0600(bw_data):
     permissions = oct(os.stat(bw_data._datafile).st_mode & 0o777)
-    # python2 -> 0600, python3 -> 0o600
-    assert permissions in ['0600', '0o600']
+    assert permissions == '0o600'
 
 
 class TestData:
--- a/tests/config/test_data.py
+++ b/tests/config/test_data.py
@@ -1,72 +1,74 @@
 import json
 import os
+from pathlib import Path
+
+import pytest
 
 from bugwarrior.config import data, schema
 
-from ..base import ConfigTest
+
+def assert_0600(bw_data):
+    permissions = oct(os.stat(bw_data._datafile).st_mode & 0o777)
+    # python2 -> 0600, python3 -> 0o600
+    assert permissions in ['0600', '0o600']
 
 
-class TestData(ConfigTest):
-    def setUp(self):
-        super().setUp()
-        self.data = data.BugwarriorData(self.lists_path)
+class TestData:
+    @pytest.fixture
+    def bw_data(self, config_environment):
+        return data.BugwarriorData(config_environment.lists_path)
 
-    def assert0600(self):
-        permissions = oct(os.stat(self.data._datafile).st_mode & 0o777)
-        # python2 -> 0600, python3 -> 0o600
-        assert permissions in ['0600', '0o600']
-
-    def test_get_set(self):
+    def test_get_set(self, bw_data):
         # "touch" data file.
-        with open(self.data._datafile, 'w+') as handle:
+        with open(bw_data._datafile, 'w+') as handle:
             json.dump({'old': 'stuff'}, handle)
 
-        self.data.set('key', 'value')
+        bw_data.set('key', 'value')
 
-        assert self.data.get('key') == 'value'
-        assert self.data.get_data() == {'old': 'stuff', 'key': 'value'}
-        self.assert0600()
+        assert bw_data.get('key') == 'value'
+        assert bw_data.get_data() == {'old': 'stuff', 'key': 'value'}
+        assert_0600(bw_data)
 
-    def test_set_first_time(self):
-        self.data.set('key', 'value')
+    def test_set_first_time(self, bw_data):
+        bw_data.set('key', 'value')
 
-        assert self.data.get('key') == 'value'
-        self.assert0600()
+        assert bw_data.get('key') == 'value'
+        assert_0600(bw_data)
 
-    def test_path_attribute(self):
-        assert self.data.path == self.lists_path
+    def test_path_attribute(self, bw_data, config_environment):
+        assert bw_data.path == config_environment.lists_path
 
 
-class TestGetDataPath(ConfigTest):
-    def setUp(self):
-        super().setUp()
-        self.main_config = schema.MainSectionConfig(targets=[])
+class TestGetDataPath:
+    @pytest.fixture
+    def main_config(self):
+        return schema.MainSectionConfig(targets=[])
 
-    def assertDataPath(self, expected_datapath):
-        assert expected_datapath == data.get_data_path(self.main_config.taskrc)
+    def assert_data_path(self, main_config, expected_datapath):
+        assert str(expected_datapath) == data.get_data_path(main_config.taskrc)
 
-    def test_TASKDATA(self):
+    def test_TASKDATA(self, main_config, monkeypatch, tmp_path):
         """
         TASKDATA should be respected, even when taskrc's data.location is set.
         """
-        datapath = os.environ['TASKDATA'] = os.path.join(self.tempdir, 'data')
-        self.assertDataPath(datapath)
+        datapath = tmp_path / 'data'
+        monkeypatch.setenv('TASKDATA', str(datapath))
+        self.assert_data_path(main_config, datapath)
 
-    def test_taskrc_datalocation(self):
+    def test_taskrc_datalocation(self, main_config, config_environment):
         """
         When TASKDATA is not set, data.location in taskrc should be respected.
         """
         assert 'TASKDATA' not in os.environ
-        self.assertDataPath(self.lists_path)
+        self.assert_data_path(main_config, config_environment.lists_path)
 
-    def test_unassigned(self):
+    def test_unassigned(self, main_config, config_environment):
         """
         When data path is not assigned, use default location.
         """
         # Empty taskrc.
-        with open(self.taskrc, 'w'):
-            pass
+        config_environment.taskrc.write_text('')
 
         assert 'TASKDATA' not in os.environ
 
-        self.assertDataPath(os.path.expanduser('~/.task'))
+        self.assert_data_path(main_config, Path.home() / '.task')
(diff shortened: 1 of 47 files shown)


tests/services/test_gerrit.py
```diff
@@ -76,10 +85,10 @@ def test_to_taskwarrior(self):
 
         assert actual == expected
 
-    def test_work_in_progress(self):
-        wip_record = dict(self.record)  # make a copy of the dict
+    def test_work_in_progress(self, service, record, extra):
+        wip_record = dict(record)  # make a copy of the dict
```
ryneeverett: This copying is no longer necessary with proper fixtures, right?

tests/config/test_data.py
```diff
@@ -1,72 +1,74 @@
...
+import pytest
 
 from bugwarrior.config import data, schema
 
-from ..base import ConfigTest
 
+def assert_0600(bw_data):
+    permissions = oct(os.stat(bw_data._datafile).st_mode & 0o777)
+    # python2 -> 0600, python3 -> 0o600
+    assert permissions in ['0600', '0o600']
```
ryneeverett: Just noticing this can be simplified since we don't support python2.

```suggestion
    assert permissions == '0o600'
```

tests/services/test_github.py
```diff
@@ -1,120 +1,125 @@
...
+        'namespace': 'arbitrary_username',
     }
 
-    def test_draft(self):
-        service = self.get_mock_service(GithubService)
-        draft = dict(ARBITRARY_ISSUE)
+
+class TestGithubIssue:
+    def test_draft(self, service, record, extra):
+        draft = dict(record)
```
ryneeverett: This copying is no longer necessary with proper fixtures, right?

tests/services/test_github.py
```diff
@@ -1,120 +1,125 @@
...
+        'number': 10,
+        'body': 'Something',
+        'user': {'login': 'arbitrary_login'},
+        'milestone': {'title': 'alpha'},
+        'labels': [{'name': 'bugfix'}],
+        'created_at': CREATED.isoformat(),
+        'closed_at': CLOSED.isoformat(),
+        'updated_at': UPDATED.isoformat(),
+        'repo': 'arbitrary_username/arbitrary_repo',
+        'state': 'CLOSED',
```
ryneeverett: I'm guessing uppercasing this was an accident.

```suggestion
        'state': 'closed',
```

tests/services/test_gitlab.py
```diff
@@ -378,131 +405,110 @@ def test_get_repos(self):
...
-        assert result == [self.data.arbitrary_project]
+        assert result == [project]
 
-        result = self.client.get_repos(
+        # A repo that cannot be fetched is skipped rather than included.
+        result = client.get_repos(
             include_repos=['non_existing'], only_membership=False, only_owned=False
         )
-        pytest.raises(OSError)
+        assert result == []
```
ryneeverett, line 456: Why did behavior change here?

tests/services/test_gitlab.py
```diff
@@ -1101,13 +1086,13 @@ def test_only_if_assigned_user_lookup(self):
         overrides = {'only_if_assigned': 'jack_smith'}
 
         # Should not raise an error
-        service = self.get_mock_service(GitlabService, config_overrides=overrides)
+        service = make_service(**overrides)
 
-        # Verify service was created successfully
+        # Verify service was CREATED successfully
```
ryneeverett:
```suggestion
        # Verify service was created successfully
```

tests/services/test_jira.py
```diff
@@ -2,32 +2,88 @@
...
+            'namedfield': {'valueinside': ARBITRARY_NAMEDFIELD_VALUEINSIDE},
+        },
+        'key': '%s-%s' % (PROJECT, ARBITRARY_ID),
+    }
+
+
+@pytest.fixture
+def record_with_due(record):
+    record_with_due = record.copy()
+    record_with_due['fields'] = record_with_due['fields'].copy()
```
ryneeverett: Neither of these copies are necessary with proper fixtures, right? This pattern is repeated several times in this file.

tests/test_templates.py
```diff
@@ -1,90 +1,83 @@
...
-            {'description': self.arbitrary_default_description, 'tags': []}
-        )
 
-        assert record == expected_record
+class TestTemplates:
+    def test_default_taskwarrior_record(self, get_issue, record):
+        issue = get_issue({})
+
+        actual = TaskConstructor(issue).get_taskwarrior_record()
+        expected_record = record.copy()
```
ryneeverett: This copying is no longer necessary with proper fixtures, right?

tests/test_templates.py
```diff
@@ -1,90 +1,83 @@
...
+    def test_override_project(self, get_issue, record):
         project_template = "wat_{{ project|upper }}"
 
-        issue = self.get_issue({'project': project_template})
+        issue = get_issue({'project': project_template})
 
-        record = TaskConstructor(issue).get_taskwarrior_record()
-        expected_record = self.arbitrary_issue.copy()
+        actual = TaskConstructor(issue).get_taskwarrior_record()
+        expected_record = record.copy()
```
ryneeverett: This copying is no longer necessary with proper fixtures, right?

ryneeverett (commented):
- FYI, I'm thinking I'll squash merge this one. The commit structure was helpful for review but I'm not sure how helpful it would be in a `git blame` scenario. I also noticed that the commits do not appear to all be atomic in the sense that the tests pass -- f825c0ccf2dbf4464665eaedbaa04f97d9424aa2 removes ConfigTest which is still depended on.
- Please update the tutorial.rst Tests section to reflect the new pattern.
- This is more an idea for the future than something I'd want added to this PR, but would is make sense to make `config` a shared fixture across services? My thought is that is might be confusing that the default differs from `SERVICE_CONFIG`. It would require some reworking in a lot of the services, but it seems nice if they all used a fixture like Trello:

```py
    @pytest.fixture
    def config(self):
        return {'general': {'targets': ['myservice']}, 'myservice': {**SERVICE_CONFIG}}
```

Lotram:
>  FYI, I'm thinking I'll squash merge this one. The commit structure was helpful for review but I'm not sure how helpful it would be in a git blame scenario. I also noticed that the commits do not appear to all be atomic in the sense that the tests pass -- https://github.com/GothenburgBitFactory/bugwarrior/commit/f825c0ccf2dbf4464665eaedbaa04f97d9424aa2 removes ConfigTest which is still depended on.

Agreed, I created multiple commits to avoid having a single really big diff, but we can definitely squash them before merging. Do you want me to do it now, or wait ? 

> Please update the tutorial.rst Tests section to reflect the new pattern.
Sure 

> This is more an idea for the future than something I'd want added to this PR, but would is make sense to make config a shared fixture across services? My thought is that is might be confusing that the default differs from SERVICE_CONFIG. It would require some reworking in a lot of the services, but it seems nice if they all used a fixture like Trello

Not sure I understand how you want to pass the `SERVICE_CONFIG` dict to a shared fixture. Through a factory maybe ? Something like:
```
@pytest.fixture
def config_factory(service_config):
    def _config():
        return {'general': {'targets': ['myservice']}, 'myservice': {**service_config}}
    return _config

# in test_github.py
SERVICE_CONFIG = {}
@pytest.fixture
def config():
    return config_factory(SERVICE_CONFIG)
```

tests/services/test_github.py
Lotram: Indeed, I used uppercased variables for this at some point, before reverting my change

tests/services/test_gitlab.py
Lotram: There was actually a bug in the test, before my change.  `pytest.raises(OSError)` was called but never used as a context manager (`with pytest.raises(OSError):`), so it was a no-op that didn't actually assert anything

The test was simply asserting nothing.

Same thing before the assert style migration, `self.assertRaises(OSError)` was not doing anything either

Lotram:
@ryneeverett I added a commit to speed up the documentation tests (as well as the actual doc generation).

ryneeverett:
I'll squash at the end since the github UI makes it super easy, I just wanted to get your opinion. I hadn't been squashing your PR's lately since the commits have been well-structured and truly atomic.

> Not sure I understand how you want to pass the `SERVICE_CONFIG` dict to a shared fixture.

It seems like we could use the [request](https://docs.pytest.org/en/stable/reference/reference.html#request) fixture and do something like:

```py
    @pytest.fixture
    def config(self, request):
        return {'general': {'targets': ['myservice']}, 'myservice': {**request.module.SERVICE_CONFIG}}
```

ryneeverett (approved):

PR #1226, opened 2026-07-22
Typed task model

Lotram:
This PR is a suggestion of data models refactoring.

### Why ? 
1. Using our own Task model decreases the dependency to taskw, and improves typing / readability, instead of using plain dicts. Having our own abstraction simplifies the changes needed to switch the dependency from  taskw to taskchampion.
2. The `Issue` (and child classes) becomes less useful, it does not hold any specific logic, I suggest an implementation where it's merged into Service. 
3. This gives us a clearer split between data coming from external services (`Service`, record) and data inserted in taskwarrior

### What's inside this PR
1. The `Task` and `Udas` models
2. Use these models inside `Service` / `Issue`, so we use that instead of plain dicts (in `db.py`, `collect.py` etc)
4. Merge `Issue` inside `Service`, as there is no clear difference between the two
5. Move several methods (the ones where `self` is only used to access the config) out of `Service`

Commits 3 and 4 are optional, I think they're improvements, but not necessary.
I only modified Linear to demonstrate this.
The idea is not to review it in details for now, there are probably some bugs, but to discuss the overall shape of the changes.

### What's next ? 
If we proceed with this changes, I would then use Task even more in `db.py`, and try avoid using dicts. 
We could also replace calls to `taskw.task.FIELDS` by our own `Task` object

### DIFF
--- a/bugwarrior/services/__init__.py
+++ b/bugwarrior/services/__init__.py
@@ -72,6 +72,163 @@ def get_processed_url(main_config: schema.MainSectionConfig, url: str) -> str:
     return url
 
 
+def get_secret(
+    config: schema.ServiceConfig, key: str, login: str = 'nousername'
+) -> str:
+    """Get a secret value, potentially from an :ref:`oracle <Secret Management>`.
+
+    The secret key need not be a *password*, per se.
+
+    :param `key`: Name of the configuration field of the given secret.
+    :param `login`: Username associated with the password in a keyring, if
+        applicable.
+    """
+    password = getattr(config, key)
+    if not password or password.startswith("@oracle:"):
+        password = secrets.get_service_password(
+            config.keyring_service, login, oracle=password
+        )
+    return password
+
+
+def build_annotations(
+    main_config: schema.MainSectionConfig,
+    annotations: Iterable[tuple[str, str]],
+    url: Optional[str] = None,
+) -> list[str]:
+    """Format annotations, respecting configuration values.
+
+    :param `annotations`: Comments from service.
+    :param `url`: Url to prepend to the annotations.
+    """
+    final = []
+    if url and main_config.annotation_links:
+        final.append(get_processed_url(main_config, url))
+    if main_config.annotation_comments:
+        for author, message in annotations:
+            message = message.strip()
+            if not message or not author:
+                continue
+
+            if not main_config.annotation_newlines:
+                message = message.replace('\n', '').replace('\r', '')
+
+            annotation_length = main_config.annotation_length
+            if annotation_length:
+                message = '%s%s' % (
+                    message[:annotation_length],
+                    '...' if len(message) > annotation_length else '',
+                )
+            final.append('@%s - %s' % (author, message))
+    return final
+
+
+def render_tags_from_labels(
+    config: schema.ServiceConfig, record: dict[str, Any], labels: list[str]
+) -> list[str]:
+    """Transform labels into suitable taskwarrior tags using the label template."""
+    return [
+        Template(config.label_template).render(
+            {**record, "label": re.sub(r'[^a-zA-Z0-9]', '_', label)}
+        )
+        for label in labels
+    ]
+
+
+def get_tags_from_labels(
+    config: schema.ServiceConfig,
+    record: dict[str, Any],
+    labels: list[str],
+    toggle_option: str = 'import_labels_as_tags',
+    template_option: str = 'label_template',
+    template_variable: str = 'label',
+) -> list[str]:
+    """Transform labels into suitable taskwarrior tags, respecting configuration options."""
+    using_deprecated_parameters = (
+        toggle_option != 'import_labels_as_tags'
+        or template_option != 'label_template'
+        or template_variable != 'label'
+    )
+    if using_deprecated_parameters:
+        log.warning(
+            "Deprecation Warning: Issue.get_tags_from_labels's toggle_option, "
+            "template_option, and template_variable parameters are deprecated and "
+            "will be removed in a future API version."
+        )
+
+    if not getattr(config, toggle_option):
+        return []
+
+    if not using_deprecated_parameters:
+        return render_tags_from_labels(config, record, labels)
+
+    # deprecated path, to be removed once we remove the deprecated parameters.
+    context = record.copy()
+    label_template = Template(getattr(config, template_option))
+    tags = []
+
+    for label in labels:
+        normalized_label = re.sub(r'[^a-zA-Z0-9]', '_', label)
+        context.update({template_variable: normalized_label})
+        tags.append(label_template.render(context))
+
+    return tags
+
+
+def build_default_description(
+    main_config: schema.MainSectionConfig,
+    title: str = '',
+    url: str = '',
+    number: str | int = '',
+    cls: str = "issue",
+) -> str:
+    """Return a default description, respecting configuration options.
+
+    :param `title`: Short description of the task.
+    :param `url`: URL to the task on the service.
+    :param `number`: Number associated with the task on the service.
+    :param `cls`: The abbreviated type of task this is. Preferred options
+        are ('issue', 'pull_request', 'merge_request', 'todo', 'task',
+        'subtask').
+    """
+    cls_markup = {
+        'issue': 'Is',
+        'pull_request': 'PR',
+        'merge_request': 'MR',
+        'todo': '',
+        'task': '',
+        'subtask': 'Subtask #',
+    }
+    url_separator = ' .. '
+    url = get_processed_url(main_config, url) if main_config.inline_links else ''
+    desc_len = main_config.description_length
+    return "(bw)%s#%s - %s%s%s" % (
+        cls_markup.get(cls, cls.title()),
+        number,
+        title[:desc_len] if desc_len else title,
+        url_separator if url else '',
+        url,
+    )
+
+
+def _apply_templates(
+    config: schema.ServiceConfig, task: "Task", extra: dict[str, Any]
+) -> None:
+    """Render the user's field and tag templates onto a mapped task.
+
+    A field template overwrites whatever the service mapped for that field
+    (including the default description); added tags are appended.
+    """
+    context = {**task.to_taskwarrior_data(), **extra}
+    for field, template in config.templates.items():
+        setattr(task, field, Template(template).render(context))
+
+    for tag_template in config.add_tags:
+        tag = Template(tag_template).render(context)
+        if tag:
+            task.tags.append(tag)
+
+
 class Service(abc.ABC):
     """Base class for fetching issues from the service.
 
@@ -109,51 +266,6 @@ class Service(abc.ABC):
 
         log.info("Working on [%s]", self.config.target)
 
-    def get_secret(self, key: str, login: str = 'nousername') -> str:
-        """Get a secret value, potentially from an :ref:`oracle <Secret Management>`.
-
-        The secret key need not be a *password*, per se.
-
-        :param `key`: Name of the configuration field of the given secret.
-        :param `login`: Username associated with the password in a keyring, if
-            applicable.
-        """
-        password = getattr(self.config, key)
-        if not password or password.startswith("@oracle:"):
-            password = secrets.get_service_password(
-                self.config.keyring_service, login, oracle=password
-            )
-        return password
-
-    def build_annotations(
-        self, annotations: Iterable[tuple[str, str]], url: Optional[str] = None
-    ) -> list[str]:
-        """Format annotations, respecting configuration values.
-
-        :param `annotations`: Comments from service.
-        :param `url`: Url to prepend to the annotations.
-        """
-        final = []
-        if url and self.main_config.annotation_links:
-            final.append(get_processed_url(self.main_config, url))
-        if self.main_config.annotation_comments:
-            for author, message in annotations:
-                message = message.strip()
-                if not message or not author:
-                    continue
-
-                if not self.main_config.annotation_newlines:
-                    message = message.replace('\n', '').replace('\r', '')
-
-                annotation_length = self.main_config.annotation_length
-                if annotation_length:
-                    message = '%s%s' % (
-                        message[:annotation_length],
-                        '...' if len(message) > annotation_length else '',
-                    )
-                final.append('@%s - %s' % (author, message))
-        return final
-
     @abc.abstractmethod
     def to_taskwarrior(self, record: dict[str, Any], extra: dict[str, Any]) -> "Task":
         """Transform a foreign record into a taskwarrior Task."""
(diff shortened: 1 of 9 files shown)


ryneeverett (commented):
>     1. Using our own Task model decreases the dependency to taskw, and improves typing / readability, instead of using plain dicts. Having our own abstraction simplifies the changes needed to switch the dependency from  taskw to taskchampion.
>     ...
> If we proceed with this changes, I would then use Task even more in `db.py`, and try avoid using dicts. We could also replace calls to `taskw.task.FIELDS` by our own `Task` object

This makes sense to me in principle, but it doesn't seem like the decreased dependency on taskw has manifested in this PR. I'm kind of thinking a well-scoped PR might be focused on completely abstracting away taskw usage to prepare for #1078 . What do you think?
 
>     2. The `Issue` (and child classes) becomes less useful, it does not hold any specific logic, I suggest an implementation where it's merged into Service.

I think there is a bit of a principled distinction between the two (described in the original doc-comments) but it might not be useful and probably isn't consistently enforced. I also think it's going to be hard to evaluate whether this change is an improvement if it is done at the same time as the Task model. Maybe we can hold off on this in order to reduce the scope of the Task model transition and so that we can better focus on getting it right if/when we do it?

ryneeverett:
Thinking about this a bit more, I'd note that I think we are essentially on the same page. I see where removing UDA's from Issue makes that class even less useful and suggests elimination. I'd still suggest pushing that out to a separate PR though.

Lotram:
> This makes sense to me in principle, but it doesn't seem like the decreased dependency on taskw has manifested in this PR. I'm kind of thinking a well-scoped PR might be focused on completely abstracting away taskw usage to prepare for https://github.com/GothenburgBitFactory/bugwarrior/issues/1078 . What do you think?

Yes, as I see it, removing taskw dependency would be enabled by introducing our own `Task` model, thus in a follow-up PR. The main point of this current PR is my point 3: clearer split between external data and taskwarrior data. 

But I definitely can remove the "Issue into Service" changes from this PR, and we'll do that later. 
In my opinion, the main reason to do everything now is to avoid doing multiple breaking changes: I'd rather have 1 really big API change than multiple smaller ones. But as long as there is no released versions in between, it's not a problem, and we can have multiple PRs, I'm ok with that

ryneeverett:
> 3. This gives us a clearer split between data coming from external services (Service, record) and data inserted in taskwarrior.

Could you elaborate on this a little bit? I noticed that some of the naming was more consistent which i appreciated, but I'm assuming you mean something more abstract/conceptual...

Lotram:
> Could you elaborate on this a little bit? I noticed that some of the naming was more consistent which i appreciated, but I'm assuming you mean something more abstract/conceptual...


 *  The validation is done bit by bit in several places: some in `get_issues`, some in `to_taskwarrior`, and some later in collect.py (`aggregate_issues,  `TaskConstructor`, etc). 
 * Said differently, or at least on the same subject, there is currently no single source of truth to handle parsing / validating the data before inserting it to taskwarrior (and especially the "generic" / non-udas data). Each Issue subclass is responsible for calling the right parsers (e.g `parse_date`), thus a lot of code duplication + potential errors
 * Because of this, there is absolutely no guarantee on what's inside the dict returned by `to_taskwarrior` and used in collect.py / db.py. That's why I had to remove some other validation logic from db.py in an earlier MR, there was nothing telling us the data was impossible (I remember at least a check for `priority = ''` which was done in multiple places). 

My point is to have a single "entrypoint" (not sure it's the right word here), the `Task` Model, for all the `collect.py` / `db.py` logic. Each service should only be responsible for retrieving the data from their source, and comply with the `Task` API contract. Once we have this `Task` object, we should be able to assume all the data has been properly validated (datetimes have the right format, priority is the right enum, etc)


At some point, I think we'll also be able to add all the "update an existing task" logic inside that model (e.g. updating the tags, the annotations, etc)

bugwarrior/services/__init__.py
```diff
@@ -265,10 +239,13 @@ class Service(abc.ABC, Generic[T_Issue]):
 
     #: Which version of the API does this service implement?
     API_VERSION: float
-    #: Which class should this service instantiate for holding these issues?
-    ISSUE_CLASS: type[T_Issue]
+    #: Which Udas model declares this service's UDAs and unique key?
+    UDAS_CLASS: type["Udas"]
```
ryneeverett: Would it make more sense to put the `TASK_SCHEMA: type["Task"]` here and access `udas` through that? Seems odd to not put the highest level schema here and like we could end up needing it some day.

ryneeverett (commented):
Gotcha. I do see the benefits of centralizing validation. I think transitioning to the Task model is an acceptable PR scope.

Lotram:
@ryneeverett I finally finished working on this MR. As all the services used to have their own implementation (e.g. no shared taskwarrior field definition, `to_taskwarrior` returning a simple `dict`), with no proper common interface, it's not easy to ensure I did not break anything. 

I kept the `Issue` class for now, we can remove it in another MR.

ryneeverett:
Sorry I've been too busy to look at this, but please rebase.

bugwarrior/docs/other-services/tutorial.rst
```diff
@@ -124,41 +126,57 @@ Unless you're using a library that closely aligns with the needs of your service
...
+      gitbugauthor: str = Field(title='Gitbug Issue Author')
+      gitbugid: str = Field(title='Gitbug UUID')
+      gitbugstate: str = Field(title='Gitbug state')
+      gitbugtitle: str = Field(title='Gitbug Title')
+
+
+  class GitBugTask(Task):
+      udas: GitBugUdas
+
+Every field of a ``Udas`` subclass is a UDA. Its name is the UDA name, so the author will be assigned to ``gitbugauthor``. The taskwarrior UDA type is derived from the annotation: ``str`` becomes "string", ``int`` and ``float`` become "numeric", ``datetime`` becomes "date" and ``timedelta`` becomes "duration". The label comes from the field's ``title``.
```
ryneeverett: A table would be a better way to express the type conversions.

CHANGELOG.rst
```diff
@@ -5,6 +5,8 @@ Highlights
 ==========
 - Deprecation of the --interactive flag. All invocations are now interactive. (https://github.com/GothenburgBitFactory/bugwarrior/pull/1204)
 - Service API 2.0 (with backwards compatiblity for 1.0 services). See https://bugwarrior.readthedocs.io/en/stable/other-services/api.html#changelog. (https://github.com/GothenburgBitFactory/bugwarrior/pull/1209)
+- Services now map a foreign record to a typed Task model rather than a plain dictionary, and declare their UDAs as fields of that model rather than as a separate UDAS dictionary. (https://github.com/GothenburgBitFactory/bugwarrior/pull/1226)
```
ryneeverett, line 8: I think there should only be a single entry for the "Service API 2.0" (with multiple PR links) because changelog readers aren't that interested in internal implementation details.

Also, backwards compatibility for 1.0 services is now unrealistic right? So we should remove that claim and also remove the `get_keyring_service` backwards compatibility shim.

CHANGELOG.rst
```diff
@@ -15,6 +17,15 @@ Dependency Updates
...
+- bugzilla: ``bugzillaneedinfo`` and ``bugzillaassignedon`` are now always written to the record. They were previously left out when empty, so a date written on an earlier run stayed there after the flag was gone. It is now cleared. (https://github.com/GothenburgBitFactory/bugwarrior/pull/1226)
+- jira: ``jiracreatedts`` is no longer declared. It has never been populated. The creation timestamp goes to the generic ``entry`` field. To record it in a UDA, use ``extra_fields = jiracreatedts:created`` and declare the UDA in Taskwarrior as you would for any other extra field. (https://github.com/GothenburgBitFactory/bugwarrior/pull/1226)
+- logseq: ``logseqdone`` is no longer declared. It has never been populated. (https://github.com/GothenburgBitFactory/bugwarrior/pull/1226)
+- redmine: spent and estimated hours are converted to a duration directly, instead of running ``task calc`` once per issue. (https://github.com/GothenburgBitFactory/bugwarrior/pull/1226)
```
ryneeverett: IMO, these items would be better noted in more verbose commit messages. I try to only put the most important highlights that are most likely to impact users in the changelog. I'd be open to a more comprehensive changelog in the future but I think consistently sticking with one approach would be better for the current release cycle.

```suggestion
```

bugwarrior/services/__init__.py
```diff
@@ -345,32 +296,59 @@ def build_annotations(
...
+            tag
+            for tag_template in self.config.add_tags
+            if (tag := Template(tag_template).render(context))
+        ]
+        if added_tags:
+            # Assign instead of appending: appending to the list does not
+            # mark tags as set, and unset fields are left out of the record.
+            task.tags = [*task.tags, *added_tags]
 
+    def process_record(
```
ryneeverett, line 332: I think one of the objectives of the API should be that people can understand a service of interest only by looking at the service code. I find `process_record` mysterious whereas `get_issue_for_record` made the intent obvious. Rename to `get_task_for_record`? Maybe you have a better idea.

bugwarrior/services/__init__.py
```diff
@@ -345,32 +296,59 @@ def build_annotations(
...
-            }
+        A field template replaces what the service mapped, and added tags are
+        appended. A template can use the fields of the task, the data the
+        service computed and the default description. The default description
+        is added last, so that a service with its own "description" in its
+        computed data does not hide it.
+        """
+        context = {
+            **task.to_taskwarrior_data(),
+            # A service which never assigns tags emits no "tags" in its record.
```
ryneeverett, line 310: I don't understand what this comment is supposed to convey.

bugwarrior/services/__init__.py
```diff
@@ -345,32 +296,59 @@ def build_annotations(
...
+                setattr(task, field, value)
+            except ValidationError as e:
+                raise_template_error(self.config.target, field, template, value, e)
+
+        added_tags = [
+            tag
+            for tag_template in self.config.add_tags
+            if (tag := Template(tag_template).render(context))
+        ]
+        if added_tags:
```
ryneeverett, line 327: Can't we go ahead and re-assign this even if added_tags is empty?
```suggestion
```

bugwarrior/docs/other-services/api.rst
```diff
@@ -22,6 +27,24 @@ Changelog
...
+- Added ``Service.TASK_SCHEMA``, naming the ``Task`` subclass a service maps to.
+- Added ``Service.process_record(record, extra)``, which maps a record to a
+  ``Task`` and applies the user's templates. ``Service.issues()`` now yields
+  those tasks rather than ``Issue`` objects.
+- Removed the type parameter from ``Service``. Subclasses declare
+  ``class MyService(Service)`` rather than ``class MyService(Service[MyIssue])``.
+- Removed ``Issue.parse_date(date)``. A date field accepts the service's raw
+  string and parses it, adding UTC when the string carries no timezone. Use
+  ``bugwarrior.task.coerce_datetime`` if you need the datetime before the task
+  is built.
```
ryneeverett, line 47: I don't think we need to make `coerce_datetime` part of the stable API, but I think it's fine that the jira service uses it. (It already isn't documented by automodule because it has no docstring.)

bugwarrior/docs/other-services/api.rst
```diff
@@ -10,6 +10,11 @@ warning, release notes, or semantic version bumping.
    :exclude-members: __init__,model_config
    :member-order: bysource
 
+.. automodule:: bugwarrior.task
+   :members:
+   :exclude-members: __init__,model_config
```
ryneeverett, line 15: These don't need to be part of the stable API's since services don't implement or use them.

```suggestion
   :exclude-members: __init__,model_config,to_taskwarrior_data,get_udas
```

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+    scheduled: IssueDatetime = None
+    start: IssueDatetime = None
+    status: Status = None
+    tags: list[str] = Field(default_factory=list)
+    until: IssueDatetime = None
+    urgency: int | float | None = None
+    uuid: str | None = None
+    wait: IssueDatetime = None
+
+    udas: Udas = Field(default_factory=Udas)
```
ryneeverett, line 192: Add a [doc-comment](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comment) to document this API.


```suggestion
    #: The service's :class:`Udas` subclass.
    udas: Udas = Field(default_factory=Udas)
```

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+def _coerce_status(value: Any) -> Any:
+    if isinstance(value, str):
+        return value.lower()
+    return value
+
+
+def _coerce_optional_config_list(value: Any) -> Any:
+    if value is None:
+        return None
+    return parse_config_list(value)
```
ryneeverett, line 45: I'm wondering if we should be less flexible with our inputs here and force services to be more consistent.

- `Any` inputs: Shouldn't these mostly be `str | None`?
- `upper()`/`lower()`: Why don't we force services to do this and allow a validation error otherwise?

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class Udas(BaseModel):
+    """A service's UDA declarations.
+
+    Every field is a UDA. The base class has no fields: it is what a task
+    without UDAs uses, and the default for Task. A service subclasses it to
+    declare its own UDAs, and the unique key which identifies a task in the
+    remote service.
+    """
```
ryneeverett, line 107: A table of python types -> taskwarrior types would be nice here for the API docs.

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+        non_none = [arg for arg in get_args(annotation) if arg is not type(None)]
+        if len(non_none) == 1:
+            annotation = non_none[0]
+    if annotation in (datetime.datetime, datetime.date):
+        return "date"
+    if annotation is datetime.timedelta:
+        return "duration"
+    if annotation in (int, float, bool):
+        return "numeric"
+    return "string"
```
ryneeverett, line 97: Maybe this would be a good place to use `match`/`case`?

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+
+
```
ryneeverett, line 99: One of the properties of the former `UDAS` dict was that there was a 1:1 mapping between the "type" field and the taskwarrior data type bugwarrior would use. The point of validating the Task model is to make sure that the data services can return is valid for the data type in which it's being stored, right? I'm wondering if there is any benefit in constraining the type to be narrower than that -- for example specifying `int` rather than `int | float | bool`? I'm also noticing that UDA's sometimes are optional (e.g., `str | None`) and sometimes required (`str`) and I'm wondering if they shouldn't always be optional **if** the sole purpose of the validation is to preempt invalid taskwarrior data. If we restored the 1:1 mapping, for uda's there might just be 5 types corresponding to the taskwarrior data types. (e.g., `String = str | None`). But maybe there are additional benefits of these pydantic models I'm missing?

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+    remote service.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    #: Field names (UDA or generic) which identify a task in the remote
+    #: service. There is no default on purpose: an empty unique key matches
+    #: every pending task in the user's database, so a service which forgets
+    #: it must fail rather than run.
+    UNIQUE_KEY: ClassVar[tuple[str, ...]]
```
ryneeverett, line 115: Should the doc-comment clarify that these are strings? Could we add a validator to ensure the values correspond to existing fields?

> (UDA or generic)

Do any services use generic fields for UNIQUE_KEY? It isn't obvious to me that that would ever be appropriate. Maybe it's simpler to limit this to UDA's?

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+    def udas_model(cls) -> type[Udas]:
+        return cls.model_fields["udas"].annotation
+
+    @classmethod
+    def get_udas(cls) -> dict[str, dict[str, str]]:
+        return cls.udas_model().get_udas()
+
+    @classmethod
+    def get_unique_key(cls) -> tuple[str, ...]:
+        return cls.udas_model().UNIQUE_KEY
```
ryneeverett, line 150: I would tend to use `@property` for these rather than `get_` methods, but if you're going to use `get_` methods I'd make it consistent by renaming `udas_model` -> `get_udas_model`.

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+    model_config = ConfigDict(extra="forbid")
+
+    #: Field names (UDA or generic) which identify a task in the remote
+    #: service. There is no default on purpose: an empty unique key matches
+    #: every pending task in the user's database, so a service which forgets
+    #: it must fail rather than run.
+    UNIQUE_KEY: ClassVar[tuple[str, ...]]
+
+    @classmethod
+    def get_udas(cls) -> dict[str, dict[str, str]]:
```
ryneeverett, line 118: Can we rename this to be more descriptive since the whole class is `Udas` -- maybe `to_taskrc`?

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+        return cls.udas_model().get_udas()
+
+    @classmethod
+    def get_unique_key(cls) -> tuple[str, ...]:
+        return cls.udas_model().UNIQUE_KEY
+
+    def to_taskwarrior_data(self) -> dict[str, Any]:
```
ryneeverett, line 152: Maybe rename `to_task_data` to distinguish from taskrc data which uda's also get formatted to.

bugwarrior/task.py
```diff
@@ -0,0 +1,192 @@
...
+    project: str | None = None
+    recur: str | datetime.timedelta | None = None
+    scheduled: IssueDatetime = None
+    start: IssueDatetime = None
+    status: Status = None
+    tags: list[str] = Field(default_factory=list)
+    until: IssueDatetime = None
+    urgency: int | float | None = None
+    uuid: str | None = None
+    wait: IssueDatetime = None
```
ryneeverett, line 190: Where did you get this list of attributes from? Some of these -- like the recurrence ones imask, mask, parent, etc. -- would make no sense for a service to set. I wonder if we shouldn't just start off with the fields currently used by services and add more on an as-needed basis. This came to my attention particularly because the recurrence attributes are going to be replaced with different attributes in an upcoming release of taskwarrior. I'm not sure if a comprehensive list of taskwarrior attributes is actually being maintained, but even if it is I'm not sure we should track it because then we'd have to target a specific taskwarrior version.

Lotram: ~I used the  `task columns` command, and remove a few fields, but you're right, I should have at least removed all the read only ones.~
Edit: I did try to use `task columns` but I actually ended up simply copying the list from `taskw` (`taskw.task.Task.FIELDS`), it does have all the `mask`, `imask`, `parent` etc fields

>  I wonder if we shouldn't just start off with the fields currently used by services and add more on an as-needed basis.

ok I'll try that

bugwarrior/task.py
Lotram: Since python 3.13, [we can't chain classmethod with another decorator anymore](https://docs.python.org/3/whatsnew/3.13.html#builtins):
> Remove support for chained [classmethod](https://docs.python.org/3/builtins/functions.html#classmethod) descriptors (introduced in [gh-63272](https://github.com/python/cpython/issues/63272)). These can no longer be used to wrap other descriptors, such as [property](https://docs.python.org/3/builtins/functions.html#property).

so I kept the classmethod, but renamed it

PR #1228, opened 2026-07-28
Use Ruff default ruleset, in addition to our own set

Lotram:
ruff recently released [v0.16](https://astral.sh/blog/ruff-v0.16.0)

While looking at the changelog, I realized we use `select` instead of `extend-select`, so it overrides any default rule configured by ruff. This switches to `extend-select`, so we benefit from the default set of rules. 

I chose to ignore some rules:
- [RUF012](https://docs.astral.sh/ruff/rules/mutable-class-default/): we rely a lot on classes, I don't think mutable defaults really are a problem here
- DTZ rules: working with timezone-aware datetimes is usually a good idea, but taskwarrior uses naive datetimes anyway, so I'm not sure it makes sense to switch to tz aware everywhere. If we do think it's better, we can change in another MR. 

I also noticed two real existing bugs: 
1. `bugwarrior/db.py` (run_hooks): the error message expected 2 inputs, while only 1 was provided.
2. `bugwarrior/services/gerrit.py` (annotations): `.lstrip('Patch Set ')` was used instead of `removeprefix`, removing any matching char (e.g. "this is a message" became "is is a message")

In the last commit, I removed the "quote-style" ruff setting, to use the default one, so we have a slightly more consistent codebase. I can remove this commit if we don't want that change for now.


ryneeverett (commented):
Why wasn't .format() -> f-string caught by [UP032](https://docs.astral.sh/ruff/rules/f-string/)?

Lotram:
It looks like UP032 does not handle `.format()` calls when its argument contains a quote or a `**kwargs`

I migrated everything for consistency sake

ryneeverett (changes requested):
I'd like you to drop 5de1539. Otherwise this seems good to go.

Style standards which are not enforced by formatters or even linters are problematic because they open the door to thinking or quibbling about style, the avoidance of which is the main point of these tools.

I think [UP032](https://docs.astral.sh/ruff/rules/f-string/) overstates the case against `.format()` when they say:

> f-strings are more readable and generally preferred over `str.format` calls.

If you look at [pyupgrade](https://github.com/asottile/pyupgrade#f-strings) from which this rule was derived, they only give examples of clear readability improvements and say:

> note: pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability).

I would give the first modification in `bugwarrior/services/__init__.py` as an example of a readability degradation:

```diff
         desc_len = self.main_config.description_length
-        return "(bw){}#{} - {}{}{}".format(
-            cls_markup.get(cls, cls.title()),
-            number,
-            title[:desc_len] if desc_len else title,
-            url_separator if url else '',
-            url,
+        title_text = title[:desc_len] if desc_len else title
+        return (
+            f"(bw){cls_markup.get(cls, cls.title())}#{number} - {title_text}"
+            f"{url_separator if url else ''}{url}"
         )
```

I'd be fine with many of the format -> f-string switches here but the motivation would be readability on a case-by-case basis and not consistency.

Lotram:
> Style standards which are not enforced by formatters or even linters are problematic because they open the door to thinking or quibbling about style, the avoidance of which is the main point of these tools.

You're right, I'll drop it

ryneeverett (approved):

PR #1230, opened 2026-09-11
gitbug: Only query open issues

### LINKED ISSUE #1229: [gitbug] Stop fetching closed issues
I have been an on-and-off user of bugwarrior with the git-bug service for a few years now and one thing that has always stopped me is that it fetches _all_ issues from the issue tracker (open and closed) into taskwarrior.

I now understand that the other services do not do this, and this may be a bug in the implementation?

Current behavior:

- Context: Repository with git-bug issue tracker with one open and one closed issue.
- `bugwarrior pull` imports _both_ issues into taskwarrior as pending tasks.
- I close the 2nd issue in git-bug.
- `bugwarrior pull` keeps both issues imported as pending tasks in taskwarrior.

Manually completing in task (`task <id> done`) and pulling will mark them as `status:pending` again in taskwarrior even if they are closed in git-bug (due to bugwarrior's pull-only model).
Only issues that are **deleted** in git-bug with `git bug bug rm <id>` will be marked completed in taskwarrior.

I believe this goes against the intended UX in the issue tracker (i.e. close finished issues instead of deleting them).

Expected behavior:

- Context: Repository with git-bug issue tracker with one open and one closed issue.
- `bugwarrior pull` imports only the open issue into taskwarrior as pending.
- I close the 2nd issue in git-bug.
- `bugwarrior pull` marks the 2nd issue's task as completed.

Taking a look at other Services, e.g. [GitHub](https://github.com/GothenburgBitFactory/bugwarrior/blob/4ed78ac2734e0979268d41ff22ef8b6d4a7a40ee/bugwarrior/services/github.py#L441) restricts queries to `status:open`, as does [GitLab](https://github.com/GothenburgBitFactory/bugwarrior/blob/4ed78ac2734e0979268d41ff22ef8b6d4a7a40ee/bugwarrior/services/gitlab.py#L325).

My question: is the current behavior the intended behavior for bugwarrior?

For the moment I am using a simple [fix which drops](https://github.com/GothenburgBitFactory/bugwarrior/commit/2a187e7032507bab82686fe6e87dc19f5e444605) closed issues before importing.
If this is indeed a bug and not intended I am happy to prepare a more polished version as PR for bugwarrior.

marty-oehme:
Closes #1229.

---

Exclude closed git-bug issues from sync query, mirroring behavior of other bugwarrior services. This means issues that are closed in git-bug will complete the corresponding tasks in taskwarrior.

Filters for only open issues directly in GraphQL query so no closed issue should reach taskwarrior. Additionally adds small python condition to guard against git-bug upstream filter failures or query syntax drift.

Adds tests for individual closed issue and mixed open and closed issue querying, and a query builder test to avoid accidental syntax regressions.

---

Let me know if either the additional code-guard or the query builder test are unwanted, kept them as additional future insurance if upstream changes on their end or someone tries to change the query on ours.

### DIFF
--- a/bugwarrior/services/gitbug.py
+++ b/bugwarrior/services/gitbug.py
@@ -83,7 +83,7 @@ class GitBugClient(Client):
 
     def get_issues(self) -> list[dict[str, Any]]:
         return self._query_graphql(
-            "{{ repository {{ allBugs {{ nodes {{ {} }} }} }} }}".format(
+            '{{ repository {{ allBugs(query: "status:open") {{ nodes {{ {} }} }} }} }}'.format(
                 " ".join(
                     [
                         "author { name }",
@@ -157,6 +157,9 @@ class GitBugService(Service[GitBugIssue]):
 
     def issues(self) -> Iterator[GitBugIssue]:
         for issue in self.client.get_issues():
+            # Guard against closed issues leaking through GraphQL query.
+            if issue.get("status") != "OPEN":
+                continue
             comments = issue.pop("comments")
             issue["description"] = comments["nodes"].pop(0)["message"]
 
--- a/tests/services/test_gitbug.py
+++ b/tests/services/test_gitbug.py
@@ -80,6 +80,40 @@ class TestGitBugIssue:
 
         assert TaskConstructor(issue).get_taskwarrior_record() == expected
 
+    def test_issues_skips_closed(self, service, record):
+        closed_record = {**record, "status": "CLOSED"}
+        service.client.get_issues = mock.MagicMock(return_value=[closed_record])
+        issues = list(service.issues())
+
+        assert issues == []
+
+    def test_issues_mixed_open_closed(self, service, record):
+        closed_record = {**record, "status": "CLOSED"}
+        service.client.get_issues = mock.MagicMock(return_value=[closed_record, record])
+        issues = list(service.issues())
+
+        assert len(issues) == 1
+        assert (
+            TaskConstructor(issues[0]).get_taskwarrior_record()["gitbugid"]
+            == record["id"]
+        )
+
+
+class TestGitBugClient:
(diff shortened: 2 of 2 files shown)


bugwarrior/services/gitbug.py
```diff
@@ -157,6 +157,9 @@ def __init__(
 
     def issues(self) -> Iterator[GitBugIssue]:
         for issue in self.client.get_issues():
+            # Guard against closed issues leaking through GraphQL query.
+            if issue.get("status") != "OPEN":
```
Lotram: is this really needed, as the graphql query already filters on open issues, in `get_issues` ?

marty-oehme: Kept it as a future insurance if git-bug changes their graphql syntax/capabilities, but realistically that is very low change. Added commit removing that part of the PR.

Lotram: sidenote on that topic of API changes and payloads: @ryneeverett , I've been wondering on how to more easily test it (and maybe detect changes). Maybe the developer documentation should include a guide on how to set up each service to have an easy way to test it live. We could also store real examples of payloads somewhere in the tests, but we wouldn't be able to detect changes with that.

Obviously well beyond this MR scope

ryneeverett: I'm not sure I see the advantage of real payloads in tests. Assuming the service works at the time the tests are written, can't we conclude that the test data accurately represents the payload structure?

I'm open to it but I'm not sure documentation is the most effective way to live test services.

For hosted services, we could have a bugwarrior development account and run live tests in CI. To me this seems more useful than documentation because I'm way too lazy to create accounts for services I don't use.

For the few open source services like gitbug, we could have additional automated tests that actually run the service. Ideally a containerized solution that can run either locally or in CI.

